PR #15, opened 2020-11-18
Adding retry strategy support.

rccarper:
*Description of changes:*
Adding retry strategy support that meta requests can use to reschedule failed requests.


source/s3_meta_request.c
```diff
@@ -401,13 +405,7 @@ void aws_s3_meta_request_send_next_request(
...
     }
 
-    /* Try grabbing a request_desc from the retry queue. */
-    if (!aws_linked_list_empty(&meta_request->synced_data.retry_queue)) {
-        struct aws_linked_list_node *request_node = aws_linked_list_pop_front(&meta_request->synced_data.retry_queue);
-
-        request = AWS_CONTAINER_OF(request_node, struct aws_s3_request, node);
-        AWS_FATAL_ASSERT(request != NULL);
-    }
+    request = aws_s3_meta_request_retry_queue_pop_synced(meta_request);
```
bretambrose, line 408: feels like this is simple enough now that unlock: could be dropped in place of a single if-else.

source/s3_meta_request.c
```diff
@@ -439,21 +440,11 @@ void aws_s3_meta_request_send_next_request(
...
-            aws_s3_request_release(request);
-
-            goto call_finished_callback;
-        } else {
-            goto error_finish;
-        }
+        goto call_finished_callback;
     }
 
     AWS_LOGF_TRACE(
```
bretambrose, line 450: Seems more like info.  TRACE is all but unusable for anything other than protocol-level debugging.

source/s3_meta_request.c
```diff
@@ -842,20 +832,269 @@ static void s_s3_meta_request_send_request_finish_default(
...
+    /* Push the request into the retry queue so that it can actually be retried. */
+    aws_s3_meta_request_retry_queue_push(meta_request, request);
+
+    /* Tell the client we have additional work that can be done. */
+    aws_s3_meta_request_schedule_work(meta_request);
+
+clean_up:
+
+    aws_s3_request_release(request);
+    aws_s3_meta_request_release(meta_request);
```
bretambrose, line 875: Just curious why this release is done.  I'd expect if you have a handle to the request (via user data) than it would be its responsibility to perform a release on the meta request if it was going away.  It seems a little confusing to have a failure lead to a double dec ref rather than a nested one.

source/s3_meta_request.c
```diff
@@ -842,20 +832,269 @@ static void s_s3_meta_request_send_request_finish_default(
...
+            aws_error_str(error_code));
+
+        goto error_ref_clean_up;
+    }
+
+    AWS_ASSERT(token);
+
+    enum aws_retry_error_type error_type = AWS_RETRY_ERROR_TYPE_TRANSIENT;
+
+    if (error_code == AWS_ERROR_S3_NO_PART_BUFFER) {
```
bretambrose, line 909: switch maybe

bretambrose: err I'm confused, doesn't the previous if make this entire if-chain useless?

source/s3_meta_request.c
```diff
@@ -842,20 +832,269 @@ static void s_s3_meta_request_send_request_finish_default(
...
+    }
+
+    return;
+
+error_ref_clean_up:
+
+    aws_s3_request_release(request);
+    aws_s3_meta_request_release(meta_request);
+
+    aws_s3_meta_request_finish(meta_request, NULL, 0, aws_last_error());
```
bretambrose, line 951: I don't think this is a reliable way of propagating error type.  I wouldnot use aws_last_error for that and instead base if off error_code or a local, mutable mirror of it.

source/s3_meta_request.c
```diff
@@ -842,20 +832,269 @@ static void s_s3_meta_request_send_request_finish_default(
...
+
+    struct aws_byte_cursor host_header_name = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("Host");
+
+    struct aws_http_headers *message_headers = aws_http_message_get_headers(message);
+    AWS_ASSERT(message_headers);
+
+    struct aws_byte_cursor host_header_value;
+
+    bool host_header_exists =
+        aws_http_headers_get(message_headers, host_header_name, &host_header_value) == AWS_ERROR_SUCCESS;
```
bretambrose, line 974: maybe just if (blah) { log, goto error_ref_clean_up} like the below failure?

source/s3_meta_request.c
```diff
@@ -842,20 +832,269 @@ static void s_s3_meta_request_send_request_finish_default(
...
+
+        goto error_ref_clean_up;
+    }
+
+    return;
+
+error_ref_clean_up:
+
+    aws_s3_request_release(request);
+    aws_s3_meta_request_finish(meta_request, NULL, 0, aws_last_error());
```
bretambrose, line 998: If aws_s3_meta_request_finish judges success/failure via the error_code, I would strongly suggest never passing aws_last_error() in but aws_last_error_or_unknown() to guarantee the failure path is taken even when your downstream calls forget to raise_error on a failure.

source/s3_meta_request.c
```diff
@@ -842,20 +832,269 @@ static void s_s3_meta_request_send_request_finish_default(
...
+            error_code,
+            aws_error_str(error_code),
+            (void *)request,
+            response_status);
+
+        aws_s3_meta_request_finish(meta_request, request, response_status, error_code);
+        return;
+    }
+
+    struct aws_s3_client *client = aws_s3_meta_request_acquire_client(meta_request);
```
bretambrose, line 1040: Why does this need to be done when presumably the meta request already has a ref?  Is it just "this is the only way to get a pointer to the client?"

source/s3_meta_request.c
```diff
@@ -842,20 +832,269 @@ static void s_s3_meta_request_send_request_finish_default(
...
+
+    /* If we were able to get the point of handling an error, the client should still be around. */
+    AWS_ASSERT(client);
+    AWS_ASSERT(request);
+    AWS_ASSERT(request->meta_request == meta_request);
+
+    aws_s3_request_acquire(request);
```
bretambrose, line 1047: These are scary.

source/s3_meta_request.c
```diff
@@ -842,20 +832,269 @@ static void s_s3_meta_request_send_request_finish_default(
...
+    aws_s3_meta_request_acquire(meta_request);
+
+    /* If the retry token is NULL, try to grab one, otherwise, just re-use the token from the request's last retry. */
+    if (request->retry_token == NULL) {
+        s_s3_meta_request_acquire_retry_token(client->retry_strategy, request);
```
bretambrose, line 1052: Can either of these fail?

source/s3_meta_request.c
```diff
@@ -842,20 +832,269 @@ static void s_s3_meta_request_send_request_finish_default(
...
+
+void aws_s3_meta_request_retry_queue_push(struct aws_s3_meta_request *meta_request, struct aws_s3_request *request) {
+    AWS_PRECONDITION(meta_request);
+    AWS_PRECONDITION(request->meta_request == meta_request);
 
     aws_s3_request_acquire(request);
 
     aws_s3_meta_request_lock_synced_data(meta_request);
 
+    aws_s3_meta_request_release(request->meta_request);
```
bretambrose, line 1070: What makes this unlink/link on retry necessary from a ref handling perspective?  Is it impossible to initiate cancel if the retry queue also has refs to the meta request?

bretambrose (commented):
One final thing, I see a lot of TRACE logging which I don't think we want to do when we're talking request-level events.  TRACE is a 100MB/s spam of task scheduling and channel buffer pushes.  At the very least I'd push request logging to DEBUG and possibly some components to INFO.

source/s3_meta_request.c
rccarper: This is coming in the already-approved signing PR.  This PR just needs to go in first.

source/s3_meta_request.c
rccarper: Meta request only maintains a reference to the client while it's being processed by the client.  Once it's finished, it gets rid of the reference.  So if something needs to access the client reference that the meta request keeps track of, we grab a reference to it, just to make sure it doesn't go away out from underneath the caller.

source/s3_meta_request.c
```diff
@@ -83,6 +83,10 @@ struct aws_s3_client *aws_s3_meta_request_acquire_client(struct aws_s3_meta_requ
     client = meta_request->synced_data.client;
 
     if (client != NULL) {
+        AWS_LOGF_TRACE(
+            AWS_LS_S3_META_REQUEST,
+            "id=%p Meta request trying to get reference to client, but client is null.",
```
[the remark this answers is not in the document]
rccarper: The reference to the client goes away when the meta request finishes.  Because of that, it's not invalid to get a NULL client ref back.

But, all places that currently need this, and all places that we should be using it in the future, we could expect that the client is not NULL.  So we could probably make this, and the other existing place that allows a NULL right now, an assert.

source/s3_meta_request.c
rccarper: Changed all traces to debug.

source/s3_meta_request.c
bretambrose: But there would just be one right?

`
if (state == finished) {
   meta_request_already_finished = true
} else {
   request = aws_s3_meta_request_retry_queue_pop_synced(..);
}
`
 and the unlock label goes away

source/s3_meta_request.c
rccarper: You're right, them meta request acquire/release is unnecessary.  I'll get that taken out.

source/s3_meta_request.c
rccarper: Yes, yes it it does.  It's meant to be looking at the error code for the request here, not the error code passed in.  Thanks for catching that!

source/s3_meta_request.c
rccarper: Internally they handle any failures right now.  We might be able to factor that out to this level if that's preferable.

source/s3_meta_request.c
rccarper: You are correct.  I will fix.

source/s3_meta_request.c
rccarper: We could--we should be guaranteed a host header in the passed in message though. :-/

source/s3_meta_request.c
rccarper: The idea was that once the meta request has the request in its queue, it shouldn't prevent clean up of the meta request by keeping a reference to it.

In the refactor, we're actually changing to keep the reference, but just clear out the queue on finish.

source/s3_meta_request.c
rccarper: Going to remove reference counting on requests in a future PR.  For now, to make things a little easier to follow, I moved the acquire into the two functions below (which were both assuming this acquire took place, and release it if they fail internally).

source/s3_meta_request.c
rccarper: This one's going to happen a lot, so keeping this at DEBUG.

bretambrose (approved):

PR #28, opened 2020-12-03
Small bug fixes

rccarper:
*Description of changes:*
* Enforcing min part size.
* Failing loudly for ContentMD5 headers.
* Not requiring Content-Type to be in a PutObject message for CreateMultipartUpload.


source/s3_client.c
```diff
@@ -180,6 +181,16 @@ struct aws_s3_client *aws_s3_client_new(
         return NULL;
     }
 
+    /* Cannot be less than zero.  If zero, use default. */
+    if (client_config->part_size != 0 && client_config->part_size < MB_TO_BYTES(s_s3_min_part_size_MB)) {
```
TingDaoK, line 185: I tried download with small part_size, it was accpeted by the server. So, the min part size seems only applied to upload? Not sure, we should rule it out totally or not.

source/s3_client.c
```diff
@@ -180,6 +181,16 @@ struct aws_s3_client *aws_s3_client_new(
...
+        AWS_LOGF_ERROR(
+            AWS_LS_S3_CLIENT,
+            "Cannot create client from client_config; part size is less than the minimum part size of %d MB.",
```
TingDaoK, line 188: `part size is not zero(default) and less than the minimum part size of %d MB` Or something like that will be better

source/s3_client.c
rccarper: Yeah, I'm on the fence with that too.

source/s3_client.c
rccarper: It's subjective, but I think calling out the default case is redundant in the messaging.

TingDaoK (approved):

PR #29, opened 2020-12-03
Parts refactor

rccarper:
*Description of changes:*
Doing refactoring and improvements:

* Refactored meta request processing flow to have less steps.
* Removed part buffers--now allocated on the fly.  We might re-add something similar in the future.  (If we do, buffers will be given out at the work-process-loop in the client, so that we don't need to worry about not being able to get a part buffer in a meta request.)
* Multipart upload parts are now only read in order.  There is no more stream seeking.
* Multipart uploads no longer query the stream length.  They do require that the Content-Length header is correctly set, however.
* Multipart get body buffers are now passed back in order, and are distributed across a designated event-loop-group for improved performance.
* Removing unnecessary config options.


source/s3_auto_ranged_put.c
```diff
@@ -218,50 +194,15 @@ static int s_s3_auto_ranged_put_next_request(
 
             AWS_FATAL_ASSERT(initial_request_body);
 
-            int64_t request_body_length = 0;
```
rccarper, line 221: We do all of this in the client's meta_request_factory function now.

source/s3_auto_ranged_put.c
```diff
@@ -332,88 +271,44 @@ static int s_s3_auto_ranged_put_next_request(
...
     AWS_PRECONDITION(request);
 
     struct aws_s3_auto_ranged_put *auto_ranged_put = meta_request->impl;
     AWS_PRECONDITION(auto_ranged_put);
 
     struct aws_http_message *message = NULL;
-    struct aws_s3_part_buffer *part_buffer = NULL;
-
-    /* If we're not sending the entire object, then we need to grab a part buffer */
-    if (request->desc_data.request_tag != AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_ENTIRE_OBJECT) {
```
rccarper, line 347: If we need to upload the entire file in one go, then we currently use the default meta request.

source/s3_auto_ranged_put.c
```diff
@@ -605,40 +503,13 @@ static int s_s3_auto_ranged_put_header_block_done(
...
-    } else if (request->desc_data.request_tag == AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_COMPLETE_MULTIPART_UPLOAD) {
-        aws_byte_buf_reset(&request->send_data.part_buffer->buffer, false);
+    } else if (request->request_tag == AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_COMPLETE_MULTIPART_UPLOAD) {
         return AWS_OP_SUCCESS;
     }
 
     return AWS_OP_SUCCESS;
 }
 
-static int s_s3_auto_ranged_put_incoming_body(
```
rccarper, line 616: Appending response body data to the correct buffer is now handled in the base Meta Request type.

source/s3_client.c
```diff
@@ -487,26 +461,16 @@ static struct aws_s3_vip *s_s3_client_vip_new(
...
+/* Releases the memory for a vip structure. */
 static void s_s3_client_vip_destroy(struct aws_s3_vip *vip) {
     AWS_PRECONDITION(vip);
 
+    aws_atomic_store_int(&vip->cleaning_up, 1);
+
     struct aws_s3_client *client = vip->owning_client;
     s_s3_client_lock_synced_data(client);
-
-    while (!aws_linked_list_empty(&vip->vip_connections)) {
```
rccarper, line 497: VIP's no longer hold a list of their VIP connections.  Instead, for clean up, the cleaning_up flag on the VIP will be retrieved in the work event loop, and they will be cleaned up there.

source/s3_auto_ranged_get.c
```diff
@@ -438,57 +386,13 @@ static int s_s3_auto_ranged_get_header_block_done(
     }
 
     if (num_parts > 1) {
-        aws_s3_meta_request_schedule_work(meta_request);
-    }
-
-    return AWS_OP_SUCCESS;
-}
-
-static int s_s3_auto_ranged_get_incoming_body(
```
rccarper, line 447: This is now handled automatically by the base meta request type.

source/s3_auto_ranged_get.c
```diff
@@ -153,25 +137,6 @@ static void s_s3_meta_request_auto_ranged_get_destroy(struct aws_s3_meta_request
     aws_mem_release(meta_request->allocator, auto_ranged_get);
 }
 
-static bool s_s3_auto_ranged_state_has_work(enum aws_s3_auto_ranged_get_state state) {
```
rccarper, line 156: Having work is now implicitly defined by the next_request virtual function: if a meta request can return a request for being sent, then it has work.

source/s3_auto_ranged_put.c
```diff
@@ -332,88 +271,44 @@ static int s_s3_auto_ranged_put_next_request(
...
         case AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_PART: {
 
             bool error_occurred = false;
 
-            s_s3_auto_ranged_put_lock_synced_data(auto_ranged_put);
+            if (is_initial_prepare) {
 
-            uint64_t range_start = (request->desc_data.part_number - 1) * meta_request->part_size;
-            struct aws_input_stream *initial_body_stream = meta_request->synced_data.initial_body_stream;
+                AWS_ASSERT(request->request_body.capacity == 0);
```
justinboswell, line 297: does `request_body.capacity == 0` universally indicate that this is the initial attempt? If so, we could do away with the `is_initial_prepare` flag, right?

source/s3_auto_ranged_put.c
```diff
@@ -458,12 +353,20 @@ static int s_s3_auto_ranged_put_prepare_request(
 
             AWS_FATAL_ASSERT(auto_ranged_put->synced_data.upload_id);
 
+            if (is_initial_prepare) {
+                AWS_ASSERT(request->request_body.capacity == 0)
```
justinboswell, line 357: Same as above

source/s3_auto_ranged_put.c
```diff
@@ -528,7 +427,7 @@ static int s_s3_auto_ranged_put_header_block_done(
     struct aws_allocator *allocator = meta_request->allocator;
     AWS_PRECONDITION(allocator);
 
-    if (request->desc_data.request_tag == AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_CREATE_MULTIPART_UPLOAD) {
+    if (request->request_tag == AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_CREATE_MULTIPART_UPLOAD) {
```
justinboswell, line 430: Oh. Is request_tag really just state machine state/process phase/stage? Should we indicate that in its name/documentation?

source/s3_client.c
```diff
@@ -25,24 +25,33 @@
...
+
+/* TODO Provide analysis on origins of this value. */
+static const double s_throughput_per_vip_gbps = 6.25;
+static const uint32_t s_num_connections_per_vip = 10;
 
 static const uint16_t s_http_port = 80;
 static const uint16_t s_https_port = 443;
 
-static const uint64_t s_default_part_size = 20 * 1024 * 1024;
+static const uint64_t s_default_part_size = 5 * 1024 * 1024;
```
justinboswell, line 49: Should also provide info on these values too

source/s3_client.c
```diff
@@ -930,253 +909,179 @@ static void s_s3_client_process_work_task(struct aws_task *task, void *arg, enum
...
 
+    /*******************/
+    /* Step 1: Move everything into thread local memory. */
+    /*******************/
     s_s3_client_lock_synced_data(client);
 
     /* Once we exit this mutex, someone can reschedule this task. */
     client->synced_data.process_work_task_scheduled = false;
 
+    aws_linked_list_swap_contents(&meta_request_work_list, &client->synced_data.pending_meta_request_work);
```
justinboswell, line 929: Should we add a precondition that the `pending_meta_request_work` list is empty?

source/s3_client.c
```diff
@@ -930,253 +909,179 @@ static void s_s3_client_process_work_task(struct aws_task *task, void *arg, enum
...
-
-        aws_linked_list_push_back(&work_destroy_list, &work->node);
-    }
+    s_s3_client_unlock_synced_data(client);
 
     /*******************/
-    /* Step 3: Check all of the incoming VIP connections for meta requests.  If they already have a meta request, that
-     * means they just finished a request, so we check for remaining work.  If there isn't any remaining work, then we
-     * remove that meta request from our list. */
+    /* Step 2: Process any meta request work. */
```
justinboswell, line 938: This section is much easier to follow that the old one, excellent!

source/s3_client.c
```diff
@@ -930,253 +909,179 @@ static void s_s3_client_process_work_task(struct aws_task *task, void *arg, enum
...
-            vip_connection->threaded_data.meta_request,
-            vip_connection,
-            s_s3_client_vip_connection_request_finished,
-            vip_connection);
+            /* Grab the next request from the meta request. */
+            request = aws_s3_meta_request_next_request(current_meta_request);
 
-        aws_linked_list_push_back(&work_destroy_list, &work->node);
-    }
+            /* If the meta request is null, then this meta request is currently out requests, so remove it. */
```
justinboswell, line 1034: out of

source/s3_client.c
```diff
@@ -930,253 +909,179 @@ static void s_s3_client_process_work_task(struct aws_task *task, void *arg, enum
...
 
-        /* If we still don't have a meta request node, there's nothing for this VIP connection to do. */
-        if (next_meta_request_node == NULL) {
-            AWS_ASSERT(vip_connection->threaded_data.meta_request == NULL);
-            aws_linked_list_push_back(&client->threaded_data.idle_vip_connections, &work->node);
+        AWS_ASSERT(vip_connection->request == NULL);
+
+        /* If we don't have any meta requests or how many requests we have in flight is over the maximum allowed, put
+         * the vip connection into the idle list. */
+        if (aws_linked_list_empty(&client->threaded_data.meta_requests) ||
```
justinboswell, line 1009: How big is this list? Could we use an array_list of pointers instead? Looping over this list twice in this function to count it is intentionally inducing likely cache misses just to produce a count. Might create a death by paper cut situation with enough requests in flight.

source/s3_client.c
```diff
@@ -1252,26 +1167,122 @@ static void s_s3_client_vip_connection_on_acquire_request_connection(
...
+    struct s3_streaming_body_payload *payload =
+        aws_mem_calloc(client->sba_allocator, 1, sizeof(struct s3_streaming_body_payload));
+
+    s_s3_client_internal_acquire(client);
+    payload->client = client;
+
+    aws_linked_list_init(&payload->requests);
+    aws_linked_list_move_all_back(&payload->requests, requests);
+
+    aws_task_init(&payload->task, s_s3_client_body_streaming_task, payload, "s3_client_body_streaming_task");
```
justinboswell, line 1250: Oh this is interesting, this is going to be the bottleneck on PUT most likely. Looks like you've already moved it off the main I/O event loop, so that might be all the mitigation we can do right now. Let's keep an eye on this when we test puts.

source/s3_default_meta_request.c
```diff
@@ -18,6 +18,9 @@ enum aws_s3_meta_request_default_state {
 struct aws_s3_meta_request_default {
     struct aws_s3_meta_request base;
 
+    bool is_get_request;
```
justinboswell, line 21: Isn't there already a request type?

justinboswell: Also, looking at how this is used, and request_type for that manner, why don't we just set the vtable up differently based on the request type when we initialize it, and then never branch?

include/aws/s3/private/s3_client_impl.h
```diff
@@ -116,9 +92,14 @@ struct aws_s3_client {
...
     struct aws_ref_count internal_ref_count;
 
+    /* Client bootstrap for setting up connection managers. */
     struct aws_client_bootstrap *client_bootstrap;
 
-    struct aws_event_loop *event_loop;
+    /* Event loop on the client bootstrap ELG for processing work/dispatching requests. */
+    struct aws_event_loop *process_work_event_loop;
+
+    /* Event loop group for streaming request bodies back to the user. */
```
justinboswell, line 101: This is actually from/to the user, right?

include/aws/s3/private/s3_util.h
```diff
@@ -36,6 +36,8 @@ extern const struct aws_byte_cursor g_accept_ranges_header_name;
 extern const struct aws_byte_cursor g_etag_header_name;
 extern const struct aws_byte_cursor g_post_method;
 
+extern const uint64_t g_max_num_upload_parts;
```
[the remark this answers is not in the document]
rccarper: This one's a service side constant.  Should probably have s3 in the title.

source/s3_auto_ranged_put.c
rccarper: It's what I originally had, but it felt a bit hacky to me, because you're having to intuit if this is the first prepare.  (Not everything will need to set up a request body either.)

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -58,26 +62,29 @@ struct aws_s3_request {
...
+    uint32_t part_number;
 
-        /* Tag that defines what the built request will actually consist of.  Request tags are different per meta
-         * request type, and do not necessarily map 1:1 with actual S3 API requests.  For example, they can be more
-         * contextual, like "first part" instead of just "part".) */
-        int request_tag;
+    /* Tag that defines what the built request will actually consist of.  Request tags are different per meta
+     * request type, and do not necessarily map 1:1 with actual S3 API requests.  For example, they can be more
+     * contextual, like "first part" instead of just "part".) */
+    int request_tag;
```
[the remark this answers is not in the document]
rccarper: It's basically an enum defined by the derived class.  We could possibly combine it with the bit field.

source/s3_auto_ranged_put.c
rccarper: Not really state--it's an enum that is defined by the derived class type that indicates the request type.  (Like create-multipart-upload for example.)  I wanted to allow them to include context where that's convenient (like *first*_ranged_get) so they don't map 1:1, which is why the aren't a global "s3 request types" enum.

source/s3_client.c
rccarper: I don't think so?  We're trying to grab everything in pending_meta_requests_work so that we can use it in the thread.

source/s3_client.c
rccarper: The body streaming stuff is only for gets.  We do the stream read for uploads in the request prepare.  (I was looking at doing the stream read in a separate thread, but walked it back when NUMA concerns were flagged.)

include/aws/s3/private/s3_client_impl.h
rccarper: Not really--the first event loop just handles scheduling of work.  The body streaming ELG actually pushes stuff back to the user.

source/s3_client.c
rccarper: Woops, the loop that always goes through the entire list was just for logging--I need to take that out.  The list is for all of the meta requests being processed.

This part is just looking for the next meta request that has requests, which we only do for the number of vip connections that need work.  It is possible that all of the meta requests in the list ran out of work at once, in which case we would go through the entire list.  We could change it over to an arraylist.

source/s3_default_meta_request.c
rccarper: We could do that!

include/aws/s3/private/s3_client_impl.h
```diff
@@ -30,6 +29,10 @@ typedef void(aws_s3_client_sign_callback)(int error_code, void *user_data);
 struct aws_s3_vip {
     struct aws_linked_list_node node;
 
+    /* True if this VIP is currently being cleaned up.  The work event loop will check for this flag and clean up
+     * related VIP connections. */
+    struct aws_atomic_var cleaning_up;
```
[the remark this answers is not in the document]
rccarper: We could call it active!

source/s3_client.c
rccarper: :-D

source/s3_client.c
justinboswell: If that's all, don't sweat it, I just didn't want this to be on the hot path.

include/aws/s3/private/s3_client_impl.h
```diff
@@ -166,29 +142,26 @@ struct aws_s3_client {
...
 
+        /* Counter for number of requests that have been finished/released, allowing us to create new requests. */
+        uint32_t pending_request_count;
+
         /* Host listener to get new IP addresses. */
         struct aws_host_listener *host_listener;
 
         /* Whether or not work processing is currently scheduled. */
-        bool process_work_task_scheduled;
+        uint32_t process_work_task_scheduled : 1;
```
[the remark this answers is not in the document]
rccarper: Maybe not?  It's just a boolean to know if the process work task was already scheduled since we only need it scheduled once.

source/s3_auto_ranged_put.c
rccarper: Changed the documentation a little bit to be more clear.

source/s3_client.c
rccarper: Added a TODO for now.

source/s3_default_meta_request.c
rccarper: Actually, we only do one branch on this right now to add an additional flag for the request.  Was wondering if this might be overkill? :-/  (Although I do like the idea.)

include/aws/s3/private/s3_meta_request_impl.h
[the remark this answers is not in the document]
rccarper: Put a TODO in for now.

justinboswell (approved):

PR #34, opened 2020-12-16
Default to tls-enabled with default tls context fallback

bretambrose:
* s3 client now detaults to tls enabled and uses default tls options if none provided
* when using tls, server name will be automatically set based on the endpoint the client is configured with


source/s3_client.c
```diff
@@ -213,12 +213,31 @@ struct aws_s3_client *aws_s3_client_new(
...
+
+            struct aws_tls_ctx *default_tls_ctx = aws_tls_client_ctx_new(allocator, &default_tls_ctx_options);
+            if (default_tls_ctx == NULL) {
+                goto on_error;
+            }
+
+            aws_tls_connection_options_init_from_ctx(client->tls_connection_options, default_tls_ctx);
 
-        aws_tls_connection_options_copy(client->tls_connection_options, client_config->tls_connection_options);
+            aws_tls_ctx_release(default_tls_ctx);
```
justinboswell, line 238: How is this going to work at connection time? The aws_tls_ctx must exist for TLS to work. We could cache this on the client, but it has to live doesn't it?

bretambrose: It's ref counted by the tls connection options that use it.

justinboswell: OH right ok.

justinboswell (approved):

PR #43, opened 2020-12-24
sse support

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/s3_request_messages.c
```diff
@@ -393,6 +393,17 @@ struct aws_http_message *aws_s3_message_util_copy_http_message(
         if (aws_http_message_add_header(message, header)) {
             goto error_clean_up;
         }
+
+        /* For SSE upload, the sse related headers should only be shown in the create_multipart_upload. which means it
+         * should only be copied once.
+         * TODO: should be all the headers begin with x-amz-server-side-encryption-* ? */
+        if (aws_byte_cursor_eq_c_str_ignore_case(&header.name, "x-amz-server-side-encryption")) {
```
rccarper, line 400: This functionality should be opt-in and not touch the base_message.  (base_message should already be const for this function.  That's my mistake!)  What I think we want to do differently is just not copy these headers to the new message if the caller passes in a flag to this function.  And then we can pass in that flag where ever we need to in any of the existing places where we use this function.

tests/s3_data_plane_tests.c
```diff
@@ -398,6 +400,90 @@ static int s_test_s3_get_object_multiple(struct aws_allocator *allocator, void *
     return 0;
 }
 
+int test_s3_get_object_sse_helper(struct aws_allocator *allocator, enum AWS_S3_TESTER_SSE_TYPE sse_type) {
+
+    const struct aws_byte_cursor test_object_path =
```
rccarper, line 405: Would it be possible to do something like you did for aws_s3_tester_send_put_object_meta_request but for aws_s3_tester_send_get_object_meta_request?  Feels like a big copy paste for just the if at the top. :)

include/aws/s3/private/s3_request_messages.h
```diff
@@ -56,7 +57,8 @@ struct aws_http_message *aws_s3_complete_multipart_message_new(
 
 struct aws_http_message *aws_s3_message_util_copy_http_message(
     struct aws_allocator *allocator,
-    struct aws_http_message *message);
+    struct aws_http_message *message,
+    bool sse);
```
rccarper, line 61: Missed this before!  Not important right now, but maybe add a TODO to break out the sse boolean into a flag.  It'll make it more clear when you're using the function (although I see that you annotated it everywhere, which is awesome), and it'll let us add other options in the future.

tests/s3_data_plane_tests.c
```diff
@@ -102,7 +102,11 @@ enum aws_s3_client_tls_usage {
     AWS_S3_TLS_DISABLED,
 };
 
-static int s_test_s3_get_object_helper(struct aws_allocator *allocator, enum aws_s3_client_tls_usage tls_usage) {
+static int s_test_s3_get_object_helper(
+    struct aws_allocator *allocator,
+    enum aws_s3_client_tls_usage tls_usage,
+    struct aws_byte_cursor s3_path,
+    enum AWS_S3_TESTER_SSE_TYPE sse_type) {
```
rccarper, line 109: Could we make this enum lower case to match the other enums, and maybe move the argument to be next to the tls_usage?

rccarper (approved):

PR #48, opened 2020-12-28
Support content-md5 header.

wps132230:
*Issue:*
If the original message has a `content-md5` header, then for multipart uploads we need to calculate an MD5 for the part.

*Description of changes:*
1. Remove `content-md5` header for `CreateMultipartUpload` and `CompleteMultipartUpload`
2. Recalculate and update `content-md5` header for `UploadPart`


source/s3_request_messages.c
```diff
@@ -199,6 +223,10 @@ struct aws_http_message *aws_s3_complete_multipart_message_new(
 
     aws_http_headers_erase(headers, g_content_type_header_name);
 
+    if (aws_http_headers_has(headers, g_content_md5_header_name)) {
+        aws_http_headers_erase(headers, g_content_md5_header_name);
```
wps132230, line 227: Should I have
```c++
if (aws_http_headers_erase(headers, g_content_md5_header_name)) {
    goto error_clean_up;
}
```
here?

Is there any reason why we don't check the result of `aws_http_headers_erase(headers, g_content_length_header_name)`?

rccarper: Because the header not being found returns an error and then it early outs in valid cases.  I guess it might be more correct for us to check the error code and early out if it's != HEADER_NOT_FOUND.

source/s3_request_messages.c
```diff
@@ -443,6 +477,53 @@ static int s_s3_message_util_add_content_range_header(
...
+    struct aws_allocator *allocator,
+    struct aws_byte_buf *byte_buf,
+    struct aws_http_message *out_message) {
+
+    AWS_PRECONDITION(out_message);
+
+    // Compute MD5
+    struct aws_byte_cursor md5_input = aws_byte_cursor_from_buf(byte_buf);
+    uint8_t md5_output[AWS_MD5_LEN] = {0};
+    struct aws_byte_buf md5_output_buf = aws_byte_buf_from_array(md5_output, sizeof(md5_output));
```
rccarper, line 491: I think you can use aws_byte_buf_from_empty_array here and then you don't have to reset the len variable.

source/s3_request_messages.c
```diff
@@ -443,6 +477,53 @@ static int s_s3_message_util_add_content_range_header(
...
+static int s_s3_message_util_add_content_md5_header(
+    struct aws_allocator *allocator,
+    struct aws_byte_buf *byte_buf,
+    struct aws_http_message *out_message) {
+
+    AWS_PRECONDITION(out_message);
+
+    // Compute MD5
+    struct aws_byte_cursor md5_input = aws_byte_cursor_from_buf(byte_buf);
+    uint8_t md5_output[AWS_MD5_LEN] = {0};
```
rccarper, line 490: Because the buffer will start with length 0, you probably don't have to zero it out.

source/s3_request_messages.c
```diff
@@ -443,6 +477,53 @@ static int s_s3_message_util_add_content_range_header(
...
 
+/* Add a content-md5 header. */
+static int s_s3_message_util_add_content_md5_header(
+    struct aws_allocator *allocator,
+    struct aws_byte_buf *byte_buf,
+    struct aws_http_message *out_message) {
+
+    AWS_PRECONDITION(out_message);
+
+    // Compute MD5
```
rccarper, line 488: Trivial: could we do all c-style comments just to keep everything consistent? :)

source/s3_request_messages.c
```diff
@@ -443,6 +477,53 @@ static int s_s3_message_util_add_content_range_header(
...
+    struct aws_byte_cursor base64_input = aws_byte_cursor_from_buf(&md5_output_buf);
+    size_t base64_output_size = 0;
+    if (aws_base64_compute_encoded_len(md5_output_buf.len, &base64_output_size)) {
+        return AWS_OP_ERR;
+    }
+    struct aws_byte_buf allocation;
+    if (aws_byte_buf_init(&allocation, allocator, base64_output_size + 2)) {
+        return AWS_OP_ERR;
+    }
+    memset(allocation.buffer, 0xdd, allocation.capacity);
```
rccarper, line 507: Is this needed for the base 64 or is it more for debugging?

rccarper (commented):
This is awesome!  Some minor comments, and then a couple of notes:

* I think right now we're checking to see if the user supplied a content MD5 header before computing one.  Could we add an option to the client (defaulted to "on") and do the MD5 in that case instead?  
* One case I did forget to call out was for the default meta request.  We use it if the upload is less than a part size right now: https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L868  We'll need to compute it in that case too.

source/s3_request_messages.c
```diff
@@ -353,6 +391,78 @@ struct aws_input_stream *aws_s3_message_util_assign_body(
...
+    size_t base64_output_size = 0;
+    if (aws_base64_compute_encoded_len(md5_output_buf.len, &base64_output_size)) {
+        return AWS_OP_ERR;
+    }
+    struct aws_byte_buf allocation;
+    if (aws_byte_buf_init(&allocation, allocator, base64_output_size + 2)) {
+        return AWS_OP_ERR;
+    }
+    struct aws_byte_buf base64_output_buf = aws_byte_buf_from_empty_array(allocation.buffer + 1, base64_output_size);
+    if (aws_base64_encode(&base64_input, &base64_output_buf)) {
```
rccarper, line 448: Why the two buffers? :)

source/s3_default_meta_request.c
```diff
@@ -216,6 +216,14 @@ static int s_s3_meta_request_default_prepare_request(
         aws_s3_meta_request_read_body(meta_request, &request->request_body);
     }
 
+    struct aws_http_headers *initial_request_headers =
+        aws_http_message_get_headers(meta_request->initial_request_message);
+
+    if (client->compute_content_md5 == AWS_MR_CONTENT_MD5_ENABLED &&
+        !aws_http_headers_has(initial_request_headers, g_content_md5_header_name)) {
```
rccarper, line 223: So the default meta request is sort of a tricky case because it's a catch-all for all S3 requests that are not gets or put.  As it is right now, I think we'll add content-md5 headers for a lot of cases where we shouldn't.

I think we should add a flag to the default meta request for if it's going to add the content-md5 header or not, and have that be specified through the default_meta_request_new function.  Then I think we can basically just move the logic for setting the flag here: https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L996

source/s3_auto_ranged_put.c
```diff
@@ -313,13 +313,22 @@ static int s_s3_auto_ranged_put_prepare_request(
                 goto message_create_failed;
             }
 
+            bool should_compute_content_md5 = false;
+            struct aws_http_headers *initial_request_headers =
+                aws_http_message_get_headers(meta_request->initial_request_message);
+            if (client->compute_content_md5 == AWS_MR_CONTENT_MD5_ENABLED ||
+                aws_http_headers_has(initial_request_headers, g_content_md5_header_name)) {
+                should_compute_content_md5 = true;
```
rccarper, line 321: Trivial: could just set the should_compute_content_md5 from the condition.

source/s3_request_messages.c
JonathanHenson: You could could just assert that if there’s an error code it’s because the header wasn’t found

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -189,6 +189,8 @@ struct aws_s3_meta_request {
     /* Part size to use for uploads and downloads.  Passed down by the creating client. */
     const size_t part_size;
```
JonathanHenson, line 191: Maybe put a book member somewhere it’s going to not add extra padding?

source/s3_default_meta_request.c
```diff
@@ -216,6 +218,10 @@ static int s_s3_meta_request_default_prepare_request(
         aws_s3_meta_request_read_body(meta_request, &request->request_body);
     }
 
+    if (meta_request->should_compute_content_md5) {
+        aws_s3_message_util_add_content_md5_header(meta_request->allocator, &request->request_body, message);
```
JonathanHenson, line 222: Is this being done on the calling thread? If so that’s likely to throttle the caller at the speed we can perform an md5 (slow) on a single core.

source/s3_request_messages.c
```diff
@@ -353,6 +384,77 @@ struct aws_input_stream *aws_s3_message_util_assign_body(
...
+    uint64_t part_size,
+    struct aws_http_message *out_message) {
+    AWS_PRECONDITION(out_message);
+
+    uint64_t range_start = part_index * part_size;
+    uint64_t range_end = range_start + part_size - 1;
+
+    /* TODO this is more than enough space, but maybe there's a better way to do this?
+     * ((2^64)-1 = 20 characters;  2*20 + length-of("bytes=-") < 128) */
+    char range_value_buffer[128] = "";
```
JonathanHenson, line 399: We have a byte buffer append Api which is safer

include/aws/s3/private/s3_meta_request_impl.h
rccarper: Could you elaborate more on that?  (This can change per meta request if the part size is too small for the maximum number of parts allowed by the service.)

source/s3_default_meta_request.c
rccarper: This step does get done in the client bootstrap ELG right now, but we have a task (which we have done some work on) to make this step of the meta request (the prepare) optionally happen async/different ELG/thread.  tldr: this is the best place to do the md5.

source/s3_request_messages.c
rccarper: I guess we could use appends for some of the string parts, but I think we still need to access the buffer directly for the uint64 -> string formatting?

include/aws/s3/private/s3_meta_request_impl.h
wps132230: Sorry.. but what is a book member?

source/s3_meta_request.c
```diff
@@ -102,7 +103,8 @@ int aws_s3_meta_request_init_base(
     /* Set up reference count. */
     aws_ref_count_init(&meta_request->ref_count, meta_request, s_s3_meta_request_destroy);
 
-    *((size_t *)&meta_request->part_size) = part_size;
+    *((uint64_t *)&meta_request->part_size) = part_size;
```
rccarper, line 106: This should be a size_t* type cast.

include/aws/s3/private/s3_meta_request_impl.h
rccarper: To Henson's point, let's make this the start of a bitfield at the end of the structure.

rccarper (approved):

PR #52, opened 2020-12-30
Improved retry logic

rccarper:
*Post Review Update:*
* Re-worked the acquiring of the token so that it always happens before a request is being tried for the first time.
* There are some test gaps here which I'll be trying to fill in soon with a separate PR.

*Description of changes:*
* Changing retry logic so that retries now happen at the vip_connection level and are not tied to just a request.  Previously, single requests would back-off, but the actual connection could be re-used immediately to process a different request.  This change makes it so that the connection and any request currently tied to it back-off together.
* This removes the need for a "retry queue"of requests on the meta request, so that logic has been removed.  Similarly, the function aws_s3_meta_request_handle_error for initiating a retry inside the meta request wasn't really necessary anymore and was factored out.


source/s3_client.c
```diff
@@ -1482,39 +1486,234 @@ static void s_s3_client_on_acquire_http_connection(
...
+
+    return;
+
+error_clean_up:
+
+    request->send_data.error_code = error_code;
+    aws_s3_client_notify_connection_finished(client, vip_connection, 0);
+}
+
+static void s_s3_client_retry_ready(struct aws_retry_token *token, int error_code, void *user_data) {
```
justinboswell, line 1669: Isn't this more `retry_complete` than `retry_ready`?

justinboswell (approved):
Fix-n-ship

source/s3_client.c
JonathanHenson: no it's retry ready, it means, "retry now please"

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -59,9 +59,6 @@ struct aws_s3_request {
     /* Owning meta request. */
     struct aws_s3_meta_request *meta_request;
 
-    /* Current retry token for the request. If it has never been retried, this will be NULL. */
```
JonathanHenson, line 62: did this behavior change? you need a token for every request, or you'll never recover from an outage.

JonathanHenson: meaning, you get a token even if it's the first request, and then use it for retries if you need to. otherwise, release it when you're finished.

JonathanHenson (changes requested):

include/aws/s3/private/s3_meta_request_impl.h
rccarper: The retry token will still be one per request, but now it's attached to the owning aws_s3_vip_connection that the request is being done on.  This is so that we can back off the actual connection that's doing the request, and also have an easier way to do a retry for an error when just trying to make the connection.

rccarper: Re-worked to always try grabbing a token when a request begins.

JonathanHenson (approved):

PR #53, opened 2020-12-30
Cancel functionality

rccarper:
*Issue #, if available:*

*Description of changes:*


source/s3_meta_request.c
```diff
@@ -212,6 +211,21 @@ int aws_s3_meta_request_init_base(
     return AWS_OP_SUCCESS;
 }
 
+void aws_s3_meta_request_cancel(struct aws_s3_meta_request *meta_request) {
+    AWS_PRECONDITION(meta_request);
+    AWS_PRECONDITION(meta_request->vtable->cancel);
+    bool cancel_called = true;
+    aws_s3_meta_request_lock_synced_data(meta_request);
```
TingDaoK, line 218: hmmm, need something better. The lock here is a little dangerous. But nothing pop in my head.

source/s3_auto_ranged_put.c
```diff
@@ -174,6 +191,45 @@ static void s_s3_meta_request_auto_ranged_put_destroy(struct aws_s3_meta_request
...
+    struct aws_s3_auto_ranged_put *auto_ranged_put = meta_request->impl;
+    bool cancel = false;
+    s_s3_auto_ranged_put_lock_synced_data(auto_ranged_put);
+    cancel = request->request_tag == AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_ABORT_MULTIPART_UPLOAD;
+    s_s3_auto_ranged_put_unlock_synced_data(auto_ranged_put);
+    if (cancel) {
+        /* TODO: What's the error code here? */
+        aws_s3_meta_request_cancel_default(
+            meta_request, auto_ranged_put->synced_data.failed_request, AWS_ERROR_S3_CANCELED_SUCCESS);
+    }
```
TingDaoK, line 230: how to propagate the error happens from cancel or failed request to here? @rccarper

source/s3_auto_ranged_put.c
```diff
@@ -174,6 +191,43 @@ static void s_s3_meta_request_auto_ranged_put_destroy(struct aws_s3_meta_request
...
+static void s_s3_auto_ranged_put_cancel(
+    struct aws_s3_meta_request *meta_request,
+    struct aws_s3_request *failed_request,
+    int error_code) {
+    // TODO: error code needs to be propagated
+    (void)error_code;
+    struct aws_s3_auto_ranged_put *auto_ranged_put = meta_request->impl;
+
+    s_s3_auto_ranged_put_lock_synced_data(auto_ranged_put);
+    if (auto_ranged_put->synced_data.state != AWS_S3_AUTO_RANGED_PUT_STATE_CANCEL &&
```
rccarper, line 203: Once this becomes a s_s3_auto_ranged_put_finish function, let's actually just always set the cancel state here, and then call aws_s3_meta_request_push_to_client.

rccarper (commented):
Thanks for taking my very rough notes and making something functional out if it!  Some comments here--there are several things that I probably should have caught in my initial notes, so sorry for that.

Based on the chat we had yesterday, let's try changing the virtual cancel to a virtual finish function.  Then we'll call aws_s3_meta_request_finish everywhere internally.  The new public-facing cancel function will call the aws_s3_meta_request_finish function with the cancel error code.  For the auto-ranged-put override, it'll do the logic you already have and then call the base finish function once the cancel operation is finished.

source/s3_auto_ranged_put.c
```diff
@@ -174,6 +136,19 @@ static void s_s3_meta_request_auto_ranged_put_destroy(struct aws_s3_meta_request
...
+    struct aws_s3_meta_request *meta_request,
+    struct aws_s3_request *failed_request,
+    int error_code) {
+    struct aws_s3_auto_ranged_put *auto_ranged_put = meta_request->impl;
+    s_s3_auto_ranged_put_lock_synced_data(auto_ranged_put);
+    auto_ranged_put->synced_data.error_code = error_code;
+    auto_ranged_put->synced_data.failed_request = failed_request;
+    s_s3_auto_ranged_put_unlock_synced_data(auto_ranged_put);
+    /* state of meta request has been set now, and the state of auto ranged put will set properly by the task */
+    aws_s3_meta_request_push_to_client(meta_request);
```
TingDaoK, line 149: lost at all the task in client... But, what I see the reason for meta request not get cleaned up is this line will push another task to client, and require another refcount on meta request, but the client does not really process the meta request twice. Probably because of this condition. @rccarper 
https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L1292-L1296

source/s3_meta_request.c
```diff
@@ -211,6 +211,21 @@ int aws_s3_meta_request_init_base(
...
+    AWS_PRECONDITION(meta_request->vtable->finish);
+
+    aws_s3_meta_request_lock_synced_data(meta_request);
+    bool active = meta_request->synced_data.state == AWS_S3_META_REQUEST_STATE_ACTIVE;
+    if (active) {
+        meta_request->synced_data.state = AWS_S3_META_REQUEST_STATE_CANCELLING;
+    }
+    aws_s3_meta_request_unlock_synced_data(meta_request);
+    if (active) {
+        meta_request->vtable->finish(meta_request, NULL, AWS_ERROR_S3_CANCELED);
```
rccarper, line 225: This should just be able to call aws_s3_meta_request_finish

source/s3_meta_request.c
```diff
@@ -868,6 +883,18 @@ static void s_s3_meta_request_send_request_finish(
     vtable->send_request_finish(vip_connection, stream, error_code);
 }
 
+void aws_s3_meta_request_finish_default(
+    struct aws_s3_meta_request *meta_request,
+    struct aws_s3_request *failed_request,
+    int error_code) {
+
+    if (failed_request != NULL) {
+        aws_s3_meta_request_finish(meta_request, failed_request, failed_request->send_data.response_status, error_code);
```
rccarper, line 892: Contents of the aws_s3_meta_request_finish should be able to become to the default implementation.  And then the status code is used for more than just the failed request, so we can't hide it from the interface.  (Let me know if that becomes an awkward issue!  We can probably make an improvement there, but I think it might be easier for this change if we just leave it as is.)

source/s3_auto_ranged_put.c
```diff
@@ -211,7 +210,8 @@ static int s_s3_auto_ranged_put_next_request(
         case AWS_S3_AUTO_RANGED_PUT_STATE_SENDING_PARTS: {
 
             /* Keep setting up to send parts until we've sent all of them at least once. */
-            if (auto_ranged_put->synced_data.num_parts_sent < auto_ranged_put->synced_data.total_num_parts) {
+            if (auto_ranged_put->synced_data.num_parts_sent < auto_ranged_put->synced_data.total_num_parts &&
```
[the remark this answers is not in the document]
TingDaoK: ha, I found that as well. It's removed from the following PR. https://github.com/awslabs/aws-c-s3/pull/64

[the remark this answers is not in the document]
TingDaoK: Never mind, thought was the same one🤪

source/s3_auto_ranged_put.c
```diff
@@ -229,14 +229,36 @@ static int s_s3_auto_ranged_put_next_request(
...
+                    auto_ranged_put->synced_data.state = AWS_S3_AUTO_RANGED_PUT_STATE_WAITING_FOR_CANCEL;
+
+                    break;
+                }
+                request = aws_s3_request_new(
+                    meta_request,
+                    AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_ABORT_MULTIPART_UPLOAD,
+                    0,
+                    AWS_S3_REQUEST_DESC_RECORD_RESPONSE_HEADERS);
+                /* TODO: which state it should be??? */
```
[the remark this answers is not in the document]
TingDaoK: Just feel like the different waiting states basically doing the same thing🤔

source/s3_client.c
```diff
@@ -1883,12 +1886,26 @@ static void s_s3_client_body_streaming_task(struct aws_task *task, void *arg, en
     while (!aws_linked_list_empty(&payload->requests)) {
         struct aws_linked_list_node *request_node = aws_linked_list_pop_front(&payload->requests);
         struct aws_s3_request *request = AWS_CONTAINER_OF(request_node, struct aws_s3_request, node);
+        struct aws_s3_meta_request *meta_request = request->meta_request;
+
+        aws_s3_meta_request_lock_synced_data(meta_request);
```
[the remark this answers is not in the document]
TingDaoK: I don't think it's worth to be a function? Since, the lock we are acquiring. The check can happen inside a lock block, and a function will make it a little bit awkward.

source/s3_meta_request.c
```diff
@@ -953,7 +980,7 @@ void aws_s3_meta_request_send_request_finish_default(
     } else {
         /* If the request failed due to an invalid, ie, unrecoverable, response status, then finish the meta request
          * with that request as the failing request. */
-        if (error_code == AWS_ERROR_S3_INVALID_RESPONSE_STATUS) {
+        if (error_code == AWS_ERROR_S3_INVALID_RESPONSE_STATUS || error_code == AWS_ERROR_S3_CANCELED) {
```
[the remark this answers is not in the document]
TingDaoK: I probably hit this before, and thought it's better to not retry for user requested cancel. Removed.

source/s3_auto_ranged_put.c
```diff
@@ -528,103 +551,134 @@ static int s_s3_auto_ranged_put_stream_complete(
...
+        }
+        case AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_PART: {
 
-        s_s3_auto_ranged_put_lock_synced_data(auto_ranged_put);
-        ++auto_ranged_put->synced_data.num_parts_completed;
+            bool notify_work_available = false;
+            aws_s3_meta_request_lock_synced_data(meta_request);
+            bool cancelling = meta_request->synced_data.state == AWS_S3_META_REQUEST_STATE_CANCELLING;
+            aws_s3_meta_request_unlock_synced_data(meta_request);
+            s_s3_auto_ranged_put_lock_synced_data(auto_ranged_put);
```
rccarper, line 590: Same as above--don't actually have to re-lock.

source/s3_client.c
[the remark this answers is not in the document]
rccarper: In this case, I think moving the lock and state check to a small function would "encapsulate" it a bit.  Doing checks on the state inside of the meta request are totally fine, but because we've left that area of code, it might be kind of nice to bundle this up so other places could also poll if a meta request is active.

source/s3_meta_request.c
```diff
@@ -868,6 +874,17 @@ static void s_s3_meta_request_send_request_finish(
     vtable->send_request_finish(vip_connection, stream, error_code);
 }
 
+void aws_s3_meta_request_finish(
+    struct aws_s3_meta_request *meta_request,
+    struct aws_s3_request *failed_request,
+    int response_status,
+    int error_code) {
+    AWS_PRECONDITION(meta_request);
+    AWS_PRECONDITION(meta_request->vtable->finish);
```
rccarper, line 883: Trivial: add precondition for the vtable.  (Highly unlikely, I know.)

tests/s3_data_plane_tests.c
```diff
@@ -889,6 +889,312 @@ static int s_test_s3_existing_host_entry(struct aws_allocator *allocator, void *
...
+    AWS_PRECONDITION(meta_request);
+    AWS_PRECONDITION(out_request);
+
+    struct aws_s3_request *request = NULL;
+    struct aws_s3_auto_ranged_put *auto_ranged_put = meta_request->impl;
+
+    int result = AWS_OP_SUCCESS;
+
+    /* TODO: haven't dived into it, but assuming the finish state has been handled well */
+    aws_s3_meta_request_lock_synced_data(meta_request);
```
rccarper, line 904: It's a little cumbersome, but you should be able to re-run all of the original logic by looking up the original vtable:

 struct aws_s3_meta_request_test_results *results = meta_request->user_data;
 struct aws_s3_tester *tester = results->tester;
 struct aws_s3_meta_request_vtable *original_meta_request_vtable =
        aws_s3_tester_get_meta_request_vtable_patch(tester, 0)->original_vtable;

(Where 0 is the index of the vtable that has been patched.  If you are only patching one vtable, this can just be 0.)

tests/s3_tester.c
```diff
@@ -785,6 +787,13 @@ int aws_s3_tester_send_meta_request(
 
     ASSERT_TRUE(meta_request != NULL);
 
+    if (flags & AWS_S3_TESTER_SEND_META_REQUEST_CANCEL) {
```
rccarper, line 790: Could we do something more deterministic here?  For example, maybe a cancel based on the aws_s3_auto_ranged_put_state?

source/s3_auto_ranged_put.c
```diff
@@ -174,6 +145,28 @@ static void s_s3_meta_request_auto_ranged_put_destroy(struct aws_s3_meta_request
...
+    struct aws_s3_auto_ranged_put *auto_ranged_put = meta_request->impl;
+    aws_s3_meta_request_lock_synced_data(meta_request);
+    bool active = meta_request->synced_data.state == AWS_S3_META_REQUEST_STATE_ACTIVE;
+    if (active) {
+        meta_request->synced_data.state = AWS_S3_META_REQUEST_STATE_CANCELLING;
+    }
+    aws_s3_meta_request_unlock_synced_data(meta_request);
+    s_s3_auto_ranged_put_lock_synced_data(auto_ranged_put);
+    auto_ranged_put->synced_data.error_code = error_code;
+    auto_ranged_put->synced_data.failed_request = failed_request;
```
rccarper, line 163: These are currently ref counted right now.  (Arguably shouldn't be, and will be trying to take that out eventually.)  Need an aws_s3_request_acquire here, and then a release when you get rid of it.

tests/s3_tester.c
TingDaoK: Yeah, I don't like the testing here, but probably will just add TODOs for the proper testing.

source/s3_request_messages.c
```diff
@@ -394,8 +421,19 @@ struct aws_http_message *aws_s3_message_util_copy_http_message(
...
+                if (aws_http_message_add_header(message, header)) {
+                    goto error_clean_up;
+                }
+            }
+            continue;
+        }
+
         /* For SSE upload, the sse related headers should only be shown in the create_multipart_upload.*/
-        if (aws_byte_cursor_eq_c_str_ignore_case(&header.name, "x-amz-server-side-encryption") && !sse) {
+        if (aws_byte_cursor_eq_c_str_ignore_case(&header.name, "x-amz-server-side-encryption") && !sse_included) {
```
rccarper, line 436: Trivial: do the !sse_included check first since it's faster.

tests/s3_data_plane_tests.c
rccarper: If we need to do this in a next pass, can we at least update the copy pasta and add a TODO? :)

source/s3_auto_ranged_put.c
rccarper: Looking at this with more awake eyes, you do need an acquire here.  And then you should be able to release it both in the destroy (to be safe) and after anywhere you call aws_s3_meta_request_finish_default when passing it in.

rccarper:
@TingDaoK Since I originally opened the PR, I can't approve it.  With the changes noted, I approve in spirit, and you can approve it/merge.

source/s3_auto_ranged_put.c
TingDaoK: yeah, I feel we do need a lock as well.😀

source/s3_request_messages.c
TingDaoK: yeah, that's fixed by the following pr https://github.com/awslabs/aws-c-s3/pull/53#discussion_r555345375

tests/s3_data_plane_tests.c
TingDaoK: I have removed this copy/paste.

TingDaoK (approved):

PR #54, opened 2020-12-31
Adding support for VIP Removal

rccarper:
*Description of changes:*
Adding support for VIPs being removed when their TTL expires.

### DIFF
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -394,11 +394,8 @@ static void s_s3_client_start_destroy(void *user_data) {
     /* Iterate through the local list, removing each VIP. */
     while (!aws_linked_list_empty(&local_vip_list)) {
         struct aws_linked_list_node *vip_node = aws_linked_list_pop_back(&local_vip_list);
-
         struct aws_s3_vip *vip = AWS_CONTAINER_OF(vip_node, struct aws_s3_vip, node);
-
         aws_s3_vip_start_destroy(vip);
-        vip = NULL;
     }
 
     aws_event_loop_group_release(client->body_streaming_elg);
@@ -1947,6 +1944,7 @@ static int s_s3_client_add_vips_default(struct aws_s3_client *client, const stru
     }
 
     struct aws_byte_cursor server_name = aws_byte_cursor_from_string(client->synced_data.endpoint);
+    bool vip_added = false;
 
     for (size_t address_index = 0; address_index < aws_array_list_length(host_addresses); ++address_index) {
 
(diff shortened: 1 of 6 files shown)


source/s3_client.c
```diff
@@ -110,20 +111,25 @@ static void s_s3_client_on_acquire_http_connection(
     void *user_data);
 
 /* Schedule task for processing work. */
-static void s_s3_client_schedule_process_work_task_synced(struct aws_s3_client *client);
+static void s_s3_client_schedule_process_work_synced(struct aws_s3_client *client);
+
```
justinboswell, line 115: What is the difference between these two functions? One is commented, one is not, and says default in it :)

justinboswell (commented):
If I'm reading this right, we no longer have a TTL override for these records. That will possibly limit the available VIPs to the point where it will affect throughput. Have you tested that?

source/s3_client.c
rccarper: Default is the actual function in the vtable, the other one calls the vtable.  I'll add a comment for that!

justinboswell (approved):

source/s3_client.c
```diff
@@ -363,33 +383,31 @@ static void s_s3_client_start_destroy(void *user_data) {
...
-            aws_s3_vip_start_destroy(vip);
-            vip = NULL;
-        }
+        struct aws_s3_vip *vip = AWS_CONTAINER_OF(vip_node, struct aws_s3_vip, node);
 
-        s_s3_client_lock_synced_data(client);
-        s_s3_client_schedule_process_work_task_synced(client);
-        s_s3_client_unlock_synced_data(client);
+        aws_s3_vip_start_destroy(vip);
+        vip = NULL;
```
bretambrose, line 401: Seems useless

source/s3_client.c
```diff
@@ -1926,13 +1905,173 @@ static void s_s3_client_on_host_resolver_address_resolved(
...
+    for (size_t address_index = 0; address_index < aws_array_list_length(host_addresses); ++address_index) {
+
+        if (client->synced_data.active_vip_count >= client->ideal_vip_count) {
+            break;
+        }
+
+        if (client->synced_data.allocated_vip_count >= aws_s3_client_get_max_allocated_vip_count(client)) {
+            AWS_LOGF_WARN(
+                AWS_LS_S3_CLIENT,
+                "id=%p Allocated VIP count (%d) is greater than or equal to twice the amount of allowed VIPs (%d). "
```
bretambrose, line 1960: text here (twice) is kind of couple to the underlying, current threshold, could be reworded to be less likely to be come inaccurate in the future.

source/s3_client.c
```diff
@@ -1926,13 +1905,173 @@ static void s_s3_client_on_host_resolver_address_resolved(
...
+        aws_linked_list_push_back(&client->synced_data.vips, &vip->node);
+
+        AWS_LOGF_INFO(
+            AWS_LS_S3_CLIENT,
+            "id=%p Initiating creation of VIP with address '%s', total active vip count %d",
+            (void *)client,
+            (const char *)host_address_byte_cursor.ptr,
+            client->synced_data.active_vip_count);
+
+        s_s3_client_schedule_process_work_synced(client);
```
bretambrose, line 2016: This being called each iteration is a little odd.

bretambrose (approved):

PR #57, opened 2021-01-04
Request jitter

rccarper:
*Description of changes:*
* Adding jitter to max request count.

### DIFF
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -771,6 +771,13 @@ static void s_s3_vip_sub_num_vip_connections_synced(struct aws_s3_vip *vip) {
     --vip->synced_data.num_vip_connections;
 }
 
+static void s_s3_client_vip_sub_num_vip_connections_synced(struct aws_s3_vip *vip) {
+    AWS_PRECONDITION(vip);
+    AWS_PRECONDITION(vip->owning_client);
+    ASSERT_SYNCED_DATA_LOCK_HELD(vip->owning_client);
+    --vip->synced_data.num_vip_connections;
+}
+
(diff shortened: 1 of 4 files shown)


source/s3_client.c
```diff
@@ -43,8 +44,9 @@ static const uint32_t s_s3_max_request_count_per_connection = 100;
 static const uint32_t s_connection_timeout_ms = 3000;
 
 /* TODO Provide analysis on origins of this value. */
-static const double s_throughput_per_vip_gbps = 6.25;
+static const double s_throughput_per_vip_gbps = 3.0;
 static const uint32_t s_num_connections_per_vip = 10;
+static const uint32_t s_max_request_jitter_range = 20;
```
justinboswell, line 49: Maybe just comment that this is 0.2 * 100, where 100 is the max per VIP, so the next poor bastard doesn't lose their mind in here

justinboswell (approved):
Fix-n-shipperooni

source/s3_client.c
```diff
@@ -1476,8 +1481,19 @@ static void s_s3_client_on_acquire_http_connection(
             *current_http_connection = NULL;
         }
 
+        AWS_ASSERT(s_s3_max_request_count_per_connection > s_max_request_jitter_range);
+
+        uint32_t jitter_value = 0;
```
JonathanHenson, line 1486: until we implement a faster random number generator that isn't hitting an entropy source directly everytime, can you make this as small as possible? In this case a uint8_t would be plenty.

JonathanHenson (approved):

justinboswell (approved):

PR #61, opened 2021-01-10
Parse XML response error

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- byte_buf to byte_cursor for the body of error
- If we have a XML type of error response, parse the message to have better error report

Debatable:
Parsing XML response should be here at native client? Or should it be some user level code?
Since I don't think any user level code will need the whole body, so put it here. But, maybe I am wrong.

TODO:
- Special error handling for getting an empty file?


TingDaoK:
Don't do that, but we do need to change the body of the result to byte_cursor instead

PR #69, opened 2021-01-13
Removing need for lock for stream read

rccarper:
*Description of changes:*
* Moving stream read to next_request for now where it doesn't require a lock.  (Need to find a way to communicate to derived types that next_request will never be called concurrently or out-of-order.)  We may push the stream read into a separate thread entirely (maybe the body streaming ELG), but this at least fixes the locking as a first step.
* Fixing up meta request cancel.
* Changing part size constants to be multiples of 8.


include/aws/s3/private/s3_auto_ranged_put.h
```diff
@@ -31,22 +31,27 @@ enum aws_s3_auto_ranged_put_request_tag {
 struct aws_s3_auto_ranged_put {
     struct aws_s3_meta_request base;
 
+    /* Useable after the Create Multipart Upload request succeeds. */
+    struct aws_string *upload_id;
+
+    struct {
+        /* Only meant for use in the next_request function, which is never called concurrently. */
+        uint32_t next_part_number;
+    } threaded_next_request_data;
```
TingDaoK, line 40: Trivial: Does it really worth to make another struct around it?

source/s3_auto_ranged_put.c
```diff
@@ -226,18 +243,34 @@ static int s_s3_auto_ranged_put_next_request(
...
                     meta_request,
                     AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_PART,
                     0,
                     AWS_S3_REQUEST_DESC_RECORD_RESPONSE_HEADERS);
 
+                aws_byte_buf_init(&request->request_body, meta_request->allocator, meta_request->part_size);
+                request->part_number = auto_ranged_put->threaded_next_request_data.next_part_number;
+                ++auto_ranged_put->threaded_next_request_data.next_part_number;
+
+                aws_s3_meta_request_read_body(meta_request, &request->request_body);
```
TingDaoK, line 271: Error check, it can fail right?

TingDaoK (commented):
Looks good

include/aws/s3/private/s3_auto_ranged_put.h
rccarper: I think so--I want to make sure that people are careful of where they change/read it.

source/s3_auto_ranged_put.c
```diff
@@ -575,49 +663,18 @@ static int s_s3_auto_ranged_put_stream_complete(
...
 
-            s_s3_auto_ranged_put_lock_synced_data(auto_ranged_put);
-
             /* Store the multipart upload id and set that we are ready for sending parts. */
-            auto_ranged_put->synced_data.upload_id = upload_id;
-            auto_ranged_put->synced_data.state = AWS_S3_AUTO_RANGED_PUT_STATE_SENDING_PARTS;
+            auto_ranged_put->upload_id = upload_id;
 
+            s_s3_auto_ranged_put_lock_synced_data(auto_ranged_put);
+            auto_ranged_put->synced_data.state = AWS_S3_AUTO_RANGED_PUT_STATE_SENDING_PARTS;
```
TingDaoK, line 670: What if the state is canceling now? So, if we were in `waiting_for_create` state and user requested cancel, this complete will happen from the other thread which will overwrite the cancel, right?

TingDaoK (approved):
More testing needed, I'll do that tomorrow!

PR #75, opened 2021-01-15
Add tests for cancel

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/s3_auto_ranged_put.c
```diff
@@ -218,7 +226,7 @@ static void s_s3_auto_ranged_put_finish(
         goto unlock;
     }
 
-    /* If the complete message has already been sent, then cancelling is not possible. */
+    /* If the complete message has already been sent, then canceling is not possible. */
     if (auto_ranged_put->synced_data.state == AWS_S3_AUTO_RANGED_PUT_STATE_WAITING_FOR_COMPLETE) {
```
TingDaoK, line 230: It will report the cancel succeed, but it's a lie. We didn't cancel the request, it actually finished successfully.

TingDaoK: We are hanging because of the error happens from callback will not send the request and the state has been changed. So, waiting for a response never comes back.😞
- We should not change the state if the complete is never really sent. It should fix the issue now.
- Maybe, if we are in waiting for complete state, we can probably do a delete for user? Since, user want to cancel the request, but abort is not realistic anymore, instead we can delete for them, if that's what they want. But it maybe too aggressive.

TingDaoK:
out of date

PR #76, opened 2021-01-15
Actually moving read outside of lock

rccarper:
* It looks like stream reads in my previous PR got moved back inside of a mutex while other bug fixes were being made in the same branch.  Re-moving it back out of the mutex.
* Removing unused state, cleaned up a little bit of logic, leaving TODOs in some places for refactoring.  (Trying not to move around too much code right now.)
* Fixing cancel bug if the create was attempted to be sent, but failed to ever make it to the service, causing the cancel.  (For now, tested this fix manually by running the test_s3_put_fail_object_inputstream_fail_reading test without credentials.)
 

source/s3_auto_ranged_put.c
```diff
@@ -290,40 +290,43 @@ static int s_s3_auto_ranged_put_next_request(
     struct aws_s3_request *request = NULL;
     struct aws_s3_auto_ranged_put *auto_ranged_put = meta_request->impl;
 
-    int result = AWS_OP_SUCCESS;
+    bool finish_canceling = false;
```
justinboswell, line 293: nitpick: spelling:`finish_cancelling`, or alternatively `cancel_complete` or something

justinboswell (approved):
Fix-n-ship

PR #79, opened 2021-01-22
Refactoring meta request types

rccarper:
*Description of changes:*
Some minor book keeping:
* aws_s3_request is now in its own header/source file.
* All meta request types now get their own private header file to make writing tests for those easier. 

Major changes:
* Meta request types no longer use state machines for tracking what request is to be sent next.  Instead, a series of conditions is evaluated each time next_request is called.
* The final meta request result is now set with a success or fail function.  Once all work that the meta request currently has in progress completes, then the meta request internally calls aws_s3_meta_request_finish, which invokes the finish callback.
* Performance Behavior Improvement: added a mitigation for global request limit being hit.  Client will first ask for new requests from meta requests with a "conservative" flag, which will cause auto-ranged-gets to limit the amount of requests returned, preventing parts from getting "backed up" waiting for earlier parts.
* Bug fix: no header/body callbacks should now ever be invoked after the finish callback.
* The client handles when a meta request is scheduled for work, instead of the meta request having to manage that at all.


include/aws/s3/private/s3_client_impl.h
```diff
@@ -100,10 +100,6 @@ struct aws_s3_client_vtable {
     struct aws_s3_meta_request *(
         *meta_request_factory)(struct aws_s3_client *client, const struct aws_s3_meta_request_options *options);
 
-    void (*push_meta_request)(struct aws_s3_client *client, struct aws_s3_meta_request *meta_request);
```
rccarper, line 103: Push and remove of meta requests for processing is now handled internally by the client.  (Before, the meta request would have to manage when to re-push itself.)

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -25,115 +26,30 @@ struct aws_http_make_request_options;
...
-typedef void(aws_s3_request_finished_callback_fn)(void *user_data);
-
 enum aws_s3_meta_request_state {
     AWS_S3_META_REQUEST_STATE_ACTIVE,
-    AWS_S3_META_REQUEST_STATE_CANCELING,
-    AWS_S3_META_REQUEST_STATE_CANCELED,
     AWS_S3_META_REQUEST_STATE_FINISHED,
 };
 
-enum aws_s3_request_desc_flags {
```
rccarper, line 43: This stuff was moved to s3_request.h

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -247,59 +175,8 @@ struct aws_s3_meta_request {
     } client_process_work_threaded_data;
 };
 
-/* Creates a new auto-ranged get meta request.  This will do multiple parallel ranged-gets when appropriate. */
-struct aws_s3_meta_request *aws_s3_meta_request_auto_ranged_get_new(
```
rccarper, line 251: derived type "new" functions were removed to their respective headers.

source/s3_auto_ranged_get.c
```diff
@@ -279,152 +280,129 @@ static int s_s3_auto_ranged_get_prepare_request(
     return AWS_OP_ERR;
 }
 
-static int s_s3_auto_ranged_get_header_block_done(
```
rccarper, line 282: Instead of using a mixture of the HTTP header callback and the notify_destroyed function later on, we instead use a new-and-improved request_finished function.  (The request will have stored the headers if we specified the appropriate flag when we created it during the update function.)

source/s3_auto_ranged_put.c
```diff
@@ -14,110 +19,37 @@ static const struct aws_byte_cursor s_create_multipart_upload_copy_headers[] = {
...
-    aws_mutex_lock(&auto_ranged_put->base.synced_data.lock);
-}
-
-static void s_s3_auto_ranged_put_unlock_synced_data(struct aws_s3_auto_ranged_put *auto_ranged_put) {
-    AWS_PRECONDITION(auto_ranged_put);
-
-    aws_mutex_unlock(&auto_ranged_put->base.synced_data.lock);
-}
-
-static struct aws_s3_meta_request_finish_options *s_copy_finish_options(
```
rccarper, line 78: We no longer cache finish options to be used later, so this is no longer needed.

source/s3_auto_ranged_put.c
```diff
@@ -557,176 +453,159 @@ static int s_s3_auto_ranged_put_prepare_request(
     return AWS_OP_ERR;
 }
 
-static int s_s3_auto_ranged_put_header_block_done(
```
rccarper, line 560: Similar to auto-ranged-get, we no longer spread response response handling over a series of callbacks.  Anything that was in those callbacks, is meant to be isolated in the request_finished function.

source/s3_auto_ranged_get.c
```diff
@@ -10,83 +16,39 @@
...
-    .incoming_body = NULL,
-    .stream_complete = NULL,
-    .notify_request_destroyed = s_s3_auto_ranged_get_notify_request_destroyed,
+    .finished_request = s_s3_auto_ranged_get_request_finished,
+    .delivered_requests = aws_s3_meta_request_delivered_requests_default,
     .destroy = s_s3_meta_request_auto_ranged_get_destroy,
     .finish = aws_s3_meta_request_finish_default,
 };
 
-static void s_s3_auto_ranged_get_lock_synced_data(struct aws_s3_auto_ranged_get *auto_ranged_get) {
```
rccarper, line 78: This was locking the base mutex anyway, so now just using the base type lock/unlock functions.

include/aws/s3/private/s3_auto_ranged_get.h
```diff
@@ -0,0 +1,38 @@
...
+#include "aws/s3/private/s3_meta_request_impl.h"
+
+enum aws_s3_auto_ranged_get_request_type {
+    AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_PART,
+};
+
+struct aws_s3_auto_ranged_get {
+    struct aws_s3_meta_request base;
+
+    struct {
```
justinboswell, line 18: Where da mutex at?

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -25,115 +26,35 @@ struct aws_http_make_request_options;
...
-        /* Recorded response body of the request. */
-        struct aws_byte_buf response_body;
-
-        /* Returned response status of this request. */
-        int response_status;
-
-    } send_data;
+    /* The client potentially has multiple meta requests that it can spread across connections, and the given meta
+       request can selectively not return a request if there is a performance reason to do so.*/
+    AWS_S3_META_REQUEST_UPDATE_FLAG_CONSERVATIVE = 0x00000002,
```
justinboswell, line 41: I feel like `_THROTTLED` might be more descriptive than `_CONSERVATIVE`, and then maybe have a version that is `_ALL` to differentiate?

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -25,115 +26,35 @@ struct aws_http_make_request_options;
...
-struct aws_s3_meta_request_finish_options {
-
-    struct aws_http_headers *error_response_headers;
-
-    struct aws_byte_buf *error_response_body;
-
-    int response_status;
-
-    int error_code;
+enum aws_s3_meta_request_update_status {
```
justinboswell, line 44: This looks startlingly like a bool? Is there any chance this grows?

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -25,115 +26,35 @@ struct aws_http_make_request_options;
...
-     * pointer. */
-    int (*next_request)(struct aws_s3_meta_request *meta_request, struct aws_s3_request **out_request);
-
-    /* Called when sending of the request has finished. */
-    void (*send_request_finish)(
-        struct aws_s3_vip_connection *vip_connection,
-        struct aws_http_stream *stream,
-        int error_code);
+    /* Update the meta request.  out_request is optional and can be NULL. If not null, and a request can be returned, it
+     * will be passed back into this variable. out_status is guaranteed to be non-null and indicates if there is
```
justinboswell, line 51: `out_status must be non-null`, it's not guaranteed since the user provides it.

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -226,6 +140,25 @@ struct aws_s3_meta_request {
...
+        uint32_t num_parts_delivery_succeeded;
+
+        /* Number of parts that have failed while trying to be delivered to the caller. */
+        uint32_t num_parts_delivery_failed;
+
+        /* The end finish result of the meta request. */
+        struct aws_s3_meta_request_result finish_result;
+
+        /* True if the finish result has been set. */
+        uint32_t finish_result_set : 1;
```
justinboswell, line 160: Is this necessary? Wouldn't the result be the way to detect this?

include/aws/s3/private/s3_request.h
```diff
@@ -0,0 +1,114 @@
...
+         * true or when this response indicates an error. */
+        struct aws_http_headers *response_headers;
+
+        /* Recorded response body of the request. */
+        struct aws_byte_buf response_body;
+
+        /* Returned response status of this request. */
+        int response_status;
+
+    } send_data;
```
justinboswell, line 84: Wouldn't this be more `attempt_data` or something?

include/aws/s3/private/s3_request.h
```diff
@@ -0,0 +1,114 @@
...
+     * be more contextual, like "first part" instead of just "part".) */
+    int request_tag;
+
+    /* When true, response headers from the request will be stored in the request's response_headers variable. */
+    uint32_t record_response_headers : 1;
+
+    /* When true, the response body buffer will be allocated in the size of a part. */
+    uint32_t part_size_response_body : 1;
+
+    struct {
```
justinboswell, line 59: 2 things:
1. this is a weird structure, why is it separated?
2. If you want to do this, put these anonymous structs above the bitfields for the containing struct, or else you'll break any size savings you were going to get due to struct packing.

source/s3_meta_request.c
```diff
@@ -208,9 +177,69 @@ int aws_s3_meta_request_init_base(
 }
 
 void aws_s3_meta_request_cancel(struct aws_s3_meta_request *meta_request) {
+    aws_s3_meta_request_lock_synced_data(meta_request);
+    aws_s3_meta_request_set_fail_synced(meta_request, NULL, AWS_ERROR_S3_CANCELED);
+    aws_s3_meta_request_unlock_synced_data(meta_request);
+}
+
+void aws_s3_meta_request_set_fail_synced(
```
justinboswell, line 185: confused why these are named `_synced` when there's no non-synced version and they don't hold any locks

justinboswell: oh, does `_synced` indicate it's only to be called inside a critical section?

include/aws/s3/private/s3_auto_ranged_get.h
justinboswell: ok after reading further these seem to be locked higher up. A comment next to each one explaining that would do wonders.

source/s3_meta_request.c
rccarper: Yeah, I was trying to indicate what functions assume a lock is held, and then they're meant to each have an ASSERT_SYNCED_DATA_LOCK_HELD that actually check if the mutex is locked.

include/aws/s3/private/s3_request.h
rccarper: I like "attempt_data" more.  I added a TODO for now.  (I'd rename it now, but it'll make the diff even bigger since this is relocated code.)

include/aws/s3/private/s3_meta_request_impl.h
rccarper: aws_s3_meta_request_result is visible to the user, so wanted to keep whether it was populated or not outside of the structure.

include/aws/s3/private/s3_meta_request_impl.h
rccarper: Yeah, I don't love using "conservative" either. I was originally thinking about something similar to "throttled" because "conservative" is really generic.  

But we don't necessarily have to do any type of request limiting outside of auto-ranged-gets, so it wasn't always applicable. :-/  I was looking for something that said, "Assume we have more meta requests that have work right now, so be careful when spinning up additional requests".

justinboswell (approved):

PR #80, opened 2021-01-27
Stream read bug fix

rccarper:
*Issue #, if available:*
Fixing the assumption that the specified content-length of the file being uploaded is the same size as the stream. The buffer we read into the from the specified input stream will now always be the exact required size for the last part, instead of always just the part size.


source/s3_auto_ranged_put.c
```diff
@@ -345,9 +347,20 @@ static int s_s3_auto_ranged_put_next_request(
                     0,
                     AWS_S3_REQUEST_DESC_RECORD_RESPONSE_HEADERS);
 
-                aws_byte_buf_init(&request->request_body, meta_request->allocator, meta_request->part_size);
                 request->part_number = auto_ranged_put->threaded_next_request_data.next_part_number;
 
+                size_t request_body_size = meta_request->part_size;
+
+                if (request->part_number == auto_ranged_put->synced_data.total_num_parts) {
```
justinboswell, line 354: `/* last part, adjust size to match remaining content length */`

justinboswell (approved):
Fix-n-shipperooni

include/aws/s3/private/s3_auto_ranged_put.h
```diff
@@ -33,6 +33,8 @@ struct aws_s3_auto_ranged_put {
     /* Useable after the Create Multipart Upload request succeeds. */
     struct aws_string *upload_id;
 
+    size_t content_length;
```
graebm, line 36: uint64_t?

source/s3_auto_ranged_put.c
graebm, line 354: just double checking: part_number starts at 1 and not 0, right?

rccarper: Correct!  S3 part numbers start at 1.

graebm (approved):

PR #83, opened 2021-02-05
Download Empty File

TingDaoK:
*Issue #, if available:*

*Description of changes:*


include/aws/s3/private/s3_request_messages.h
```diff
@@ -30,7 +30,8 @@ struct aws_http_message *aws_s3_get_object_message_new(
     struct aws_allocator *allocator,
     struct aws_http_message *base_message,
     uint32_t part_number,
-    size_t part_size);
+    size_t part_size,
+    uint8_t without_range);
```
bretambrose, line 34: bool?

source/s3_request_messages.c
```diff
@@ -33,7 +33,8 @@ struct aws_http_message *aws_s3_get_object_message_new(
     struct aws_allocator *allocator,
     struct aws_http_message *base_message,
     uint32_t part_number,
-    size_t part_size) {
+    size_t part_size,
+    uint8_t without_range) {
```
bretambrose, line 37: bool

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -264,6 +264,10 @@ void aws_s3_meta_request_set_fail_synced(
     struct aws_s3_request *failed_request,
     int error_code);
 
+/* Reset the request result. */
+AWS_S3_API
+void aws_s3_meta_request_reset_synced(struct aws_s3_meta_request *meta_request);
```
bretambrose, line 269: Feel like the name could be better.  The comment says it resets the result but that's not a part of the name.  Also, document caller assumptions here and maybe add invariants that the lock is held when called.

source/s3_auto_ranged_get.c
```diff
@@ -182,6 +182,39 @@ static void s_s3_auto_ranged_get_next_request(
         }
     }
 
+    /* One sepcial error case is an empty file download */
```
bretambrose, line 185: not directly related, but this function is turning into a nightmare and should probably be split/refactored if possible.

I feel like a lot of this could be moved into a helper function.

source/s3_auto_ranged_get.c
```diff
@@ -182,6 +182,39 @@ static void s_s3_auto_ranged_get_next_request(
...
+    struct aws_s3_meta_request_result finish_result = meta_request->synced_data.finish_result;
+    if (finish_result.error_response_body && finish_result.error_response_headers) {
+        struct aws_byte_cursor content_type;
+        AWS_ZERO_STRUCT(content_type);
+        if (!aws_http_headers_get(
+                finish_result.error_response_headers, aws_byte_cursor_from_c_str("content-type"), &content_type)) {
```
bretambrose, line 191: find existing or define new constants for all these strings

source/s3_auto_ranged_get.c
```diff
@@ -182,6 +182,39 @@ static void s_s3_auto_ranged_get_next_request(
...
+            /* Content type found */
+            if (aws_byte_cursor_eq_c_str_ignore_case(&content_type, "application/xml")) {
+                /* XML response */
+                struct aws_byte_cursor body_cursor = aws_byte_cursor_from_buf(finish_result.error_response_body);
+                struct aws_byte_cursor tag = aws_byte_cursor_from_c_str("ActualObjectSize");
+                struct aws_string *size = get_top_level_xml_tag_value(meta_request->allocator, &tag, &body_cursor);
+                if (aws_string_eq_c_str(size, "0")) {
```
bretambrose, line 198: you could store the comparison result and then destroy the string before checking it.  Then you wouldn't have to have multiple destroys in different branches.

include/aws/s3/private/s3_meta_request_impl.h
TingDaoK: `aws_s3_meta_request_reset_synced` has "reset" in it. The assumption of held a lock when called is implied by "_synced", but yeah, we should document it out to make it clear

include/aws/s3/private/s3_request_messages.h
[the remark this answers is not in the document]
TingDaoK: yeah, changed to boolean.

source/s3_auto_ranged_get.c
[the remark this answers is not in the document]
TingDaoK: removed the comment and followed Bret's suggestion to make it a helper function, which is more clear😀

TingDaoK:
moved to https://github.com/awslabs/aws-c-s3/pull/86

PR #86, opened 2021-02-12
New empty download

TingDaoK:
Because Git is messed up, make a new PR instead. 
previous review is https://github.com/awslabs/aws-c-s3/pull/83


source/s3_auto_ranged_get.c
```diff
@@ -100,6 +102,30 @@ static void s_s3_meta_request_auto_ranged_get_destroy(struct aws_s3_meta_request
     aws_mem_release(meta_request->allocator, auto_ranged_get);
 }
 
+/* Check the finish result of meta request, in case of the request failed because of downloading an empty file */
+static bool s_check_empty_file_download_error_synced(struct aws_s3_meta_request *meta_request) {
```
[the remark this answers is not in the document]
rccarper: The request_finished could set a flag in that case, and then you can branch on it in the update to spin up the request with your new request tag. :)

source/s3_auto_ranged_get.c
```diff
@@ -400,9 +448,20 @@ static void s_s3_auto_ranged_get_request_finished(
...
-
-        aws_s3_meta_request_set_fail_synced(meta_request, request, error_code);
+        if (s_check_empty_file_download_error_synced(request)) {
+            AWS_LOGF_DEBUG(
+                AWS_LS_S3_META_REQUEST,
+                "id=%p Getting an empty file, create a new request without range header to fetch the empty "
+                "file",
+                (void *)meta_request);
+            auto_ranged_get->synced_data.get_without_range = true;
+            ++meta_request->synced_data.num_parts_delivery_completed;
```
TingDaoK, line 458: maybe I should update the num parts failed and invoke the `s_s3_meta_request_delivered_requests` instead?

rccarper: I think we'll need to put the if/else that starts on line 434 inside of an if(request_tag == AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_PART).  And then we can have a separate block for if(request_tag == AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_PART_WITHOUT_RANGE) which sets request_without_range_completed to true, and then calls aws_s3_meta_request_set_fail_synced if the request completed with an error.

source/s3_auto_ranged_get.c
```diff
@@ -139,6 +166,22 @@ static bool s_s3_auto_ranged_get_update(
             }
         }
 
+        if (auto_ranged_get->synced_data.get_without_range) {
```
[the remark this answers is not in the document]
TingDaoK: Previously, I thought it can help if we want to have a get request without range from the beginning of the process. But, yeah, seems like it should just for the special error handling.

source/s3_auto_ranged_get.c
```diff
@@ -161,6 +188,28 @@ static bool s_s3_auto_ranged_get_update(
             goto has_work_remaining;
         }
 
+        if (auto_ranged_get->synced_data.get_without_range) {
+            if (out_request == NULL) {
```
rccarper, line 192: Move this conditional right after the next one, ie, if(auto_ranged_get->synced_data.get_without_range_sent).  At this point we don't know for certain there is actually work remaining.

source/s3_auto_ranged_get.c
```diff
@@ -381,28 +442,44 @@ static void s_s3_auto_ranged_get_request_finished(
...
+            } else {
+                ++auto_ranged_get->synced_data.num_parts_failed;
+                aws_s3_meta_request_set_fail_synced(meta_request, request, error_code);
+            }
+        }
     } else {
-        ++auto_ranged_get->synced_data.num_parts_failed;
-
-        aws_s3_meta_request_set_fail_synced(meta_request, request, error_code);
+        AWS_LOGF_DEBUG(AWS_LS_S3_META_REQUEST, "id=%p Get empty file completed", (void *)meta_request);
```
rccarper, line 478: Could we do an else if(request->request_tag == AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_PART) here?  (Just to make it more explicit.)

tests/s3_data_plane_tests.c
```diff
@@ -601,6 +706,35 @@ static int s_test_s3_put_object_less_than_part_size(struct aws_allocator *alloca
...
 
+AWS_TEST_CASE(test_s3_put_object_empty_object, s_test_s3_put_object_empty_object)
+static int s_test_s3_put_object_empty_object(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+
+    struct aws_s3_tester tester;
+    ASSERT_SUCCESS(aws_s3_tester_init(allocator, &tester));
+
+    struct aws_s3_client_config client_config = {
+        .part_size = 20 * 1024 * 1024,
```
rccarper, line 717: Trivial: not critical, could just zero out the config here.

source/s3_auto_ranged_get.c
```diff
@@ -161,6 +188,28 @@ static bool s_s3_auto_ranged_get_update(
...
+                goto has_work_remaining;
+            }
+            if (auto_ranged_get->synced_data.get_without_range_sent) {
+                if (auto_ranged_get->synced_data.get_without_range_completed) {
+                    goto no_work_remaining;
+                } else {
+                    goto has_work_remaining;
+                }
+            }
+            /* Set to send a request without range to meet some special case, eg: empty file download */
```
rccarper, line 202: Trivial: comment looks out of date now?

tests/s3_data_plane_tests.c
```diff
@@ -572,6 +580,103 @@ static int s_test_s3_multipart_put_object_with_acl(struct aws_allocator *allocat
...
+
+    ASSERT_SUCCESS(aws_s3_tester_bind_client(
+        &tester, &client_config, AWS_S3_TESTER_BIND_CLIENT_REGION | AWS_S3_TESTER_BIND_CLIENT_SIGNING));
+
+    struct aws_s3_client *client = aws_s3_client_new(allocator, &client_config);
+
+    struct aws_string *host_name =
+        aws_s3_tester_build_endpoint_string(allocator, &g_test_bucket_name, &g_test_s3_region);
+
+    for (size_t i = 0; i < num_meta_requests; ++i) {
```
rccarper, line 613: Trivial: not critical, but looks like you could combine these two for loops.

tests/s3_data_plane_tests.c
```diff
@@ -572,6 +580,103 @@ static int s_test_s3_multipart_put_object_with_acl(struct aws_allocator *allocat
...
+
+    ASSERT_TRUE(
+        num_meta_requests == (sizeof(meta_request_test_results) / sizeof(struct aws_s3_meta_request_test_results)));
+
+    struct aws_s3_tester tester;
+    AWS_ZERO_STRUCT(tester);
+    ASSERT_SUCCESS(aws_s3_tester_init(allocator, &tester));
+
+    struct aws_s3_client_config client_config = {
+        .part_size = 64 * 1024,
```
rccarper, line 602: Trivial: not critical, but could just zero the config here.

source/s3_auto_ranged_get.c
```diff
@@ -381,28 +442,44 @@ static void s_s3_auto_ranged_get_request_finished(
...
+        } else {
+            if (s_check_empty_file_download_error(request)) {
+                AWS_LOGF_DEBUG(
+                    AWS_LS_S3_META_REQUEST,
+                    "id=%p Getting an empty file, create a new request without range header to fetch the empty "
+                    "file",
+                    (void *)meta_request);
+                auto_ranged_get->synced_data.get_without_range = true;
+            } else {
+                ++auto_ranged_get->synced_data.num_parts_failed;
```
rccarper, line 473: I missed this--I think we should move this right after line 475 so that it increments for empty file download too (since the part did fail in this case).

rccarper (approved):
Fix and ship!

PR #94, opened 2021-02-22
Fixing clean up bugs

rccarper:
Fixing race condition clean up bugs:
* If client exits before the first host resolver callback, then it would crash.
* If client were to clean up all resources while and call the finish_destroy function while start_destroy is still executing, it could try to schedule the work task at the end of the function with a destroyed client.

### DIFF
--- a/include/aws/s3/private/s3_client_impl.h
+++ b/include/aws/s3/private/s3_client_impl.h
@@ -213,6 +213,8 @@ struct aws_s3_client {
         /* Whether or not the client has started cleaning up all of its resources */
         uint32_t active : 1;
 
+        uint32_t waiting_for_first_callback : 1;
+
         /* Whether or not work processing is currently scheduled. */
         uint32_t process_work_task_scheduled : 1;
 
--- a/include/aws/s3/private/s3_client_impl.h
+++ b/include/aws/s3/private/s3_client_impl.h
@@ -213,7 +213,12 @@ struct aws_s3_client {
         /* Whether or not the client has started cleaning up all of its resources */
         uint32_t active : 1;
 
-        uint32_t waiting_for_first_callback : 1;
+        /* True if the start_destroy function is still executing, which blocks shutdown from completing. */
+        uint32_t start_destroy : 1;
+
+        /* True if the client has called aws_host_resolver_resolve_host but hasn't received a callback yet. There isn't
+         * a way to cancel this first callback, so this will block shutdown from completing. */
+        uint32_t waiting_for_first_host_resolve_callback : 1;
 
         /* Whether or not work processing is currently scheduled. */
         uint32_t process_work_task_scheduled : 1;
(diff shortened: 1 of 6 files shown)


include/aws/s3/private/s3_client_impl.h
```diff
@@ -213,6 +213,13 @@ struct aws_s3_client {
         /* Whether or not the client has started cleaning up all of its resources */
         uint32_t active : 1;
 
+        /* True if the start_destroy function is still executing, which blocks shutdown from completing. */
+        uint32_t start_destroy : 1;
```
TingDaoK, line 217: I feel this name is a little confusing, if I don't know that it's actually the name of the function. So, maybe something like `start_destroy_executing` would be more clear?

source/s3_client.c
```diff
@@ -1967,6 +1987,15 @@ static void s_s3_client_body_streaming_task(struct aws_task *task, void *arg, en
     aws_s3_client_release(client);
 }
 
+/* Called by aws_s3_request when it has finished being destroyed */
+static void s_s3_client_clear_waiting_for_first_host_resolve_callback(struct aws_s3_client *client) {
```
TingDaoK, line 1991: ***_synced? Keep the naming consistent with other function that touches the synced data.

TingDaoK (approved):
Overall, the change looks good, but the shutdown process of client is just too complicated... But, it seems like inevitable, as we have all those asynced jobs😞

rccarper:
> Overall, the change looks good, but the shutdown process of client is just too complicated... But, it seems like inevitable, as we have all those asynced jobs😞

I agree.  We should be able to simplify it by centering it more around the work_task instead of having everything async shutting down to shut down in-place.  (But definitely a more involved fix and not for this PR.)

rccarper:
> > Overall, the change looks good, but the shutdown process of client is just too complicated... But, it seems like inevitable, as we have all those asynced jobs😞
> 
> I agree. We should be able to simplify it by centering it more around the work_task instead of having everything async shutting down to shut down in-place. (But definitely a more involved fix and not for this PR.)

I take that back, that change wasn't actually that bad. :)  It should get better too with some of the client simplifications we're looking at.

TingDaoK (commented):
I don't like the idea of scheduling task during shutdown process. 
- The task may run synced if the eventloop is gone during shutdown, [doc](https://github.com/awslabs/aws-c-io/blob/main/include/aws/io/event_loop.h#L265), which will lead to a possible recursive lock. Unless we can make sure the eventloop is the last thing get cleaned up.
- It does not really simplify the shutdown process. To clarify a little about I think it's complicated since we are blocked by different asynced shutdown using lots of booleans. Maybe, to improve from my perspective, we could use refcount on the client for shutdown?

TingDaoK (approved):
Convinced. Ship it.🛳 Summarize a bit about our discussion. 
- We do hold the reference to eventloop, which does make sure the task will not run synced.
- The task helps the shutdown process happens from one single place. It's clear from the state for debugging.
- The refcounting will lead to issues like cannot phased shutdown.

PR #95, opened 2021-02-22
Moving response body streaming

rccarper:
*Description of Changes*
Moving response body streaming from the client into the meta request so that the streaming logic is in one place.  In an upcoming change, the meta request will be scheduling more on its io_event_loop, and this will also help keep all tasks for that event loop under the same type.

Important call out: I would have liked to put the small block allocator in front of the client allocator and just have meta request use it for everything.  Unfortunately, a meta request and a client can have different lifetimes.


source/s3_meta_request.c
```diff
@@ -825,62 +829,105 @@ void aws_s3_meta_request_stream_response_body_synced(
...
-}
+    struct aws_s3_client *client = meta_request->client;
+    AWS_PRECONDITION(client);
 
-void aws_s3_meta_request_delivered_requests_default(
-    struct aws_s3_meta_request *meta_request,
-    int error_code,
-    uint32_t num_failed,
-    uint32_t num_successful) {
+    /* Client owns this event loop group. A cancel should not be possible. */
```
justinboswell, line 888: Are we sure about this? Does that imply that the client would wait for all tasks to complete somewhere before shutting down?

rccarper: Yes--the meta request holds a reference to the client while the meta request is in progress, so the related ELG will never get shutdown until the meta request is finished, which indirectly causes the "wait".

If the meta request gets canceled and a task has been scheduled, the task still needs to run before the meta request can finish (and the client to clean up).  But, it'll just throw away the request in that case instead of trying to stream it to the user.

justinboswell (approved):

PR #112, opened 2021-04-13
Fixing signing bug

rccarper:
*Description of changes:*
Fixing signing bug where additional slash could cause a signing mismatch.


justinboswell (approved):

TingDaoK (approved):
Interesting that we normalized uri path for signing but s3 did not. 
Trivial: `aws_s3_tester_send_meta_request_with_options` is a real monster now.

rccarper:
> Interesting that we normalized uri path for signing but s3 did not.
> Trivial: `aws_s3_tester_send_meta_request_with_options` is a real monster now.

Yeah, it's friendly to use, but could be broken up a bit.

PR #114, opened 2021-04-15
Fix for abort not being sent

rccarper:
*Description of changes:*
This fixes an issue where the abort message for multipart uploads is not being sent.

This is because we send the abort when the meta request fails or is canceled, which sets it as having a "finish result".  This currently prevents additional requests from being sent, which we ordinarily want for messages sending additional parts, but not for the abort message.

### DIFF
--- a/include/aws/s3/private/s3_request.h
+++ b/include/aws/s3/private/s3_request.h
@@ -15,10 +15,10 @@ struct aws_http_message;
 struct aws_signable;
 struct aws_s3_meta_request;
 
-enum aws_s3_request_desc_flags {
-    AWS_S3_REQUEST_DESC_RECORD_RESPONSE_HEADERS = 0x00000001,
-    AWS_S3_REQUEST_DESC_PART_SIZE_RESPONSE_BODY = 0x00000002,
-    AWS_S3_REQUEST_DESC_ALWAYS_SEND = 0x00000004,
+enum aws_s3_request_flags {
+    AWS_S3_REQUEST_FLAG_RECORD_RESPONSE_HEADERS = 0x00000001,
+    AWS_S3_REQUEST_FLAG_PART_SIZE_RESPONSE_BODY = 0x00000002,
+    AWS_S3_REQUEST_FLAG_ALWAYS_SEND = 0x00000004,
 };
 
 /* Represents a single request made to S3. */
@@ -88,7 +88,7 @@ struct aws_s3_request {
     /* When true, this request is being tracked by the client for limiting the amount of in-flight-requests/stats. */
     uint32_t tracked_by_client : 1;
 
-    /* When true, even when the meta request is finished, this request will be sent. */
+    /* When true, even when the meta request has a finished result, this request will be sent. */
     uint32_t always_send : 1;
 };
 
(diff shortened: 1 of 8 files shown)


include/aws/s3/private/s3_request.h
```diff
@@ -17,7 +17,8 @@ struct aws_s3_meta_request;
 
 enum aws_s3_request_desc_flags {
     AWS_S3_REQUEST_DESC_RECORD_RESPONSE_HEADERS = 0x00000001,
-    AWS_S3_REQUEST_DESC_PART_SIZE_RESPONSE_BODY = 0x0000002,
+    AWS_S3_REQUEST_DESC_PART_SIZE_RESPONSE_BODY = 0x00000002,
+    AWS_S3_REQUEST_DESC_ALWAYS_SEND = 0x00000004,
```
TingDaoK, line 21: Irrelevant: What does desc stand for here?

include/aws/s3/private/s3_request.h
```diff
@@ -86,6 +87,9 @@ struct aws_s3_request {
 
     /* When true, this request is being tracked by the client for limiting the amount of in-flight-requests/stats. */
     uint32_t tracked_by_client : 1;
+
+    /* When true, even when the meta request is finished, this request will be sent. */
```
TingDaoK, line 91: `is finished` is a little confusing. Maybe `has finish result`?
Since when the abort is in process, we are not at the finished state, right?

include/aws/s3/private/s3_request.h
rccarper: It's left over from when there used to a be an aws_s3_request_desc structure, ie, "AWS S3 Request Description".  I have removed it!

source/s3_client.c
```diff
@@ -1676,8 +1676,9 @@ void aws_s3_client_update_connections_threaded(struct aws_s3_client *client, boo
         uint32_t num_conns_per_vip = s_num_conns_per_vip_meta_request_look_up[request->meta_request->type];
         const uint32_t max_active_connections = aws_s3_client_get_max_active_connections(client, num_conns_per_vip);
 
-        /* If this meta request was finished, then let the meta request know and release the request. */
-        if (aws_s3_meta_request_has_finish_result(request->meta_request)) {
+        /* Unless the request is marked "always send", if this meta request is finished, then finish the request now and
+         * release it. */
+        if (!request->always_send && aws_s3_meta_request_has_finish_result(request->meta_request)) {
```
TingDaoK, line 1681: Haven't dived deep into it, can you birefly explain about why it's checking the finish_result instead of `aws_s3_meta_request_is_finished`?

rccarper: has_finish_result = We know that this thing failed or succeeded, but it could still be winding down and have in flight requests.
is_finished = Nothing left for this thing to do at all.

TingDaoK: Yeah, maybe add it to the commends? Since, the comments saying `if this meta request is finished` is confusing as well.

TingDaoK (approved):
Fix & ship

source/s3_client.c
rccarper: Added it!

PR #125, opened 2021-05-18
Exposing test globals as environment variables

rccarper:
*Description of changes:*
Exposing constants for buckets/testing-region as environment variables.  This allows users without access to those resources to specify their own.

### DIFF
--- a/tests/s3_tester.c
+++ b/tests/s3_tester.c
@@ -8,6 +8,7 @@
 #include "aws/s3/private/s3_meta_request_impl.h"
 #include "aws/s3/private/s3_util.h"
 #include <aws/auth/credentials.h>
+#include <aws/common/environment.h>
 #include <aws/common/system_info.h>
 #include <aws/http/request_response.h>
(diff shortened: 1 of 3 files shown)


tests/s3_tester.c
```diff
@@ -474,6 +498,21 @@ void aws_s3_tester_clean_up(struct aws_s3_tester *tester) {
     aws_event_loop_group_release(tester->el_group);
     tester->el_group = NULL;
 
+    if (tester->test_region_string != NULL) {
+        aws_string_destroy(tester->test_region_string);
```
bretambrose, line 502: string destroy is null-safe

bretambrose (approved):
It's probably fine that the tests here are unconditional, but eventually I think the s3 tests in the CRTs should be skipped if the environment isn't set up for them.

TingDaoK:
We already did it [here](https://github.com/awslabs/aws-c-s3/pull/258)

PR #128, opened 2021-05-27
Adding utility functions

rccarper:
*Description of changes:*
Adding several utility functions in preparation for supporting requests that contain range headers.


source/s3_util.c
```diff
@@ -307,3 +314,161 @@ void aws_s3_add_user_agent_header(struct aws_allocator *allocator, struct aws_ht
...
+    if (aws_http_headers_get(response_headers, g_content_length_header_name, &content_length_header_value)) {
+        aws_raise_error(AWS_ERROR_S3_MISSING_CONTENT_LENGTH_HEADER);
+        return AWS_OP_ERR;
+    }
+
+    struct aws_string *content_length_header_value_str =
+        aws_string_new_from_cursor(allocator, &content_length_header_value);
+    char *content_length_str_end = NULL;
+
+    *out_content_length = strtoull((const char *)content_length_header_value_str->bytes, &content_length_str_end, 10);
```
TingDaoK, line 395: Error check here? Check the `content_length_str_end == '\0'`? And referring to a stack overflow [answer](https://stackoverflow.com/a/11279793), why not use `sscanf` instead like the previous function?

include/aws/s3/private/s3_util.h
```diff
@@ -108,11 +111,38 @@ AWS_S3_API
...
+    uint64_t *out_object_size);
+
+AWS_S3_API
+int aws_s3_parse_content_length_response_header(
+    struct aws_allocator *allocator,
+    struct aws_http_headers *response_headers,
+    uint64_t *out_content_length);
+
+AWS_S3_API
+uint32_t aws_s3_get_num_parts(size_t part_size, uint64_t object_range_start, uint64_t object_range_end);
```
TingDaoK, line 135: Please add some documentations here, it will help to understand what's this function doing. I feel this function is easy to misunderstand as it calculate the number of parts in a different way than I expected (aligned to the start of a part and includes both the first part and last part even if they are partial of one part)

source/s3_util.c
```diff
@@ -307,3 +314,161 @@ void aws_s3_add_user_agent_header(struct aws_allocator *allocator, struct aws_ht
...
+
+        if ((aligned_range_remainder % part_size) > 0) {
+            ++num_parts;
+        }
+    }
+
+    return num_parts;
+}
+
+/* Calculate the part range for a given part. Intended to be used in conjunction with aws_s3_get_num_parts. */
```
TingDaoK, line 432: Maybe move this comments to header file?

include/aws/s3/private/s3_util.h
```diff
@@ -108,11 +111,38 @@ AWS_S3_API
...
+
+AWS_S3_API
+void aws_s3_get_part_range(
```
TingDaoK, line 138: Same here, what are those params? It'll help if we can clarify it a little more than just naming.

include/aws/s3/private/s3_util.h
```diff
@@ -108,11 +111,38 @@ AWS_S3_API
...
 
 /* TODO could be moved to aws-c-common. */
+AWS_S3_API
 int aws_last_error_or_unknown(void);
 
 AWS_S3_API
 void aws_s3_add_user_agent_header(struct aws_allocator *allocator, struct aws_http_message *message);
 
+AWS_S3_API
+int aws_s3_parse_content_range_response_header(
```
TingDaoK, line 121: I'll comment it out that `out_object_size` is required and `out_range_start/out_range_end` are optional

source/s3_util.c
```diff
@@ -307,3 +314,161 @@ void aws_s3_add_user_agent_header(struct aws_allocator *allocator, struct aws_ht
...
+    uint64_t part_size_uint64 = (uint64_t)part_size;
+    uint64_t first_part_size = part_size_uint64;
+    uint64_t first_part_alignment_offset = object_range_start % part_size_uint64;
+
+    AWS_ASSERT(part_number > 0);
+
+    const uint32_t part_index = part_number - 1;
+
+    /* Part index is assumed to be in a valid range. */
+    AWS_ASSERT(part_index < aws_s3_get_num_parts(part_size, object_range_start, object_range_end));
```
TingDaoK, line 452: Trivial: can you move those checks to the beginning of the function? Right after the `AWS_PRECONDITION(out_part_range_end);`

source/s3_util.c
```diff
@@ -307,3 +314,161 @@ void aws_s3_add_user_agent_header(struct aws_allocator *allocator, struct aws_ht
...
+
+    return num_parts;
+}
+
+/* Calculate the part range for a given part. Intended to be used in conjunction with aws_s3_get_num_parts. */
+void aws_s3_get_part_range(
+    uint64_t object_range_start,
+    uint64_t object_range_end,
+    size_t part_size,
+    uint32_t part_number,
```
TingDaoK, line 437: Maybe comment it out: This part number is # of part inside the range should be smaller or equal to the result from `aws_s3_get_num_parts`, I was confused until read through the implementation.

TingDaoK (commented):
Looks good, most of the comments are about adding documentation to clarify what the function is doing.

source/s3_util.c
rccarper: Using sscanf, added error check.

include/aws/s3/private/s3_util.h
rccarper: Added a more detailed comment.  Let me know if those params are still confusing!

TingDaoK (approved):
🛳

PR #132, opened 2021-06-01
Independent Dashboard

TingDaoK:
*Issue #, if available:*

*Description of changes:*
- Create IAM roles instead of use the existing role to make the dashboard independent from pre-configured AWS account
- Used https://www.npmjs.com/package/cdk-ec2-key-pair to generate the key pair and stored it in Secret Manager.
- Note: You may need to run `cdk bootstrap aws://<ACCOUNT-ID>/<REGION>`

### TODO
- I don't have a good idea about how to test it. But, as I tested it manually in my AWS account it works well
- The performance test is still depend on a pre-configured bucket and object. Will create one in the following PR.


benchmarks/dashboard-stack/lib/dashboard-stack.ts
```diff
@@ -201,5 +229,12 @@ export class DashboardStack extends cdk.Stack {
             projectName: "S3BenchmarksDeploy"
         });
 
+        // Create the Key Pair
+        const key = new KeyPair(this, 'EC2CanaryKeyPair', {
```
rccarper, line 233: Not necessary to change now (or maybe ever), but I think it would be okay to have this passed in as an argument to the stack as a pre-existing resource.

benchmarks/benchmarks-stack/benchmarks-stack/lib/canary-policy-doc.json
```diff
@@ -0,0 +1,118 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "VisualEditor0",
```
rccarper, line 5: Extra credit: rename this Sid to something more useful and sort the actions so that we can more easily see what is allowed for each service.

rccarper (approved):

benchmarks/dashboard-stack/lib/dashboard-stack.ts
TingDaoK: hmmm, I think It's not bad to have it. Any reason we want to remove this?

benchmarks/benchmarks-stack/benchmarks-stack/lib/canary-policy-doc.json
TingDaoK: yeah, changed, also removed the actions related to CFN, I think I added them when we want EC2 to delete the CFN stack. Now, we use Lambda instead

PR #135, opened 2021-06-11
Creation of elg can fail. Handle the error

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -239,6 +239,9 @@ struct aws_s3_client *aws_s3_client_new(
 #endif
 
     struct aws_s3_client *client = aws_mem_calloc(allocator, 1, sizeof(struct aws_s3_client));
+    if (!client) {
+        return NULL;
+    }
 
     client->allocator = allocator;
     client->sba_allocator = aws_small_block_allocator_new(allocator, true);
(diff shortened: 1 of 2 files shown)


rccarper:
I think we have an existing bug with the on_error label here.  I don't expect that we can do reliably do a s3_client_release in this case.

TingDaoK:
> I think we have an existing bug with the on_error label here. I don't expect that we can do reliably do a s3_client_release in this case.

What existing bug? We should probably fix it as well

source/s3_client.c
```diff
@@ -393,9 +398,22 @@ struct aws_s3_client *aws_s3_client_new(
...
-    aws_s3_client_release(client);
-
+    aws_event_loop_group_release(client->body_streaming_elg);
+    client->body_streaming_elg = NULL;
+    if (client->tls_connection_options) {
+        aws_tls_connection_options_clean_up(client->tls_connection_options);
+        aws_mem_release(client->allocator, client->tls_connection_options);
+        client->tls_connection_options = NULL;
+    }
+elg_create_fail:
```
rccarper, line 408: I think this should be moved after the event_loop_group_release?

TingDaoK: Do, you mean after the `aws_event_loop_group_release(client->client_bootstrap->event_loop_group);`? But, this `elg_create_fail` tag will handle the error when body_streaming_elg failed to create, and the `client_config->client_bootstrap->event_loop_group` is acquired before the creation of body_streaming_elg.

rccarper (approved):

PR #136, opened 2021-06-23
Multiple Bucket Support

rccarper:
*Description of changes:*
* Adding support for multiple buckets/endpoints per client.  I've removed the manual connection management and tracking of individual IP addresses.  Instead, we track endpoint names that each have a connection manager via aws-c-http.  Some logic is still done around the number of IP's for a particular endpoint, but we know longer track each actual IP.
* The name "aws_s3_vip_connection" no longer makes sense because we no longer track connections in relation to VIPs.  That will be done here: https://github.com/awslabs/aws-c-s3/pull/137


include/aws/s3/private/s3_client_impl.h
```diff
@@ -21,113 +22,104 @@
...
-    /* S3 Client that owns this vip. */
-    struct aws_s3_client *owning_client;
+struct aws_s3_endpoint_options {
+    /* URL of the host that this endpoint refers to. */
+    struct aws_string *host_name;
 
-    /* Connection manager shared by all VIP connections. */
-    struct aws_http_connection_manager *http_connection_manager;
+    /* Callback for when the reference count of this endpoint hits zero. */
+    aws_s3_endpoint_ref_zero_fn *ref_count_zero_callback;
```
justinboswell, line 47: just call this a destructor, I know it's nomenclature that aws_refcount uses, but this is weird to read

include/aws/s3/private/s3_client_impl.h
```diff
@@ -21,113 +22,104 @@
...
+        const struct aws_string *host_name,
+        uint32_t flags);
 
     void (*schedule_process_work_synced)(struct aws_s3_client *client);
 
     void (*process_work)(struct aws_s3_client *client);
 
-    void (
-        *setup_vip_connection_retry_token)(struct aws_s3_client *client, struct aws_s3_vip_connection *vip_connection);
+    bool (*endpoint_ref_count_zero)(struct aws_s3_endpoint *endpoint);
```
justinboswell, line 120: Destructor!

source/s3_endpoint.c
```diff
@@ -0,0 +1,339 @@
...
+#include <aws/io/uri.h>
+
+#include <inttypes.h>
+#include <math.h>
+
+static const uint32_t s_connection_timeout_ms = 3000;
+static const uint16_t s_http_port = 80;
+static const uint16_t s_https_port = 443;
+
+static size_t s_dns_host_address_ttl_seconds = 5 * 60;
```
justinboswell, line 41: This should be on the options struct, and default values like this should be centralized (so that when we hook all of this up to ENV vars and config APIs it's easy to find)

justinboswell (approved):
Fix-n-ship!

PR #140, opened 2021-06-25
Fixing cancelling during request preparation

rccarper:
*Description of changes:*
* Fixing case where canceling during request preparation (reading from a stream, for example), could still allow already queued requests to also be prepared (causing additional reads from the stream).
* Adding endpoint reference to connection structure so that it can never go away out from under the connection.


TingDaoK (approved):
LGTM

PR #141, opened 2021-07-01
Small bug fixes

rccarper:
*Description of changes:*
* Improving ASSERT_SYNCED_DATA_LOCK_HELD
* Taking out null checks for out_request, which should no longer be necessary.
* Falling back to default meta request for a request of a part, as I don't believe this is splittable. Still needs a corresponding test.
* Adding user agent header in the prepare instead of modifying the original request. Tests may be failing because of the way they are overriding the prepare_request virtual function, which only overrides some of the virtual function behavior.

### DIFF
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -752,6 +752,11 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
 
     /* Call the appropriate meta-request new function. */
     if (options->type == AWS_S3_META_REQUEST_TYPE_GET_OBJECT) {
+
+        if (aws_http_headers_has(initial_message_headers, aws_byte_cursor_from_c_str("partNumber"))) {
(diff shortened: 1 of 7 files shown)


source/s3_client.c
```diff
@@ -752,6 +752,14 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
 
     /* Call the appropriate meta-request new function. */
     if (options->type == AWS_S3_META_REQUEST_TYPE_GET_OBJECT) {
+
+        /* If the initial request already has partNumber, the request is not splittable(?). Treat it as a Default
```
justinboswell, line 756: This should be a backlog task somewhere.

justinboswell (approved):

source/s3_client.c
TingDaoK: updated the Asana task, included the test is still needed in it.

TingDaoK (approved):

PR #144, opened 2021-07-19
ifndef BYO_CRYPTO

ilevyor:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -336,7 +336,7 @@ struct aws_s3_client *aws_s3_client_new(
             aws_tls_connection_options_copy(client->tls_connection_options, client_config->tls_connection_options);
         } else {
 #ifdef BYO_CRYPTO
-            AWS_ASSERT(false);
+            AWS_FATAL_ASSERT(false);
             goto on_error;
 #else
             struct aws_tls_ctx_options default_tls_ctx_options;
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -753,6 +753,14 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
 
     /* Call the appropriate meta-request new function. */
     if (options->type == AWS_S3_META_REQUEST_TYPE_GET_OBJECT) {
+
+        /* If the initial request already has partNumber, the request is not splittable(?). Treat it as a Default
+         * request.
+         * TODO: Still need tests to verify that the request of a part is splittable or not */
+        if (aws_http_headers_has(initial_message_headers, aws_byte_cursor_from_c_str("partNumber"))) {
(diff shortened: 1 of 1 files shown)


justinboswell (changes requested):
I don't understand this change, the body_streaming_elg exists whether or not BYO_CRYPTO is enabled, how does this relate?

TingDaoK (changes requested):
it's complaining about on_error not used for `BYO_CRYPTO`, I guess you need to add something like `goto on_error` on line 339. And raise some error

ilevyor:
commit is necessary due to BYO_CRYPTO builds failing in aws-crt-cpp 
https://github.com/awslabs/aws-crt-cpp/runs/3106110215?check_suite_focus=true

> `/root/aws-crt-cpp/crt/aws-c-s3/source/s3_client.c:411:1: error: label 'on_error' defined but not used [-Werror=unused-label]
 on_error:`

source/s3_client.c
```diff
@@ -752,14 +753,6 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
 
     /* Call the appropriate meta-request new function. */
     if (options->type == AWS_S3_META_REQUEST_TYPE_GET_OBJECT) {
-
-        /* If the initial request already has partNumber, the request is not splittable(?). Treat it as a Default
-         * request.
-         * TODO: Still need tests to verify that the request of a part is splittable or not */
-        if (aws_http_headers_has(initial_message_headers, aws_byte_cursor_from_c_str("partNumber"))) {
```
TingDaoK, line 759: why delete this part? And have you tried to use this branch for aws-crt-cpp and see the CI is happy or not?

ilevyor: mistake, fixing now. How do I run CI on crt-cpp with s3 set to this branch?

ilevyor: BYO_CRYPTO works on crt-cpp with s3 on this branch.
https://github.com/awslabs/aws-crt-cpp/runs/3107882821?check_suite_focus=true

source/s3_client.c
```diff
@@ -337,6 +337,7 @@ struct aws_s3_client *aws_s3_client_new(
         } else {
 #ifdef BYO_CRYPTO
             AWS_ASSERT(false);
+            goto on_error;
```
TingDaoK, line 340: Not sure we want to handle this error.
- If we want to crash in this case, use `AWS_FATAL_ASSERT(false);`
- If we want to handle it more properly, add `aws_raise_error(); /* Some proper error */`

TingDaoK (approved):

justinboswell (approved):

PR #145, opened 2021-07-28
remove sba allocator

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/aws/s3/private/s3_client_impl.h
+++ b/include/aws/s3/private/s3_client_impl.h
@@ -137,9 +137,6 @@ struct aws_s3_client_vtable {
 struct aws_s3_client {
     struct aws_allocator *allocator;
 
-    /* Small block allocator for our small allocations. */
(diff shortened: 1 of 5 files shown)


justinboswell:
Do we have the ability to test this against the existing benchmarks? If we can't do it ahead of check-in, we should certainly verify it after. It's an easy fix if we need to address those languages' allocators.

justinboswell (approved):

TingDaoK:
> Do we have the ability to test this against the existing benchmarks? If we can't do it ahead of check-in, we should certainly verify it after. It's an easy fix if we need to address those languages' allocators.

Our dashboard is still blocked by this issue https://github.com/awslabs/aws-c-s3/pull/145#issuecomment-888693118

TingDaoK:
Tested the sba_allocator doesn't have huge impact to the performance. Merging this.

PR #150, opened 2021-09-20
support proxy envrionment variable from aws-c-http

TingDaoK:
*Issue #, if available:*

- The previous proxy environment variable support is insufficient and wrong

*Description of changes:*

- Remove the support for proxy environment variable from S3 endpoint
- Use the support from underlying lib to support proxy env var


justinboswell (approved):

saso5:
@TingDaoK  I think AWS_ERROR_S3_UNSUPPORTED_PROXY_SCHEME and AWS_ERROR_S3_PROXY_PARSE_FAILED constants can now also be removed.

TingDaoK:
> @TingDaoK I think AWS_ERROR_S3_UNSUPPORTED_PROXY_SCHEME and AWS_ERROR_S3_PROXY_PARSE_FAILED constants can now also be removed.

yeah, thank you for pointing it out. But we are trying to make sure we don't break our downstream binding that may check the number of the error code. So, we will not remove the error code even though it's not used anywhere.

saso5:
A OK, makes sense although you did remove the AWS_ERROR_S3_PROXY_ENV_NOT_FOUND.  

Anyway, I've just stumble upon this while looking into how to set proxy to the s3 client and was a bit confused so I wanted to mention it :). 
Tnx for the replay!

PR #162, opened 2021-12-02
Enable headers callback invocation for failed requests

cmello:
For Default meta requests, the response headers callback was called
only for successful responses.

This commit changes that behavior, so the headers callback is invoked
also on error responses to allow CRT customers to obtain
more information about the error.

*Issue #, if available:*

*Description of changes:*


source/s3_default_meta_request.c
```diff
@@ -254,7 +254,7 @@ static void s_s3_meta_request_default_request_finished(
     struct aws_s3_meta_request_default *meta_request_default = meta_request->impl;
     AWS_PRECONDITION(meta_request_default);
 
-    if (error_code == AWS_ERROR_SUCCESS && meta_request->headers_callback != NULL &&
+    if (meta_request->headers_callback != NULL && request->send_data.response_headers != NULL &&
```
TingDaoK, line 257: - At what case, the response headers will be null?
- What about get_meta_request/put_meta_request? Do those have the same issue?

cmello: Thanks for looking! 

1) The response headers will be null for errors that prevent the HTTP request to be executed (e.g. error codes like `AWS_IO_DNS_INVALID_NAME`). In this case, the headers callback won't be invoked.
2) Get/Put meta requests are composed of sub-requests, and each sub-request could have distinct errors. Still evaluating how granular the error reporting of the parent meta request needs to be. I suggest handling them in a separate PR unless it could impact our decision to call or not to call the headers callback on errors for `Default` meta requests. Any thoughts?

TingDaoK: 1. We should add some documentations [here](https://github.com/awslabs/aws-c-s3/blob/main/include/aws/s3/s3_client.h#L35-L62) or [here](https://github.com/awslabs/aws-c-s3/blob/main/include/aws/s3/s3_client.h#L142-L152) like what we did for [http](https://github.com/awslabs/aws-c-http/blob/main/include/aws/http/request_response.h#L130-L188), which will help others to understand how to callbacks will be invoked. 
2. I would like to have a consistent behavior for the meta requests. It may be not too bad to add the handle for get/put in this PR(This PR is relative small). I think we offer the header callback when error happens is a right thing. The put/get will also have a combined header to be offered to the user.

cmello: Thank you for the references! I'll address both issues in this PR.

include/aws/s3/s3_client.h
```diff
@@ -24,20 +24,57 @@ struct aws_s3_request;
...
+     * to improve throughput, when possible.
+     */
     AWS_S3_META_REQUEST_TYPE_PUT_OBJECT,
 
     AWS_S3_META_REQUEST_TYPE_MAX,
 };
 
+/**
+ * Invoked to provide response headers received during execution of the meta request.
+ * If an error prevents the execution of the HTTP request (e.g. error codes like
```
TingDaoK, line 59: hmmm, just thought about "If we don't ever send a request, there will not be headers for sure." So, suggestions:
```c
/**
 * Invoked to provide response headers received during execution of the meta request, both for success and  
 * error HTTP status codes.
 *
 * Return AWS_OP_SUCCESS to continue processing the request.
 * Return AWS_OP_ERR to indicate failure and cancel the request.
 */
```

source/s3_client.c
```diff
@@ -1397,6 +1397,9 @@ static void s_s3_client_acquired_retry_token(
 
     AWS_ASSERT(client->vtable->acquire_http_connection);
 
+    /* client needs to be kept alive until s_s3_client_on_acquire_http_connection completes */
+    aws_s3_client_acquire(client);
```
TingDaoK, line 1401: Discussed offline:
Feel like this is a life time problem between aws_s3_endpoint and aws_s3_client. We need to dig more into it, but probably not a blocker. We should add a TODO here in the code.

source/s3_auto_ranged_get.c
```diff
@@ -520,6 +520,15 @@ static void s_s3_auto_ranged_get_request_finished(
     bool found_object_size = false;
     bool request_failed = error_code != AWS_ERROR_SUCCESS;
 
+    if (meta_request->headers_callback != NULL && request_failed && request->send_data.response_headers != NULL) {
```
TingDaoK, line 523: Will the `discovers_object_size` always be 0 when the `request_failed` is set? Otherwise, it feels like it's possible to invoke the header callback twice.

tests/s3_data_plane_tests.c
```diff
@@ -2519,6 +2520,121 @@ static int s_test_s3_default_fail_headers_callback(struct aws_allocator *allocat
...
+    const struct aws_http_headers *headers,
+    int response_status,
+    void *user_data) {
+    (void)meta_request;
+    (void)headers;
+    (void)response_status;
+    (void)user_data;
+
+    /* sets flag used to check if callback was invoked */
+    aws_atomic_store_int(&s_test_headers_callback_invoked, 1);
```
TingDaoK, line 2536: Maybe increment the atomic to check the callback only invoked once?

include/aws/s3/s3_client.h
```diff
@@ -139,13 +179,23 @@ struct aws_s3_meta_request_options {
...
     void *user_data;
 
-    /* Callback for receiving incoming headers. */
+    /* Invoked to provide response headers received during execution of the meta request.
+     * If an error prevents the execution of the HTTP request (e.g. error codes like
+     * AWS_IO_DNS_INVALID_NAME), the callback won't be invoked. Otherwise the callback will be invoked,
+     * both for success and error HTTP status codes.
+     *
+     * Return AWS_OP_SUCCESS to continue processing the request.
+     * Return AWS_OP_ERR to indicate failure and cancel the request. */
```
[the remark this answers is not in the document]
cmello: Fixed, thanks!

include/aws/s3/s3_client.h
cmello: Fixed as suggested.

tests/s3_data_plane_tests.c
cmello: Great suggestion, changed to a counter and improved test to check the callback is invoked exactly once.

include/aws/s3/s3_client.h
```diff
@@ -139,13 +177,21 @@ struct aws_s3_meta_request_options {
...
+     * See `aws_s3_meta_request_headers_callback_fn`.
+     */
     aws_s3_meta_request_headers_callback_fn *headers_callback;
 
-    /* Callback for incoming body data. */
+    /*
+     * Invoked to provide the request body as it is received.
+     *
+     * Return AWS_OP_SUCCESS to continue processing the request.
+     * Return AWS_OP_ERR to indicate failure and cancel the request. */
```
TingDaoK, line 191: use reference will be better

include/aws/s3/s3_client.h
```diff
@@ -139,13 +177,21 @@ struct aws_s3_meta_request_options {
...
     aws_s3_meta_request_receive_body_callback_fn *body_callback;
 
-    /* Callback for when the meta request is completely finished. */
+    /* Invoked when the entire meta request execution is complete. */
```
TingDaoK, line 194: ditto

TingDaoK (commented):
I will suggest another solution:
you can add the following code to [here](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_meta_request.c#L1147)
```c
    if (meta_request->headers_callback && finish_result.error_response_headers) {
        if (meta_request->headers_callback(
                meta_request,
                finish_result.error_response_headers,
                finish_result.response_status,
                meta_request->user_data)) {
            /* TODO: decide whether to override the existing error code */
            finish_result.error_code = aws_last_error_or_unknown();
        }
        meta_request->headers_callback = NULL;
    }
```
And remove those change around invoking header callbacks while error happens. But remember to unlink the headers callback (`meta_request->headers_callback = NULL;`) once it's invoked. In case error happens from the callback will invoke the header callback twice. 

The reason I suggest this way is:
- The auto put can probably failed from different type and seems like we only invoke the headers callback when completed part finishes, right?
- All the meta requests can share the same `finish` call.
- I think it's fine to invoke the headers callback for error right before the finish callback.

tests/s3_data_plane_tests.c
```diff
@@ -2519,6 +2520,121 @@ static int s_test_s3_default_fail_headers_callback(struct aws_allocator *allocat
...
+    aws_atomic_init_int(&s_test_headers_callback_invoked, 0);
+
+    struct aws_s3_tester_meta_request_options options = {
+        .allocator = allocator,
+        .meta_request_type = AWS_S3_META_REQUEST_TYPE_PUT_OBJECT,
+        .headers_callback = s_s3_test_headers_callback_check,
+        .validate_type = AWS_S3_TESTER_VALIDATE_TYPE_EXPECT_FAILURE,
+        .put_options =
+            {
+                .object_size_mb = 1,
```
TingDaoK, line 2624: The default client part size is 8 MB, when we send put with size less than part size, we will not create auto_put meta request as we don't need to split the request. Check code [here](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L812-L820). So, your test can pass as the default meta request is used here. Maybe similar issue for the get one. I will leave it to you to dig into it😬

Change the size to 10, and you solution here will fail.

TingDaoK: Also, as @JonathanHenson mentioned before, we don't have unit tests :(. It's hard to test the error from each type of put requests. It's a real concern... But, if you chose my suggestion, we can add tests like when the request has a success response code, and user error out the request from callback, make sure our headers callback will not be invoked again.

cmello: > I think it's fine to invoke the headers callback for error right before the finish callback.

That's a good approach, thanks! Working on it.

cmello: > The default client part size is 8 MB. When we send put with size less than part size, we will not create auto_put meta request as we don't need to split the request. Check code here. So, your test can pass as the default meta request is used here. 

Great catch, that explains why the Put meta request test is passing. Changed `object_size_mb` to 10 MB as suggested and the test fails. Fixing...

source/s3_auto_ranged_get.c
cmello: Moved callback invocation to meta request finish handler as suggested in the other comment. Thanks!

include/aws/s3/s3_client.h
cmello: Fixed.

source/s3_meta_request.c
```diff
@@ -1145,6 +1145,18 @@ void aws_s3_meta_request_finish_default(struct aws_s3_meta_request *meta_request
         aws_s3_request_release(release_request);
     }
 
+    if (meta_request->headers_callback && finish_result.error_response_headers) {
+        if (meta_request->headers_callback(
+                meta_request,
+                finish_result.error_response_headers,
+                finish_result.response_status,
+                meta_request->user_data)) {
+            /* TODO: decide whether to override the existing error code */
```
TingDaoK, line 1154: oops, I put the todo just as I don't know which is better. This code is override the error code using the one from callback, I feel like it's better when I wrote it. If you agree, we can just remove this todo.

source/s3_default_meta_request.c
```diff
@@ -255,6 +255,7 @@ static void s_s3_meta_request_default_request_finished(
     AWS_PRECONDITION(meta_request_default);
 
     if (error_code == AWS_ERROR_SUCCESS && meta_request->headers_callback != NULL &&
+        request->send_data.response_headers != NULL &&
```
TingDaoK, line 258: trivial: remove this change, when the error code is success, there should always be response headers.

source/s3_default_meta_request.c
```diff
@@ -255,6 +255,7 @@ static void s_s3_meta_request_default_request_finished(
...
         meta_request->headers_callback(
```
TingDaoK, line 259: add `meta_request->headers_callback = NULL;` In case the error happens from the callback, and invoked the finish call. And, as we are not using the callback anymore, it's safer to just reset it.
Same for s3_auto_ranged_get/put as well. Just clear the callback once its invoked.

source/s3_meta_request.c
cmello: Removed TODO, thanks!

source/s3_default_meta_request.c
cmello: Letting the check just to be 100% safe a manufactured response with status code 200 without headers wouldn't cause trouble.

source/s3_default_meta_request.c
cmello: Added for `default_meta_request`, `auto_ranged_get` and `auto_ranged_put`. Thanks!

TingDaoK (approved):
Thanks! Good job

PR #166, opened 2021-12-14
CopyObject Meta request implementation

cmello:
This meta request type performs either a multi-part copy
using multiple `UploadPartCopy` calls in parallel, or bypasses
a `CopyObject` request to S3 if the object size is not large enough for
a multi-part upload.

*Issue #, if available:*

*Description of changes:*


samples/s3-cp/s3-cp.c
```diff
@@ -0,0 +1,403 @@
...
+                (char *)source_uri_cursor.ptr,
+                aws_error_debug_str(aws_last_error()));
+            s_usage(1);
+        };
+    } else {
+        fprintf(stderr, "An URI for the source object must be supplied.\n");
+        s_usage(1);
+    }
+
+    if (aws_cli_optind < argc) {
```
JonathanHenson, line 115: don't you need to check that both of these branches were taken?

samples/s3-cp/s3-cp.c
```diff
@@ -0,0 +1,403 @@
...
+    const struct aws_byte_cursor *body,
+    uint64_t range_start,
+    void *user_data) {
+
+    (void)meta_request;
+    (void)body;
+    (void)range_start;
+    (void)user_data;
+
+    return 0;
```
JonathanHenson, line 205: if the semantics of this callback are (use all of it or none of it) and return success or error, return AWS_OP_SUCCESS directly so that's clear. If it's the opposite, this might be a bug on the tcp back pressure stuff and should be checked.

source/s3_copy_object.c
```diff
@@ -0,0 +1,728 @@
...
+    struct aws_s3_copy_object *copy_object = meta_request->impl;
+    AWS_PRECONDITION(copy_object);
+
+    aws_s3_meta_request_lock_synced_data(meta_request);
+
+    struct aws_http_message *message = NULL;
+
+    switch (request->request_tag) {
+
+        case AWS_S3_COPY_OBJECT_REQUEST_TAG_GET_OBJECT_SIZE: {
```
JonathanHenson, line 354: reasoning for what each case means above each case statement would be helpful

tests/s3_data_plane_tests.c
```diff
@@ -3431,3 +3432,200 @@ static int s_test_s3_not_satisfiable_range(struct aws_allocator *allocator, void
...
+    aws_http_message_destroy(message);
+    aws_s3_client_release(client);
+    client = NULL;
+
+    aws_s3_tester_clean_up(&tester);
+
+    return 0;
+}
+
+AWS_TEST_CASE(test_s3_copy_small_object, s_test_s3_copy_small_object)
```
JonathanHenson, line 3617: it seems to me that the state machinery around which network path is taken had more variations than just small and large objects. Are you sure at least all the branches that can be taken without a proper wire mocking system are taken?

source/s3_copy_object.c
```diff
@@ -0,0 +1,728 @@
...
+    /* part size and content length will be fetched later using a HEAD object request */
+    const size_t UNKNOWN_PART_SIZE = 0;
+    const size_t UNKNOWN_CONTENT_LENGTH = 0;
+    const int UNKNOWN_NUM_PARTS = 0;
+
+    if (aws_s3_meta_request_init_base(
+            allocator,
+            client,
+            UNKNOWN_PART_SIZE,
+            client->compute_content_md5 == AWS_MR_CONTENT_MD5_ENABLED ||
```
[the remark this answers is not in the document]
cmello: Fixed, thanks!

source/s3_copy_object.c
```diff
@@ -0,0 +1,728 @@
...
+    bool work_remaining = false;
+
+    struct aws_s3_copy_object *copy_object = meta_request->impl;
+
+    aws_s3_meta_request_lock_synced_data(meta_request);
+
+    if (!aws_s3_meta_request_has_finish_result_synced(meta_request)) {
+
+        /* If we haven't already sent the GetObject HEAD request to get the source object size, do so now. */
+        if (!copy_object->synced_data.head_object_sent) {
```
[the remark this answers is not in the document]
cmello: I agree, and this code is currently very difficult to read. I'm looking for a way to break it into smaller pieces. Thank you for the feedback!

source/s3_copy_object.c
```diff
@@ -7,6 +7,7 @@
 #include "aws/s3/private/s3_request_messages.h"
 #include "aws/s3/private/s3_util.h"
 #include <aws/common/string.h>
+#include <aws/http/private/strutil.h>
```
JonathanHenson, line 10: this is a layering violation: private headers can only be used inside the module that declares them. We should refactor this so it can be more generic (possibly in aws-c-common)

source/s3_copy_object.c
```diff
@@ -0,0 +1,728 @@
...
+            client->compute_content_md5 == AWS_MR_CONTENT_MD5_ENABLED ||
+                aws_http_headers_has(aws_http_message_get_headers(options->message), g_content_md5_header_name),
+            options,
+            copy_object,
+            &s_s3_copy_object_vtable,
+            &copy_object->base)) {
+        goto error_clean_up;
+    }
+
+    if (aws_array_list_init_dynamic(
```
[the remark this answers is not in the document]
cmello: It checks for invalid parameters only, but doesn't return on allocation failures. Removed.

include/aws/s3/private/s3_util.h
```diff
@@ -165,6 +165,29 @@ void aws_s3_get_part_range(
...
+ * Rejects things like:
+ * "-1" // negative numbers not allowed
+ * "1,000" // only characters 0-9 allowed
+ * "" // blank string not allowed
+ * " 0 " // whitespace not allowed
+ * "0x0" // hex not allowed
+ * "FF" // hex not allowed
+ * "999999999999999999999999999999999999999999" // larger than max u64
+ */
+AWS_S3_API
```
JonathanHenson, line 188: Pretty sure this will break on certain compilers depending on linker flags due to symbol conflicts. I’d in-line it where you need it until it’s moved to aws-c-common

samples/s3-cp/s3-cp.c
cmello: Removed those callback handlers from sample. They were being used only for debugging and are not needed.

tests/s3_data_plane_tests.c
cmello: Added one more test checking for failure due to invalid source key. The intermediary request failures are not being tested currently and will require a proper wire mocking system. I'll handle the mocking in a distinct PR.

source/s3_copy_object.c
cmello: Created a task to move `aws_strutil_read_unsigned_num` to `aws-c-common` to enable reuse between `aws-c-http` and `aws-c-s3`. The refactor will be implemented in distinct PRs.

include/aws/s3/private/s3_util.h
cmello: Renamed to `aws_s3_strutil_read_unsigned_num` to avoid name clash with `aws_s3_strutil_read_unsigned_num`, until the refactor is done. As part of the refactor task I'll delete the function definition in `aws-c-http` and this corresponding copy `aws_s3_strutil_read_unsigned_num` in `aws-c-s3`, and fix callers of the function in `aws-c-http` and `aws-c-s3` so they use the version from `aws-c-common`.

samples/s3-cp/s3-cp.c
cmello: It is currently expecting the last two parameters to be the source and target URIs respectively. If any is missing, it will print the corresponding error.

source/s3_copy_object.c
cmello: Added comments for each case label, thanks!

JonathanHenson (approved):

PR #170, opened 2022-01-22
Add CopyObject progress callback and update cli sample

cmello:
*Issue #, if available:*

*Description of changes:*
* Added progress callback for CopyObject meta requests
* Updated CLI sample to show the s3 to s3 copy in the progress bar
* Refactored the code to extract the `etag` from UploadPartCopy response into a distinct function

### DIFF
--- a/include/aws/s3/s3_client.h
+++ b/include/aws/s3/s3_client.h
@@ -229,7 +229,9 @@ struct aws_s3_meta_request_options {
     aws_s3_meta_request_shutdown_fn *shutdown_callback;
 
     /**
-     * Invoked to report progress of multi-part upload and copy object requests.
+     * Invoked to report progress of the meta request execution.
+     * Currently, the progress callback is invoked only for the CopyObject meta request type.
+     * TODO: support this callback for all the types of meta requests
      * See `aws_s3_meta_request_progress_fn`
      */
     aws_s3_meta_request_progress_fn *progress_callback;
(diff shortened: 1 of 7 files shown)


source/s3_copy_object.c
```diff
@@ -540,6 +540,37 @@ static int s_s3_copy_object_prepare_request(struct aws_s3_meta_request *meta_req
     return AWS_OP_ERR;
 }
 
+/* For UploadPartCopy requests, etag is sent in the request body, within XML entity quotes */
+static struct aws_string *s_get_etag_from_upload_part_copy_response(
```
TingDaoK, line 544: trivial: maybe rename to `s_etag_new_from_upload_part_copy_response`? Since this function creates a new `aws_string`, which will be a bit more clear about the lifetime.

include/aws/s3/s3_client.h
```diff
@@ -206,6 +227,12 @@ struct aws_s3_meta_request_options {
 
     /* Callback for when the meta request has completely cleaned up. */
     aws_s3_meta_request_shutdown_fn *shutdown_callback;
+
+    /**
+     * Invoked to report progress of multi-part upload and copy object requests.
+     * See `aws_s3_meta_request_progress_fn`
+     */
+    aws_s3_meta_request_progress_fn *progress_callback;
```
TingDaoK, line 235: I don't see any where related "multi-part upload" in this PR. Is that something "TODO"? And, what about get object? What if people set this callback for get object request? Do this callback just never be invoked or error out?

I think we could try to support this callback for all the types of meta requests we have. We can add TODO here, if it's something not related to this PR.

cmello: Good point! Updated the comment to reflect the current behavior and added `TODO` to add support for the other meta request types (to be addressed in distinct CRs). Thanks!

TingDaoK (approved):
The PR looks good. I just a bit worry about how can we test it. I know you proved it working via the sample, but it will be better if we can have some unit/integration test for it. 

But proper test is a huge task right in front of us... I am okay to ship it and revisit the testing with all those other tests we need...

PR #174, opened 2022-02-09
Support meta requests on HTTP and HTTPS with custom ports

aceeric:
*Issue #, if available:* https://github.com/awslabs/aws-crt-java/issues/431

*Description of changes:*

This works in tandem with pull request: https://github.com/awslabs/aws-crt-java/pull/435

The purpose is to allow meta requests over both http and https, and also to support custom ports in addition to the 80/443 hard-codes.


include/aws/s3/s3_client.h
```diff
@@ -228,6 +228,12 @@ struct aws_s3_meta_request_options {
     /* Callback for when the meta request has completely cleaned up. */
     aws_s3_meta_request_shutdown_fn *shutdown_callback;
 
+    /* Support meta requests over HTTP and HTTPS */
+    enum aws_s3_meta_request_tls_mode tls_mode;
```
TingDaoK, line 232: Do we need this tls_mode override the configuration for the whole client?

TingDaoK: Okay, I read thought the issue again. Well, I think we should move the logic of taking an URL to the C part instead. So, the java will just create a C URL from java URL and pass in here. You can use the structure [here](https://github.com/awslabs/aws-c-io/blob/main/include/aws/io/uri.h#L15)

aceeric: @TingDaoK Thank you. Before I begin work on the changes, I would like to make sure I'm clear on your intent.

Presently (before my patch) the `S3Client.java` / `makeMetaRequest` method receives an instance of `software.amazon.awssdk.crt.http.HttpRequest` containing a host header. The `HttpRequest` is passed to the JNI as a struct, and the host name in the request header eventually makes its way down to the `s_s3_endpoint_create_http_connection_manager` function.

If I understand - your direction for this PR is to have the `S3Client.java` / `makeMetaRequest` method accept a URI string. The Java method would pass that string to the JNI `...s3ClientMakeMetaRequest` function, which would parse the string into the struct you referenced, and pass that struct down the call chain where it would be used instead of the http request host header. At some point, a decision needs to be made how far down the call chain to pass that struct intact vs. picking it apart passing the pieces as function args. But I can work through that. (Probaby the struct should go all the way to the `s_s3_endpoint_create_http_connection_manager` function and replace any duplicative args along the way.)

Please let me know if I captured your intent. Thank you.

aceeric:
@TingDaoK Hello I updated this PR and the corresponding one here: https://github.com/awslabs/aws-crt-java/pull/435 to pass the URI into the C code.

include/aws/s3/s3_client.h
```diff
@@ -10,6 +10,7 @@
 #include <aws/s3/s3.h>
 
 #include <aws/auth/signing_config.h>
+#include "aws/io/uri.h"
```
TingDaoK, line 13: Trivial: You don't need to include this at header file. Using forward declaration, you can move this include to `s3_client.c` instead

include/aws/s3/s3_client.h
```diff
@@ -198,6 +200,9 @@ struct aws_s3_meta_request_options {
 
     /* Callback for when the meta request has completely cleaned up. */
     aws_s3_meta_request_shutdown_fn *shutdown_callback;
+
+    /* If present, then overrides HTTP Request host header */
```
TingDaoK, line 204: That confuses me without read the implementation. We don't use the URI to update the host header right?
Also, what if the URI is set, and host is set as well? Who takes in charge? _Should it be error if they both set and different with each other?_ TBH, I don't know what's the right answer here, maybe it's fine to let the URI to take in charge as user set it specifically.
Suggestion:
```
/** 
 * Optional.
 * The endpoint for the meta request to connect to. If not set, the "Host" header from `message` will be used instead.
 **/
```

source/s3_client.c
```diff
@@ -609,29 +609,54 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
...
+                aws_string_destroy(endpoint_host_name);
+            }
+            endpoint_host_name = uri_host_name;
+        }
+        struct aws_byte_cursor https_scheme = aws_byte_cursor_from_c_str("https");
+        is_https = aws_byte_cursor_eq(aws_uri_scheme(options->uri), &https_scheme);
+        port = aws_uri_port(options->uri);
+    }
+
+    if (endpoint_host_name == NULL) {
```
TingDaoK, line 647: Is it valid to have a request without host header for S3? Will server 500 us? I think we should still follow the logic before (Check the Host header exist or not).

source/s3_client.c
```diff
@@ -649,10 +674,11 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
             .ref_count_zero_callback = client->vtable->endpoint_ref_count_zero,
             .shutdown_callback = client->vtable->endpoint_shutdown_callback,
             .client_bootstrap = client->client_bootstrap,
-            .tls_connection_options = client->tls_connection_options,
+            .tls_connection_options = is_https ? client->tls_connection_options : NULL,
```
TingDaoK, line 677: I think you requested to expose the client config to Java, right? That's totally reasonable. 

For different meta request, we can use different endpoint to connect, so, it makes sense to override the tls options from the client. 
We can probably add customized tls when needed.
I think that's fine, but we need to document it out.

include/aws/s3/s3_client.h
aceeric: In the Java (above JNI) I replace the host header if the uri is provided. The reason is - if the host header is incorrect (or - as in my testing I used constant string "INVALID") then the connection fails. I looked but was not able to easily see why. It made sense that the Java should force the host header to match the host in the uri - so that's the approach I took.  It's clear that some other path through the code relies on the host header, not that endpoint arg that is passed around. I felt  it would be too extensive of a change to try to find and replace that functionality.

include/aws/s3/s3_client.h
aceeric: ok

source/s3_client.c
aceeric: It is not valid - in this implementation - host header is required and is enforced to match uri if uri is supplied

source/s3_client.c
aceeric: I have tested with custom TLS context against Minio, and built-in TLS against <my-bucket>.us-east-1.s3.amazonaws.com

source/s3_client.c
TingDaoK: > host header is required and is enforced to match uri if uri is supplied

I don't think the code does any enforcement here, did I miss something?

include/aws/s3/s3_client.h
TingDaoK: Well, if the host header is "incorrect/invalid", the user who pass the http request to java should be the one to fix it, right? I don't think overwrite the host header is a right way to do even for java side.

aceeric: I would add one observation: replacing the host header because of the URI makes that change visible to the caller of the `makeMetaRequest` method of `S3Client.java`. This might be surprising to the caller - they pass host header 'foo' and it changes to 'bar' on return. If that's undesirable, other options are 1) clone/modify the `HttpRequest` in the `makeMetaRequest` method and pass the clone to the JNI so the caller's original is untouched 2) error if uri hostname != http request host header 3) disallow host headers in the http request if URI is also provided, 4) other? Fundamentally, allowing the caller to supply URI creates the potential for a clash with http request host header so maybe some additional handling is indicated...

source/s3_client.c
aceeric: Just to clarify - I think you are recommending that the `TlsContext` be supported in the `S3MetaRequestOptions` and then each meta request can use a different TLS Context. Is that correct? If so I will modify the PR. (Result would be an additional arg `tlsContext` to the JNI `s3ClientMakeMetaRequest` method.)

include/aws/s3/s3_client.h
TingDaoK: I agree on those besides

>uri hostname non null && HttpRequest -> host header null = ok - C code constructs host header from uri hostname for lower layers in the call chain

I don't like the client to modify the passed in request. Our S3 client is not responsible for correcting the request user passed in. So, I'd just error out at that case.

source/s3_client.c
TingDaoK: Nope, I think for now, it's fine. But, if anyone requests to add customized TLS for each meta request, it's also reasonable. Let's not do it until anyone asks for it.

source/s3_client.c
aceeric: well - that is (currently) in the Java but as discussed above, manipulation of host header will be removed from the Java and the enforcement will be as discussed  [above](https://github.com/awslabs/aws-c-s3/pull/174#discussion_r807194886)

include/aws/s3/s3_client.h
aceeric: Pushed updates in this repo and for https://github.com/awslabs/aws-crt-java/pull/435

TingDaoK: two trivial things:
- Can you rename this to `endpoint`? I think it's more informative than `uri`.
- Can we update the documentation to something like:
```
/** 
 * Optional. 
 * The endpoint for the meta request to connect to. 
 * If not set, the tls settings from client will be used to determine the port and host header from `message` determine the host to connect.
 * Note: need to meet the host header from `message`
 **/
```

TingDaoK (approved):
Looks good, thank you!

include/aws/s3/s3_client.h
aceeric: changed in latest commit

source/s3_endpoint.c
```diff
@@ -50,7 +50,8 @@ static struct aws_http_connection_manager *s_s3_endpoint_create_http_connection_
     const struct aws_string *host_name,
     struct aws_client_bootstrap *client_bootstrap,
     const struct aws_tls_connection_options *tls_connection_options,
-    uint32_t max_connections);
+    uint32_t max_connections,
+    int port);
```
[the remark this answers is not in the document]
aceeric: changed in latest commit

include/aws/s3/private/s3_client_impl.h
```diff
@@ -65,6 +65,9 @@ struct aws_s3_endpoint_options {
 
     /* Maximum number of connections that can be spawned for this endpoint. */
     uint32_t max_connections;
+
+    /* HTTP port override. If zero, determine port based on TLS context */
+    int port;
```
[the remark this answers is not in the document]
aceeric: changed in latest commit

source/s3_client.c
aceeric: changed in latest commit

PR #178, opened 2022-02-17
Support leading slash in x-amz-copy-source

cmello:
The CopyObject meta request required the x-amz-copy-source to be in strict form of bucket/key.
But as per S3 docs, /bucket/key is also valid.

This fix adds support for both formats.

*Issue #, if available:*

*Description of changes:*


source/s3_request_messages.c
```diff
@@ -407,14 +407,20 @@ struct aws_http_message *aws_s3_get_source_object_size_message_new(
...
+        source_bucket.ptr = &decode_buffer.buffer[1];
+    } else {
+        source_bucket.ptr = decode_buffer.buffer;
+    }
+
+    /* as we skipped the optional leading slash, from this point source format is always {bucket}/{key}. split them. */
+    for (uint8_t *p = source_bucket.ptr; p < (decode_buffer.buffer + decode_buffer.len); p++) {
+        if (*p == '/') {
+            source_bucket.len = p - source_bucket.ptr;
+            source_key.ptr = p + 1; /* skip the / */
```
JonathanHenson, line 422: Is this pointer arithmetic safe? Can we use byte_cursor_advance instead?

cmello: Great idea, thanks! I'll switch to byte_cursor instead of raw pointer.

cmello: Changed to `aws_byte_cursor_advance`.

source/s3_request_messages.c
```diff
@@ -402,24 +402,28 @@ struct aws_http_message *aws_s3_get_source_object_size_message_new(
...
-    /* source format is {bucket}/{key}. split them */
-    for (size_t i = 0; i < decode_buffer.len; i++) {
-        if (decode_buffer.buffer[i] == '/') {
-            source_bucket.len = i;
-            source_key.ptr = decode_buffer.buffer + i + 1; /* skip the / */
-            source_key.len = decode_buffer.len - i - 1;
+    /* extracts source bucket and key */
+    struct aws_byte_cursor source_bucket = aws_byte_cursor_from_buf(&decode_buffer);
+
+    /* x-amz-copy-source might have an optional leading slash. if so let's skip it */
```
TingDaoK, line 408: Is it an error for get_object to have a leading slash? This function is also used for get_object right? Will this break the current behavior for that?

cmello: This function is not used by get_object. It is used only by Copy meta requests to extract info from the source object specified in the `x-amz-copy-source` header. Maybe I should rename the function to `aws_s3_get_copy_source_object_size_message_new` to avoid confusion? The format of the header value can be either `/{bucket}/{key}` or `{bucket}/{key}` as described in the [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html). Thanks for checking!

TingDaoK: If it's only used for copy object, that's fine. Yeah, maybe rename or update the documentation for this function to avoid the confusion. I saw the documentation saying:
```
/* Creates a HEAD GetObject request to get the size of the source object. */
```
So, I assume GetObject also uses this function.

TingDaoK (approved):
LGTM. Fix or not & ship.

PR #181, opened 2022-02-21
Checksum feature

ilevyor:
*Issue #, if available:*

*Description of changes:*


source/s3_auto_ranged_get.c
```diff
@@ -154,162 +130,148 @@ static bool s_s3_auto_ranged_get_update(
...
-
-            /* auto-ranged-gets make use of body streaming, which will hold onto response bodies if parts earlier in
-             * the file haven't arrived yet. This can potentially create a lot of backed up requests, causing us to
-             * hit our global request limit. To help mitigate this, when the "conservative" flag is passed in, we
-             * only allow the total amount of requests being sent/streamed to be inside of a set limit.  */
-            if (num_requests_in_flight > s_conservative_max_requests_in_flight) {
-                goto has_work_remaining;
+    /* BEGIN CRITICAL SECTION */
+    {
+        aws_s3_meta_request_lock_synced_data(meta_request);
```
TingDaoK, line 135: WE left here

include/aws/s3/private/s3_checksums.h
```diff
@@ -0,0 +1,111 @@
...
+ * Get the multipart upload header name corresponding to the aws_s3_checksum_algorithm enum value.
+ */
+AWS_S3_API
+const struct aws_byte_cursor *aws_get_create_mpu_header_name_from_algorithm(enum aws_s3_checksum_algorithm algorithm);
+
+/**
+ * Get the complete multipart upload name corresponding to the aws_s3_checksum_algorithm enum value.
+ */
+AWS_S3_API
+const struct aws_byte_cursor *aws_get_complete_mpu_name_from_algorithm(enum aws_s3_checksum_algorithm algorithm);
```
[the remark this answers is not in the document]
ilevyor: No, this isn't in the header, it's found in the complete mpu body

source/s3_meta_request.c
```diff
@@ -716,6 +905,11 @@ static int s_s3_meta_request_incoming_headers(
     bool successful_response =
         s_s3_meta_request_error_code_from_response_status(request->send_data.response_status) == AWS_ERROR_SUCCESS;
 
+    if (successful_response && connection->request->meta_request->type == AWS_S3_META_REQUEST_TYPE_GET_OBJECT &&
+        connection->request->request_tag == AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_PART) {
+        s_get_part_response_checksum_callback(connection, headers, headers_count);
```
TingDaoK, line 910: they are not callbacks, so, renaming it to `helper`?

source/s3_meta_request.c
```diff
@@ -843,25 +1043,39 @@ void aws_s3_meta_request_send_request_finish_default(
...
         (void *)request,
         error_code,
+        aws_error_debug_str(error_code),
         response_status);
 
     enum aws_s3_connection_finish_code finish_code = AWS_S3_CONNECTION_FINISH_CODE_FAILED;
 
     if (error_code == AWS_ERROR_SUCCESS) {
+        if (connection->request->meta_request->type == AWS_S3_META_REQUEST_TYPE_GET_OBJECT && request->did_validate &&
+            !request->checksum_match) {
```
TingDaoK, line 1057: probably the `error_code` should be set to the MISMATCH?

source/s3_request_messages.c
```diff
@@ -192,6 +194,14 @@ struct aws_http_message *aws_s3_create_multipart_upload_message_new(
             goto error_clean_up;
         }
     }
+    if (algorithm) {
```
TingDaoK, line 197: If the algorithm is md5, will this be an error?

source/s3_request_messages.c
```diff
@@ -283,12 +296,13 @@ struct aws_http_message *aws_s3_upload_part_copy_message_new(
     }
 
     if (buffer != NULL) {
-        if (aws_s3_message_util_assign_body(allocator, buffer, message) == NULL) {
+        /************************* TODO Update to use Flexible Checksums *********************************************/
```
TingDaoK, line 299: remove it as you said.

source/s3_request_messages.c
```diff
@@ -539,7 +588,7 @@ struct aws_http_message *aws_s3_complete_multipart_message_new(
             goto error_clean_up;
         }
 
-        aws_s3_message_util_assign_body(allocator, body_buffer, message);
+        aws_s3_message_util_assign_body(allocator, body_buffer, message, AWS_SCA_NONE, NULL);
```
TingDaoK, line 591: trivial: we usually do things like `aws_s3_message_util_assign_body(allocator, body_buffer, message, AWS_SCA_NONE, NULL/* out_checksum */);`

source/s3_request_messages.c
```diff
@@ -586,11 +635,22 @@ struct aws_http_message *aws_s3_abort_multipart_upload_message_new(
     return NULL;
 }
 
+static int s_is_get_request(const struct aws_http_message *message, bool *is_request) {
```
TingDaoK, line 638: ~is_request~ to out?

source/s3_request_messages.c
```diff
@@ -586,11 +635,22 @@ struct aws_http_message *aws_s3_abort_multipart_upload_message_new(
...
+    struct aws_byte_cursor method;
+    if (aws_http_message_get_request_method(message, &method)) {
```
TingDaoK, line 640: feel like this should not fail, maybe an ASSERT on it, so, you don't need a return code?

source/s3_request_messages.c
```diff
@@ -608,13 +668,50 @@ struct aws_input_stream *aws_s3_message_util_assign_body(
...
+    if (s_is_get_request(out_message, &is_get)) {
+        goto error_clean_up;
+    }
+    if (is_get) {
+        aws_http_headers_set(headers, g_request_validation_mode, g_enabled);
+    }
+
+    if (algorithm) {
+        /* set Content-Encoding header */
+        if (!is_get && aws_http_headers_set(headers, g_content_encoding_header_name, g_content_encoding_header_value)) {
```
TingDaoK, line 681: `g_content_encoding_header_value = "aws-chunked"`
Rename it, there are multiple encoding types, this is a specific one so naming it after the `aws-chunked`.

source/s3_chunk_stream.c
```diff
@@ -0,0 +1,268 @@
...
+    return AWS_OP_SUCCESS;
+}
+
+static int s_aws_input_chunk_stream_seek(
+    struct aws_input_stream *stream,
+    int64_t offset,
+    enum aws_stream_seek_basis basis) {
+
+    struct aws_chunk_stream *impl = stream->impl;
+    return aws_input_stream_seek(impl->current_stream, offset, basis);
```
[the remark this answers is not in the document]
TingDaoK: Just remember that we have `aws_http1_stream_write_chunk`, why not just use that?

source/s3_copy_object.c
```diff
@@ -407,7 +409,7 @@ static int s_s3_copy_object_prepare_request(struct aws_s3_meta_request *meta_req
 
             /* Create the message to create a new multipart upload. */
             message = aws_s3_create_multipart_upload_message_new(
-                meta_request->allocator, meta_request->initial_request_message);
+                meta_request->allocator, meta_request->initial_request_message, AWS_SCA_NONE);
```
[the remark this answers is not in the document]
ilevyor: Unclear, the smithy model is wrong, created discussion in slack channel

source/s3_chunk_stream.c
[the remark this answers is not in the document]
ilevyor: because, this is not an an http chunk, I wasted a month or two adding trailing headers support to http client, before I realized that brawn used aws-chunked encoding 🤦 which is chunked at the client level.

source/s3_auto_ranged_get.c
```diff
@@ -384,123 +347,70 @@ static int s_discover_object_range_and_content_length(
...
+    /* When doing a head object request, we currently assume that we did so with a message that had a Range
+     * header. This means that there should also be a Content-Range header that specifies the object range and
+     * total object size.*/
+    if (total_content_length == 0) {
+        AWS_LOGF_DEBUG(
+            AWS_LS_S3_META_REQUEST,
+            "id=%p Detected content_len 0 with request %p.",
+            (void *)meta_request,
+            (void *)request);
+    } else if (aws_s3_parse_content_range_response_header(
```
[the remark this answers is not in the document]
ilevyor: S3 documentation is wrong, I cut a ticket to them.

source/s3_copy_object.c
```diff
@@ -358,11 +359,12 @@ static int s_s3_copy_object_prepare_request(struct aws_s3_meta_request *meta_req
...
          * S3. */
         case AWS_S3_COPY_OBJECT_REQUEST_TAG_BYPASS: {
-            message = aws_s3_message_util_copy_http_message(
+            message = aws_s3_message_util_copy_http_message_no_body(
                 meta_request->allocator, meta_request->initial_request_message, NULL, 0);
 
             if (meta_request->should_compute_content_md5) {
-                aws_s3_message_util_add_content_md5_header(meta_request->allocator, &request->request_body, message);
+                aws_s3_message_util_add_checksum_header(
+                    meta_request->allocator, &request->request_body, message, AWS_SCA_MD5);
```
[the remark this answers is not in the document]
ilevyor: This can actually be deleted. Confirmed by @cmello .

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -179,6 +183,12 @@ struct aws_s3_meta_request {
     } client_process_work_threaded_data;
 
     const bool should_compute_content_md5;
+
+    enum aws_s3_checksum_algorithm checksum_algorithm;
+
+    struct aws_checksum *running_response_sum;
+
+    struct aws_byte_buf response_header_checksum;
```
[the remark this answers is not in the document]
ilevyor: This is also used by a default get, so it doesn't make sense for it to be in auto_range_get.

source/s3_checksum_stream.c
```diff
@@ -11,18 +11,24 @@ struct aws_checksum_stream {
...
-    return aws_input_stream_seek(impl->old_stream, offset, basis);
+    (void)stream;
+    (void)offset;
+    (void)basis;
+    AWS_LOGF_ERROR(
+        AWS_LS_S3_CLIENT,
+        "Cannot seek on checksum stream, as it will cause the checksum output to mismatch the checksum of the stream "
+        "contents");
+    AWS_ASSERT(false);
+    return AWS_OP_ERR;
```
TingDaoK, line 31: you need to raise an error. found or define an error to raise here.

source/s3_checksums.c
```diff
@@ -17,31 +17,23 @@ size_t aws_get_digest_size_from_algorithm(enum aws_s3_checksum_algorithm algorit
...
-        case AWS_SCA_MD5:
-            return AWS_MD5_LEN;
         default:
-            AWS_ASSERT(false);
             return 0;
     }
 }
 
-struct aws_byte_cursor aws_get_http_header_name_from_algorithm(enum aws_s3_checksum_algorithm algorithm) {
+const struct aws_byte_cursor *aws_get_http_header_name_from_algorithm(enum aws_s3_checksum_algorithm algorithm) {
```
TingDaoK, line 25: good catch. I didn't notice that

source/s3_chunk_stream.c
```diff
@@ -253,12 +255,13 @@ struct aws_input_stream *aws_chunk_stream_new(
...
      * stream */
-    impl->length = prechunk_stream_len + stream_length + final_chunk_len + impl->checksum_header_name.len + colon_len +
+    impl->length = prechunk_stream_len + stream_length + final_chunk_len + impl->checksum_header_name->len + colon_len +
                    encoded_checksum_len + post_trailer_len - 1;
     AWS_FATAL_ASSERT(impl->current_stream);
     return stream;
 
 error:
+    /* TODO: the stream should refcount, otherwise, this can lead to crash */
     aws_input_stream_destroy(impl->checksum_stream);
```
TingDaoK, line 265: Feel like a bug.
So, if error from place after `impl->checksum_stream == NULL`, this will destroy the `existing_stream`, if error from place before that, it will NOT destroy the `existing_stream`.

The right thing should be if this function fails, it should NEVER destroy the `existing_stream`. Will be easier if we have refcount, but we can still do that to avoid the bug.

TingDaoK (commented):
Post it now to keep you have stuff to work on. Will be back to finish remaining.

include/aws/s3/s3_client.h
```diff
@@ -195,9 +194,27 @@ struct aws_s3_client_config {
...
     enum aws_s3_meta_request_compute_content_md5 compute_content_md5;
 
-    /* If a flexible checksum is specified it will override compute_content_md5 */
+    /* If a flexible checksum is specified it will disable compute_content_md5 */
+    /**
+     * checksum algorithm will cause single part put requests to append a trailing checksum of the body corresponding to
+     * the specified algorithm. Multipart uploads will have a checksum uploaded corresponding to each individual part,
+     * and the body of the complete multipart upload request will contain a list of the checksums from erach part.
+     * Get requests will be unaffected by this setting. Copy object requests will have a header specifying to s3 to
+     * calculate a new checksum with the corresponding algorithm. multipart copies
```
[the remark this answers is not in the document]
ilevyor: what Cesar has been working on, I looked into expected behavior for that. Not necessary now, but will be for Java integration.

include/aws/s3/s3_client.h
```diff
@@ -195,9 +194,27 @@ struct aws_s3_client_config {
...
+     */
     enum aws_s3_checksum_algorithm checksum_algorithm;
 
+    /**
+     * Enable checksumode header will be attached to get requests, this will tell s3 to send back checksums headers if
```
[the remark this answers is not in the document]
ilevyor: I felt the same, but look at the meta_request_options struct vs the client_options struct. Meta request has almost no besides the minimum required.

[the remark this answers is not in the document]
ilevyor: Alternatively I can just add to both.

source/s3_chunk_stream.c
```diff
@@ -0,0 +1,271 @@
...
+            parent_stream->checksum_result_output, aws_default_allocator(), checksum_result_cursor)) {
+        return AWS_OP_ERR;
+    }
+    if (aws_byte_buf_init(
+            &parent_stream->post_chunk_buffer,
+            aws_default_allocator(),
+            final_chunk_cursor.len + parent_stream->checksum_header_name->len + colon_cursor.len +
+                checksum_result_cursor.len + post_trailer_cursor.len)) {
+        goto error;
+    }
```
TingDaoK, line 72: Feel like a bug.

So, the checksum trailer header will be added to the input stream here. However, the length of the input stream will not count that. So, the length of the stream is not the same as the REAL length.

I am not sure how it will affect other things. You probably want to disable the get length of the stream as well.

source/s3_checksum_stream.c
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/* We take ownership of the old inputstream, and destroy it with this input stream. This is because we want to be able
+ * to substitute in the chunk_stream for the cursor stream currently used in s_s3_meta_request_default_prepare_request
+ * which returns the new stream. So in order to prevent the need of keeping track of two input streams we instead
+ * consume the cursor stream and destroy it with this one */
+static void s_aws_input_checksum_stream_destroy(struct aws_input_stream *stream) {
+    if (stream) {
+        struct aws_checksum_stream *impl = stream->impl;
+        aws_checksum_finalize(impl->checksum, &impl->checksum_result, 0);
```
TingDaoK, line 71: this can fail right? Error handling?

source/s3_checksum_stream.c
```diff
@@ -0,0 +1,122 @@
...
+        struct aws_byte_cursor checksum_result_cursor = aws_byte_cursor_from_buf(&impl->checksum_result);
+        aws_base64_encode(&checksum_result_cursor, impl->encoded_checksum_output);
```
TingDaoK, line 73: this can fail right? Error handling? Unless it only fails from OOM, we will need error handling.

source/s3_chunk_stream.c
```diff
@@ -0,0 +1,271 @@
...
+    /* should I use aws_likely here? */
+    if (current_stream_length > 0) {
+        final_chunk_cursor = aws_byte_cursor_from_string(s_final_chunk);
+    } else {
+        final_chunk_cursor = aws_byte_cursor_from_string(s_empty_chunk);
+    }
+    struct aws_byte_cursor post_trailer_cursor = aws_byte_cursor_from_string(s_post_trailer);
+    struct aws_byte_cursor colon_cursor = aws_byte_cursor_from_string(s_colon);
+
+    struct aws_byte_cursor checksum_result_cursor = aws_byte_cursor_from_buf(&parent_stream->checksum_result);
```
TingDaoK, line 60: what if the `checksum_result` failed?

source/s3_request_messages.c
```diff
@@ -608,13 +670,48 @@ struct aws_input_stream *aws_s3_message_util_assign_body(
         goto error_clean_up;
     }
 
+    bool is_get = s_is_get_request(out_message);
```
TingDaoK, line 673: Will get request ever assign body? Should this just be an ASSERT?

include/aws/s3/s3_client.h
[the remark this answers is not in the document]
TingDaoK: that's related directly to the request, so, it makes sense to add it to the meta request level. I am not against to add both. So, the meta request level will override the client level config.

source/s3_request_messages.c
ilevyor: Yes, a default/single part get will be assigned a body.

source/s3_auto_ranged_put.c
```diff
@@ -72,32 +72,33 @@ struct aws_s3_meta_request *aws_s3_meta_request_auto_ranged_put_new(
...
-            &auto_ranged_put->synced_data.etag_list,
-            allocator,
-            s_etags_initial_capacity,
-            sizeof(struct aws_string *))) {
-        goto error_clean_up;
-    }
-
+    /* @dengket do the etags need to be synced? each etag is accessed exactly twice, once when initialized by the
+     * corresponding part, and once during complete multipart upload when the list is used to create the message body.
+     */
```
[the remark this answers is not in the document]
ilevyor: Super old comment, had since discussed with mike, no need to be synced.

source/s3_auto_ranged_put.c
```diff
@@ -72,32 +72,33 @@ struct aws_s3_meta_request *aws_s3_meta_request_auto_ranged_put_new(
...
+    struct aws_string **etag_c_array = aws_mem_calloc(allocator, sizeof(struct aws_string *), num_parts);
+    aws_array_list_init_static(
+        &auto_ranged_put->synced_data.etag_list, etag_c_array, num_parts, sizeof(struct aws_string *));
     auto_ranged_put->content_length = content_length;
     auto_ranged_put->synced_data.total_num_parts = num_parts;
     auto_ranged_put->threaded_update_data.next_part_number = 1;
 
+    auto_ranged_put->checksums_list = aws_mem_calloc(allocator, sizeof(struct aws_byte_buf), num_parts);
```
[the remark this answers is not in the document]
ilevyor: I think static is preferable as we lose guarantees on thread safety if the list can be resized.

source/s3_auto_ranged_put.c
```diff
@@ -119,7 +120,13 @@ static void s_s3_meta_request_auto_ranged_put_destroy(struct aws_s3_meta_request
         aws_array_list_get_at(&auto_ranged_put->synced_data.etag_list, &etag, etag_index);
         aws_string_destroy(etag);
     }
-
+    /* probably just better to store num parts in the meta-request */
```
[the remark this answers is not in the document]
ilevyor: num parts is stored in auto_ranged_put , it's part of synced data, though I think that's unnecessary as we can just const it.

source/s3_request_messages.c
TingDaoK: The get request should not have body. 
https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html Request Body section.

The default request does support the case for both body exist and not exist.

When there is no body, it should NOT call this function.

ilevyor: updated default requests to only assign if not a get request. Though, it is kind of odd to only check for get requests, and no other request types. so idk if that check really makes sense, a default head would still have it's body assigned. A copy object request (which is a Put with no body) will have a body assigned.

source/s3_chunk_stream.c
ilevyor: The stream needs a length, I believe it's used by http somewhere, I don't recall where exactly off the top of my head, but it's definitly used. Fortunatly this isn't a bug, as the length is calculated when creating the stream, and that value is what is returned by get length.
[https://github.com/awslabs/aws-c-s3/blob/0cb50d2b7ffb73188e97a221d75e4930067d85a7/source/s3_chunk_stream.c#L256-L259](url)

ilevyor:
```    /* we subtract one since aws_base64_compute_encoded_len accounts for the null terminator which won't show up in our
     * stream */
    impl->length = prechunk_stream_len + stream_length + final_chunk_len + impl->checksum_header_name->len + colon_len +
                   encoded_checksum_len + post_trailer_len - 1;```

Comment (ilevyor) [reply source/s3_chunk_stream.c:72]:
```
static int s_aws_input_chunk_stream_get_length(struct aws_input_stream *stream, int64_t *out_length) {
    struct aws_chunk_stream *impl = stream->impl;
    *out_length = impl->length;
    return AWS_OP_SUCCESS;
}
```

Comment (ilevyor) [reply source/s3_chunk_stream.c:72]: Ah, used for the content length, right after assigning the body, note that this is also a place using a http-chunked header vs an aws-chunked header could have gone wrong.
```
int64_t stream_length = 0;
    if (aws_input_stream_get_length(input_stream, &stream_length)) {
        goto error_clean_up;
    }
    char content_length_buffer[64] = "";
    snprintf(content_length_buffer, sizeof(content_length_buffer), "%" PRIu64, (uint64_t)stream_length);
```

Comment (TingDaoK) [reply source/s3_chunk_stream.c:72]: okay, you are right. You know the length before the chunk sum is computed.

File: source/s3_request_messages.c (hunk 25 above)
Comment (TingDaoK) [reply source/s3_request_messages.c:673]: I have not seen any update

Comment (ilevyor) [reply source/s3_request_messages.c]: Sorry didn't push, pushed now. Ran into errors moving checksum options from client settings to request settings. Testing infra really not built to support that.

File: source/s3_checksums.c [hunk 29]
```diff
@@ -0,0 +1,270 @@
...
+    }
+    if (available_buffer < len) {
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+    }
+    AWS_PRECONDITION(aws_byte_buf_is_valid(out));
+    uint32_t tmp = aws_hton32(*(uint32_t *)checksum->impl);
+    if (aws_byte_buf_write(out, (uint8_t *)&tmp, len)) {
+        return AWS_OP_SUCCESS;
+    }
+    return AWS_OP_ERR;
```
TingDaoK, line 119: you need raise error for this case, since `aws_byte_buf_write` will not raise error when it fails

source/s3_chunk_stream.c
```diff
@@ -57,6 +57,10 @@ static int s_set_post_chunk_stream(struct aws_chunk_stream *parent_stream) {
     struct aws_byte_cursor post_trailer_cursor = aws_byte_cursor_from_string(s_post_trailer);
     struct aws_byte_cursor colon_cursor = aws_byte_cursor_from_string(s_colon);
 
+    if (parent_stream->checksum_result.len == 0) {
+        AWS_LOGF_ERROR(AWS_LS_S3_META_REQUEST, "Failed to extract base64 encoded checksum of stream");
+        aws_raise_error(AWS_ERROR_S3_CHECKSUM_CALCULATION_FAILED);
```
TingDaoK, line 62: return

source/s3_checksum_stream.c
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/* We take ownership of the old inputstream, and destroy it with this input stream. This is because we want to be able
+ * to substitute in the chunk_stream for the cursor stream currently used in s_s3_meta_request_default_prepare_request
+ * which returns the new stream. So in order to prevent the need of keeping track of two input streams we instead
+ * consume the cursor stream and destroy it with this one */
+static void s_aws_input_checksum_stream_destroy(struct aws_input_stream *stream) {
+    if (stream) {
+        struct aws_checksum_stream *impl = stream->impl;
+        AWS_FATAL_ASSERT(aws_checksum_finalize(impl->checksum, &impl->checksum_result, 0) == AWS_OP_SUCCESS);
```
TingDaoK, line 71: Instead of FATAL ASSERT, maybe just ASSERT the return code and if error happens, set the `checksum_result` to empty. So that, the chunked_stream can check the empty checksum and return error properly from https://github.com/awslabs/aws-c-s3/pull/181/files#diff-4be41bb6016233fcddc834d9baff1a97b14989916f4323e3d7ca167e05f37a8aR60-R62

source/s3_checksum_stream.c
```diff
@@ -28,7 +28,7 @@ static int s_aws_input_checksum_stream_seek(
         "Cannot seek on checksum stream, as it will cause the checksum output to mismatch the checksum of the stream "
         "contents");
     AWS_ASSERT(false);
-    return AWS_OP_ERR;
+    return aws_raise_error(AWS_IO_STREAM_INVALID_SEEK_POSITION);
```
TingDaoK, line 31: use `AWS_ERROR_UNSUPPORTED_OPERATION` instead. That's more proper.

source/s3_chunk_stream.c
```diff
@@ -108,7 +108,7 @@ static int s_aws_input_chunk_stream_seek(
         "Cannot seek on chunk stream, as it will cause the checksum output to mismatch the checksum of the stream"
         "contents");
     AWS_ASSERT(false);
-    return AWS_OP_ERR;
+    return aws_raise_error(AWS_IO_STREAM_INVALID_SEEK_POSITION);
```
TingDaoK, line 115: use `AWS_ERROR_UNSUPPORTED_OPERATION` instead. That's more proper.

source/s3_chunk_stream.c
```diff
@@ -247,22 +254,27 @@ struct aws_input_stream *aws_chunk_stream_new(
...
+        impl->current_stream = impl->checksum_stream;
+        final_chunk_len = s_empty_chunk->len;
+        impl->checksum_stream = NULL;
+        impl->set_current_stream_fn = s_set_post_chunk_stream;
+    }
     AWS_FATAL_ASSERT(impl->current_stream);
     return stream;
 
 error:
     /* TODO: the stream should refcount, otherwise, this can lead to crash */
```
TingDaoK, line 277: trivial: `AWS_ASSERT(impl->checksum_stream == NULL);`

source/s3_meta_request.c
```diff
@@ -959,6 +962,9 @@ static int s_s3_meta_request_incoming_body(
         request->send_data.response_status,
         (uint64_t)data->len,
         (void *)connection);
+    if (request->send_data.response_status < 200 || request->send_data.response_status > 299) {
+        AWS_LOGF_TRACE(AWS_LS_S3_META_REQUEST, "response body: \n" PRInSTR "\n", AWS_BYTE_CURSOR_PRI(*data));
```
TingDaoK, line 966: log the status as well.

tests/s3_data_plane_tests.c
```diff
@@ -2390,6 +2503,9 @@ static int s_test_s3_round_trip_mpu_fc(struct aws_allocator *allocator, void *ct
     return 0;
 }
 
+// cancel upload
+// cancel download
+
```
TingDaoK, line 2508: remove this

include/aws/s3/s3_client.h
```diff
@@ -223,7 +223,7 @@ struct aws_s3_meta_request_options {
      * Note: If set, it will disable compute_content_md5
      */
     /* TODO add functionality for copy object */
-    enum aws_s3_checksum_algorithm checksum_algorithm;
+    enum aws_s3_checksum_algorithm put_checksum_algorithm;
```
TingDaoK, line 226: well, if this will affect copy object in the future, maybe not renaming this to `put_*`

TingDaoK (approved):
fix & ship

PR #182, opened 2022-02-24
PutObject pause/resume support

cmello:
Added support to pause and resume PutObject meta requests.

* In order to pause an ongoing upload, call `aws_s3_meta_request_pause()` . It will return a resume token that can be persisted and used to resume the upload.
* To resume an upload that was paused, supply the resume token in the meta request options structure member `aws_s3_meta_request_options.persistable_state` . The upload can be resumed either from the same client or a different one.

*Issue #, if available:*

*Description of changes:*


source/s3_auto_ranged_put.c
```diff
@@ -36,6 +36,43 @@ static void s_s3_auto_ranged_put_request_finished(
...
+    auto_ranged_put->synced_data.create_multipart_upload_sent = true;
+    auto_ranged_put->synced_data.create_multipart_upload_completed = true;
+    auto_ranged_put->upload_id = aws_string_new_from_string(allocator, persistable_state->multipart_upload_id);
+
+    struct aws_array_list *persisted_etag_list = &persistable_state->etag_list;
+
+    for (size_t etag_index = 0; etag_index < aws_array_list_length(persisted_etag_list); etag_index++) {
+        struct aws_string *etag = NULL;
+        aws_array_list_get_at(&auto_ranged_put->synced_data.etag_list, &etag, etag_index);
+        if (etag != NULL) {
```
TingDaoK, line 58: I don't think there is a case the etag is not NULL. The etag list is just created with empty instead.

TingDaoK:
Overall I can see two major bugs.
- The PutObject resume doesn't seem like to skip any uploaded part.
- Assume the input stream for the resume request will be the same as the input stream before. It doesn't seek or ignore the uploaded part (This is not a bug for now as the resume will start from the beginning anyway.)

include/aws/s3/private/s3_auto_ranged_put.h
```diff
@@ -39,8 +41,12 @@ struct aws_s3_auto_ranged_put {
 
     /* Members to only be used when the mutex in the base type is locked. */
     struct {
+        /* Array list of `struct aws_string *` */
         struct aws_array_list etag_list;
 
+        struct aws_s3_paginated_operation *list_parts_operation;
+        struct aws_string *continuation_token;
```
TingDaoK, line 48: renaming it to `list_parts_continuation_token`? It's for list parts right? It's confusing with the put object is also paused/resume(continues)

include/aws/s3/private/s3_auto_ranged_put.h
```diff
@@ -20,7 +22,7 @@ enum aws_s3_auto_ranged_put_request_tag {
 struct aws_s3_auto_ranged_put {
     struct aws_s3_meta_request base;
 
-    /* Useable after the Create Multipart Upload request succeeds. */
+    /* Initialized either during creation in resume flow or as result of create multi part upload during normal flow. */
```
TingDaoK, line 25: super trivial: ~multi part~ to `multipart`

source/s3_auto_ranged_put.c
```diff
@@ -86,8 +255,22 @@ struct aws_s3_meta_request *aws_s3_meta_request_auto_ranged_put_new(
...
     auto_ranged_put->synced_data.total_num_parts = num_parts;
+
     auto_ranged_put->threaded_update_data.next_part_number = 1;
 
+    if (options->resume_token) {
+        int load_persistable_state_error = s_load_persistable_state(
+            allocator, auto_ranged_put, content_length, multipart_upload_id, part_size, num_parts);
+        if (load_persistable_state_error) {
+            aws_raise_error(load_persistable_state_error);
+            goto error_clean_up;
```
TingDaoK, line 266: memory leak: `multipart_upload_id` needs to be destroyed.
Maybe clean it up from `error_clean_up`

source/s3_auto_ranged_put.c
```diff
@@ -169,6 +372,88 @@ static bool s_s3_auto_ranged_put_update(
...
+
+                /* Check if the etag/checksum list has the result already */
+                int part_index = auto_ranged_put->threaded_update_data.next_part_number - 1;
+                for (size_t etag_index = part_index;
+                     etag_index < aws_array_list_length(&auto_ranged_put->synced_data.etag_list);
+                     ++etag_index) {
+                    struct aws_string *etag = NULL;
+
+                    if (!aws_array_list_get_at(&auto_ranged_put->synced_data.etag_list, &etag, etag_index) && etag) {
+                        // part already downloaded, adjust stream and go to the next part
```
TingDaoK, line 386: super trivial: comment format inconsistent

TingDaoK: RESUME FROM HERE.

TingDaoK (commented):
WIP

source/s3_request_messages.c
```diff
@@ -833,21 +890,13 @@ struct aws_http_message *aws_s3_message_util_copy_http_message_no_body_filter_he
             }
         }
 
-        if (aws_http_message_add_header(message, header)) {
-            goto error_clean_up;
+        if (aws_http_message_add_header(dest_message, header)) {
+            return AWS_OP_ERR;
+            ;
```
TingDaoK, line 895: trivial: extra `;`

tests/s3_list_objects_tests.c
```diff
@@ -113,10 +113,13 @@ static void s_on_list_bucket_page_finished_fn(struct aws_s3_paginator *paginator
     struct list_bucket_test_data *test_data = user_data;
 
     test_data->error_code = error_code;
+    AWS_LOGF_INFO(AWS_LS_S3_GENERAL, "error code %d", error_code);
```
TingDaoK, line 116: remove the debug code

source/s3_paginator.c
```diff
@@ -0,0 +1,453 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/s3/private/s3_paginator.h>
```
TingDaoK, line 6: Are this revamping the list_object to be shared between different operation? How much new code here?

include/aws/s3/private/s3_auto_ranged_put.h
```diff
@@ -49,10 +55,13 @@ struct aws_s3_auto_ranged_put {
 
         struct aws_http_headers *needed_response_headers;
 
+        int list_parts_error_code;
```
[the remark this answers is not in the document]
DmitriyMusatkin: there was a typo in the part where this code should have been set. thats fixed now.
currently we only use complete_multipart_upload_error_code and 3 other ones are just set. I think we should keep it for now, if anything just for consistency

include/aws/s3/s3_client.h
```diff
@@ -337,6 +345,18 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
...
+ * In order to pause an ongoing upload, call aws_s3_meta_request_pause(). It will return a resume token that can be
+ * persisted and used to resume the upload. To resume an upload that was paused, supply the resume token in the meta
+ * request options structure member aws_s3_meta_request_options.persistable_state.
+ * The upload can be resumed either from the same client or a different one.
+ * @param meta_request pointer to the aws_s3_meta_request of the upload to be paused
+ * @param resume_token outputs the json string with the state that can be used to resume the operation.
+ * @return
+ */
+AWS_S3_API
+int aws_s3_meta_request_pause(struct aws_s3_meta_request *meta_request, struct aws_string **resume_token);
```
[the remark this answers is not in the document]
DmitriyMusatkin: addressed he the first 2 comments. 

Token is where it gets a bit complicated. If we want to support generic pause resume for all operations (get and copy) then the field encoded in the token need to be different. 
So this pr implements token as an opaque json string that callers should not parse. The format is currently just trivial multipart id, part size and number of parts. Don't think it's something that will be used across sdks and its more implementation detail of crt. Open to suggestions to make it better

source/s3_auto_ranged_put.c
```diff
@@ -36,6 +38,164 @@ static void s_s3_auto_ranged_put_request_finished(
     struct aws_s3_request *request,
     int error_code);
 
+static int s_s3_auto_ranged_put_pause(struct aws_s3_meta_request *meta_request, struct aws_string **resume_token);
+
+bool s_process_part_info(const struct aws_s3_part_info *info, void *user_data) {
+    struct aws_s3_auto_ranged_put *auto_ranged_put = user_data;
```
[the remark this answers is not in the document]
DmitriyMusatkin: dont think c warns about this. it an implicit cast from void. i've seen this pattern used in a bunch of other places in our codebase

source/s3_paginator.c
DmitriyMusatkin: yeah, this is refactoring of old list_object code to break paginator out of it and reusing paginator for both list objects and parts. A lot of code is reused and its mostly just rewiring old code

source/s3_auto_ranged_put.c
```diff
@@ -598,6 +573,106 @@ static bool s_s3_auto_ranged_put_update(
...
+    uint32_t skip_until_part_number) {
+
+    AWS_PRECONDITION(meta_request);
+
+    struct aws_s3_auto_ranged_put *auto_ranged_put = meta_request->impl;
+
+    AWS_PRECONDITION(skip_until_part_number <= auto_ranged_put->synced_data.total_num_parts);
+
+    if (num_parts_read_from_stream == skip_until_part_number) {
+        return AWS_ERROR_SUCCESS;
```
TingDaoK, line 620: trivial: use `AWS_OP_SUCCESS` instead. We usually use `AWS_OP_SUCCESS` and `AWS_OP_ERROR` as return code.

source/s3_auto_ranged_put.c
```diff
@@ -598,6 +573,106 @@ static bool s_s3_auto_ranged_put_update(
...
+            aws_byte_buf_init(&temp_body_buf, meta_request->allocator, request_body_size);
+        } else {
+            // reuse buffer
+            aws_byte_buf_reset(&temp_body_buf, false);
+        }
+
+        if (aws_s3_meta_request_read_body(meta_request, &temp_body_buf)) {
+            AWS_LOGF_ERROR(AWS_LS_S3_META_REQUEST, "Failed to resume upload. Input steam cannot be read.");
+            aws_byte_buf_clean_up(&temp_body_buf);
+            return AWS_ERROR_S3_SKIP_UPLOADED_PART_FAILED;
```
TingDaoK, line 649: Should we just use the error from `aws_s3_meta_request_read_body`? So, just return AWS_OP_ERROR

source/s3_auto_ranged_put.c
```diff
@@ -598,6 +573,106 @@ static bool s_s3_auto_ranged_put_update(
...
+            if (!aws_checksum_compute(
+                    meta_request->allocator, meta_request->checksum_algorithm, &body_cur, &checksum, 0) ||
+                !aws_byte_buf_eq(&checksum, &auto_ranged_put->checksums_list[part_index])) {
+                AWS_LOGF_ERROR(
+                    AWS_LS_S3_META_REQUEST,
+                    "Failed to resume upload. Checksum for previously uploaded part does not match "
+                    "new part.");
+                aws_byte_buf_clean_up(&temp_body_buf);
+                aws_byte_buf_clean_up(&checksum);
+                return AWS_ERROR_S3_RESUMED_PART_CHECKSUM_MISMATCH;
```
TingDaoK, line 669: well, if `aws_checksum_compute(meta_request->allocator, meta_request->checksum_algorithm, &body_cur, &checksum, 0)` failed, we will still raise the error of `AWS_ERROR_S3_RESUMED_PART_CHECKSUM_MISMATCH`.

source/s3_auto_ranged_put.c
```diff
@@ -652,24 +729,30 @@ static int s_s3_auto_ranged_put_prepare_request(
...
-                if (content_remainder > 0) {
-                    request_body_size = content_remainder;
+            if (request->num_times_prepared == 0) {
+                int skip_error = s_skip_parts_from_stream(
+                    meta_request, auto_ranged_put->prepare_data.num_parts_read_from_stream, request->part_number - 1);
+                if (skip_error) {
+                    /* BEGIN CRITICAL SECTION */
+                    {
+                        aws_s3_meta_request_lock_synced_data(meta_request);
+                        aws_raise_error(skip_error);
```
TingDaoK, line 743: just raise error within `s_skip_parts_from_stream`

source/s3_auto_ranged_put.c
```diff
@@ -1058,6 +1161,23 @@ static int s_s3_auto_ranged_put_pause(struct aws_s3_meta_request *meta_request,
...
+    /**
+     * generate pause token for put.
+     * current format is json with the following structure
+     * {
+     *      type: String
+     *      multipart_upload_id: String
+     *      partition_size: number
+     *      total_num_parts: number
+     * }
+     */
```
TingDaoK, line 1175: Why not add this to public header file?

source/s3_auto_ranged_put.c
```diff
@@ -122,6 +151,11 @@ static int s_parse_resume_info_from_token(
     aws_json_module_cleanup();
 
     return AWS_ERROR_SUCCESS;
```
TingDaoK, line 153: Trivial: Use AWS_OP_SUCCESS instead

source/s3_auto_ranged_put.c
```diff
@@ -122,6 +151,11 @@ static int s_parse_resume_info_from_token(
...
+
+invalid_argument_cleanup:
+    aws_json_value_destroy(root);
+    aws_json_module_cleanup();
+    return AWS_ERROR_INVALID_ARGUMENT;
```
TingDaoK, line 158: just `return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);`

We usually just return AWS_OP_ERROR/AWS_OP_SUCCESS, the detailed error code will be raised within the function.

source/s3_auto_ranged_put.c
```diff
@@ -598,6 +573,106 @@ static bool s_s3_auto_ranged_put_update(
...
+ * and will read into temp buffer until it gets to skip_until_part_number (i.e. skipping does include
+ * that part). If checksum is set on the request and parts with checksums were uploaded before, checksum will be
+ * verified.
+ */
+static int s_skip_parts_from_stream(
+    struct aws_s3_meta_request *meta_request,
+    uint32_t num_parts_read_from_stream,
+    uint32_t skip_until_part_number) {
+
+    AWS_PRECONDITION(meta_request);
```
TingDaoK, line 613: Maybe add another `AWS_PRECONDITION(num_parts_read_from_stream<=skip_until_part_number);`?

source/s3_auto_ranged_put.c
DmitriyMusatkin: my reasoning for not adding it to public headers would be to avoid giving callers any impression that its ok to parse tokens

TingDaoK: hmm, iirc, one of the reason about we adding the complexity of listing parts is:
- Consistent with v1 implementations and mental model that customers might already have.

So, if we don't expose the format, the Java SDK cannot adapt it, right?

DmitriyMusatkin: what do you mean by adopting here? To java sdk its a transparent token, they should not be parsing. on java side, they just wrap the token in additional data needed to serialize the request and it operates the same way as v1.

source/s3_auto_ranged_put.c
```diff
@@ -36,6 +38,196 @@ static void s_s3_auto_ranged_put_request_finished(
...
+    }
+
+    if (total_num_parts != num_parts) {
+        AWS_LOGF_ERROR(
+            AWS_LS_S3_META_REQUEST,
+            "Could not create auto-ranged-put meta request; persisted number of parts %" PRIu32
+            " does not match expected number of parts based on length of the body.",
+            total_num_parts);
+
+        return AWS_ERROR_INVALID_ARGUMENT;
```
TingDaoK, line 185: `return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);`

source/s3_auto_ranged_put.c
```diff
@@ -36,6 +38,196 @@ static void s_s3_auto_ranged_put_request_finished(
...
+    auto_ranged_put->synced_data.num_parts_completed = 0;
+    auto_ranged_put->synced_data.create_multipart_upload_sent = true;
+    auto_ranged_put->synced_data.create_multipart_upload_completed = true;
+    auto_ranged_put->upload_id = upload_id;
+
+    struct aws_byte_cursor request_path;
+    if (aws_http_message_get_request_path(auto_ranged_put->base.initial_request_message, &request_path)) {
+        AWS_LOGF_ERROR(AWS_LS_S3_META_REQUEST, "Could not load persisted state. Request path could not be read.");
+        return AWS_ERROR_INVALID_ARGUMENT;
+    }
```
TingDaoK, line 198: trivial: can we move this a bit to line 187? Feel more natural if all those argument check stay together. Also, `return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);`

source/s3_auto_ranged_put.c
```diff
@@ -36,6 +38,196 @@ static void s_s3_auto_ranged_put_request_finished(
...
+        AWS_ZERO_STRUCT(header_value);
+
+        if (!aws_http_headers_get(initial_headers, *header_name, &header_value)) {
+            aws_http_headers_set(needed_response_headers, *header_name, header_value);
+        }
+    }
+
+    auto_ranged_put->synced_data.needed_response_headers = needed_response_headers;
+
+    return AWS_ERROR_SUCCESS;
```
TingDaoK, line 228: `return AWS_OP_SUCCESS;`

source/s3_auto_ranged_put.c
```diff
@@ -81,13 +282,29 @@ struct aws_s3_meta_request *aws_s3_meta_request_auto_ranged_put_new(
...
     auto_ranged_put->threaded_update_data.next_part_number = 1;
+    auto_ranged_put->prepare_data.num_parts_read_from_stream = 0;
+
+    if (options->resume_token) {
+        int load_persistable_state_error = s_load_persistable_state(
+            allocator, auto_ranged_put, content_length, multipart_upload_id, part_size, num_parts);
+        if (load_persistable_state_error) {
+            aws_raise_error(load_persistable_state_error);
+            goto error_clean_up;
+        }
```
TingDaoK, line 298: if we raise error within the function, we can do:
```
        if (s_load_persistable_state(
                allocator, auto_ranged_put, content_length, multipart_upload_id, part_size, num_parts)) {
            goto error_clean_up;
        }
```

source/s3_auto_ranged_put.c
```diff
@@ -308,6 +575,116 @@ static bool s_s3_auto_ranged_put_update(
...
+            // reuse buffer
+            aws_byte_buf_reset(&temp_body_buf, false);
+        }
+
+        int read_error = aws_s3_meta_request_read_body(meta_request, &temp_body_buf);
+        if (read_error) {
+            AWS_LOGF_ERROR(AWS_LS_S3_META_REQUEST, "Failed to resume upload. Input steam cannot be read.");
+            aws_byte_buf_clean_up(&temp_body_buf);
+            return aws_raise_error(read_error);
+        }
```
TingDaoK, line 652: Basically all of our code base will not return the real error code back. The return code will only be `AWS_OP_SUCCESS/AWS_OP_ERR`. And if `AWS_OP_ERR` returned, the error code should have been raised. So: 
```
if (aws_s3_meta_request_read_body(meta_request, &temp_body_buf)) {
     AWS_LOGF_ERROR(AWS_LS_S3_META_REQUEST, "Failed to resume upload. Input steam cannot be read.");
     aws_byte_buf_clean_up(&temp_body_buf);
     return AWS_OP_ERR;
}
```

source/s3_auto_ranged_put.c
```diff
@@ -308,6 +575,116 @@ static bool s_s3_auto_ranged_put_update(
...
+                    "new part.");
+                aws_byte_buf_clean_up(&temp_body_buf);
+                aws_byte_buf_clean_up(&checksum);
+                return aws_raise_error(AWS_ERROR_S3_RESUMED_PART_CHECKSUM_MISMATCH);
+            }
+        }
+    }
+
+    aws_byte_buf_clean_up(&temp_body_buf);
+    return AWS_ERROR_SUCCESS;
```
TingDaoK, line 685: `return AWS_OP_SUCCESS;`

source/s3_auto_ranged_put.c
```diff
@@ -320,6 +697,40 @@ static int s_s3_auto_ranged_put_prepare_request(
...
+                        aws_byte_cursor_from_string(auto_ranged_put->synced_data.list_parts_continuation_token);
+                    message_creation_result = aws_s3_construct_next_request_http_message(
+                        auto_ranged_put->synced_data.list_parts_operation, &continuation_cur, &message);
+                } else {
+                    message_creation_result = aws_s3_construct_next_request_http_message(
+                        auto_ranged_put->synced_data.list_parts_operation, NULL, &message);
+                }
+
+                if (message_creation_result) {
+                    goto message_create_failed;
```
TingDaoK, line 718: add `aws_s3_meta_request_unlock_synced_data(meta_request);` before exist. Otherwise, it could lead to a deadlock.

source/s3_auto_ranged_put.c
```diff
@@ -364,6 +781,26 @@ static int s_s3_auto_ranged_put_prepare_request(
...
+                 * Read it from input stream and potentially verify checksum */
+                int skip_error = s_skip_parts_from_stream(
+                    meta_request,
+                    auto_ranged_put->prepare_data.num_parts_read_from_stream,
+                    auto_ranged_put->synced_data.total_num_parts);
+                if (skip_error) {
+                    /* BEGIN CRITICAL SECTION */
+                    {
+                        aws_s3_meta_request_lock_synced_data(meta_request);
+                        aws_raise_error(skip_error);
```
TingDaoK, line 795: same

source/s3_auto_ranged_put.c
```diff
@@ -330,24 +741,30 @@ static int s_s3_auto_ranged_put_prepare_request(
...
+                if (s_skip_parts_from_stream(
+                        meta_request,
+                        auto_ranged_put->prepare_data.num_parts_read_from_stream,
+                        request->part_number - 1)) {
+                    /* BEGIN CRITICAL SECTION */
+                    {
+                        aws_s3_meta_request_lock_synced_data(meta_request);
+                        aws_s3_meta_request_set_fail_synced(meta_request, NULL, AWS_ERROR_S3_RESUME_FAILED);
+                        aws_s3_meta_request_unlock_synced_data(meta_request);
+                    }
```
TingDaoK, line 756: seems like from `s_s3_meta_request_prepare_request_task` function, as long as you raised the error code correctly and return NULL; it will call `aws_s3_meta_request_set_fail_synced`.  [here](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_meta_request.c#L558-L578).

Why not follow the previous way of error handling?

source/s3_auto_ranged_put.c
```diff
@@ -458,6 +895,61 @@ static void s_s3_auto_ranged_put_request_finished(
...
+            {
+                aws_s3_meta_request_lock_synced_data(meta_request);
+
+                bool has_more_results = false;
+
+                if (error_code == AWS_ERROR_SUCCESS) {
+
+                    struct aws_byte_cursor body_cursor = aws_byte_cursor_from_buf(&request->send_data.response_body);
+                    aws_string_destroy(auto_ranged_put->synced_data.list_parts_continuation_token);
+                    if (aws_s3_paginated_operation_on_response(
```
TingDaoK, line 909: `aws_s3_paginated_operation_on_response` This function doesn't seem like it have a failure path.

include/aws/s3/private/s3_auto_ranged_put.h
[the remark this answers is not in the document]
DmitriyMusatkin: I added it mostly for consistency with how the other calls are done. There are already 3 existing error code defined here for create/complete/abort. out of those 3, only 1 error code is used. 
I do agree thats its a bit of a code smell having 3/4 being unused. We should either stay consistent with previous approach and have error variable for every operation or remove all of the unused ones.

source/s3_auto_ranged_put.c
DmitriyMusatkin: This is mostly for consistency with the rest of interface returning error status. I dont have strong preference here. can be made void as well

TingDaoK: If it cannot fail, just not return error status, it will be confusing to user that apply an error handling but actually handles nothing.

DmitriyMusatkin: it can actually fail. updated to throw error on parse errors

TingDaoK (approved):
looks good. thank you

PR #183, opened 2022-03-14
Proposal to fix reuse endpoint with refcount can reach zero

TingDaoK:
*Issue #, if available:*

- There was a race condition when the endpoint ref drops to zero, the other thread can bump it up to one again.
- There was a work around to the race condition is to allow this from happening and handle it by the caller to check the refcount when clean up with lock and decide to not really clean up the resource if the refcount bumps up to one.

*Description of changes:*

- I believe the workaround described above is not a good solution
- The caller should be responsible to not bump up the ref once it drops to zero
- Add the helper for caller to hold lock to clean the resource before drop the refcount if needed.


graebm (approved):
Looks good. Trivial feedback:
Can you write more in the PR description?

I've been doing a lot of Git archeology and it helps to know exactly why changes were made

PR #185, opened 2022-03-28
init copy default signing config

ilevyor:
*Issue #, if available:*

*Description of changes:*
Add init_copy_default_signing_config function


TingDaoK (commented):
how you want to use it? If you want to get the config from the client, it should be from `aws_cached_signing_config_aws`, right? This just makes no sense, you can just use `=` ti copy each member.

source/s3_meta_request.c
```diff
@@ -610,6 +610,20 @@ static void s_s3_meta_request_sign_request(
     meta_request->vtable->sign_request(meta_request, request, on_signing_complete, user_data);
 }
 
+static bool s_is_put_request(const struct aws_s3_meta_request *meta_request) {
+    if (meta_request->type == AWS_S3_META_REQUEST_TYPE_PUT_OBJECT) {
+        return true;
+    }
+    if (meta_request->type == AWS_S3_META_REQUEST_TYPE_DEFAULT) {
+        struct aws_byte_cursor method;
+        int err = aws_http_message_get_request_method(meta_request->initial_request_message, &method);
```
bretambrose, line 619: Feels like you could simplify this to something like:

```
if (!aws_http_message_get_request_method(...)) {
   return aws_byte_cursor_eq(...);
}
```
and let the fall through the hit the return false;

source/s3_meta_request.c
```diff
@@ -663,6 +677,10 @@ void aws_s3_meta_request_sign_request_default(
         return;
     }
 
+    if (meta_request->checksum_algorithm != AWS_SCA_NONE && s_is_put_request(meta_request) &&
```
DmitriyMusatkin, line 680: nit: one condition per line would be more readable.
also comment on why it needs to be done would be helpful

bretambrose (approved):

TingDaoK (approved):

source/s3_meta_request.c
ilevyor: formatter doesn't like separating conditions here.

PR #189, opened 2022-05-02
Don't re-send "x-amz-meta-" headers in each part of a multipart upload.

graebm:
**Issue:**
Multipart uploads with [user defined metadata](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata) ("x-amz-meta-" headers) would fail due to these headers being repeated in each individual part that was uploaded. 

**Description of Changes:**
Filter "x-amz-meta-" headers out of any multipart upload request that isn't the initial "start" request.

### DIFF
--- a/source/s3_request_messages.c
+++ b/source/s3_request_messages.c
@@ -121,6 +121,8 @@ const struct aws_byte_cursor g_s3_abort_multipart_upload_excluded_headers[] = {
     AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-copy-source-range"),
 };
 
+static const struct aws_byte_cursor s_x_amz_meta_prefix = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-meta-");
+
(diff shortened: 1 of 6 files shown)


source/s3_request_messages.c
```diff
@@ -169,11 +172,12 @@ struct aws_http_message *aws_s3_create_multipart_upload_message_new(
     AWS_PRECONDITION(allocator);
 
     /* For multipart upload, sse related headers should only be shown in create-multipart request */
-    struct aws_http_message *message = aws_s3_message_util_copy_http_message_no_body(
+    struct aws_http_message *message = aws_s3_message_util_copy_http_message_no_body_filter_headers(
```
TingDaoK, line 175: trivial: can we add a comment here, something like "Only include the x-amz-meta-* when the object/upload created"? It was clear before, as we have different list of headers to exclude for different type of request, but when we include `x-amz-meta-*` is quite unclear

PR #194, opened 2022-06-21
remove unused error code

TingDaoK:
Remove the error code that not used anywhere now.


bretambrose (commented):
Some of the managed languages include the integer error code in their notion of error.  This will potentially change things  for anyone checking the code and not the string (which is a bad thing to check anyways).  What makes this necessary?

TingDaoK:
> Some of the managed languages include the integer error code in their notion of error. This will potentially change things for anyone checking the code and not the string (which is a bad thing to check anyways). What makes this necessary?

thank you for pointing it out! Nothing really necessary here, just got notified by people that those error code are unused. Will close this PR

PR #197, opened 2022-07-20
Pause json cleanup

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
remove init of json lib from ranged put. json lib relies on global state and should only be initialized one.
add test for resuming uploads with all parts completed, but complete mpu is not called (since pause waits on all parts in flight to finish, this can occur often when paused towards the end of transfer)


TingDaoK (approved):
The fix looks good, but I think if the `aws_json_module_init` is meant to be called once with `aws_common_library_init`, should it be private to aws-c-common? Can we also make that change to avoid it from happening again?

TingDaoK:
I created a PR for it. https://github.com/awslabs/aws-c-common/pull/920

PR #198, opened 2022-08-11
Flow Control API proposal

graebm:
Note that this is going to a WIP branch.

I hope to build this out incrementally before merging to `main`

This matches the flow control API we're using elsewhere (see [here](https://github.com/awslabs/aws-c-http/blob/eb9a90aa73fe1caa7d24e3a8a3069c97bdae3f8b/include/aws/http/connection.h#L307-L326) and [here](https://github.com/awslabs/aws-c-http/blob/eb9a90aa73fe1caa7d24e3a8a3069c97bdae3f8b/include/aws/http/request_response.h#L985-L999))


include/aws/s3/s3_client.h
```diff
@@ -200,6 +200,27 @@ struct aws_s3_client_config {
...
     aws_s3_client_shutdown_complete_callback_fn *shutdown_callback;
     void *shutdown_callback_user_data;
+
+    /**
+     * Set to true to manually manage the flow-control window of each meta request.
+     *
+     * If false, no back-pressure is applied and data will download as fast as possible.
+     *
+     * If true, the flow-control window of each meta request will shrink as
+     * response body data is downloaded (headers do not affect the window).
```
TingDaoK, line 210: Can we clarify the `response body data` a bit more? 

It's set for the whole client. The response for meta requests besides GET will be consider as part of the response body? (List part, list object and etc.)

graebm: Let's chat. I'm still wrapping my head around everything aws-c-s3 does under the hood.

My intention is: from the user's perspective, they are only adjusting window for the bytes they receive via the `body_callback`.

If aws-c-s3 is doing other requests under the hood, where the user doesn't directly see the response, then aws-c-s3 should be automagically managing the window for those responses

graebm: Ah, I see you're talking about the special [list object API](https://github.com/awslabs/aws-c-s3/blob/b93780769953f661bcae294cd26fef76562aabe8/include/aws/s3/private/s3_list_objects.h). Our code would handle flow-control here (either disable it, or let data flow as fast as possible).

TingDaoK (approved):

include/aws/s3/s3_client.h
```diff
@@ -200,6 +213,27 @@ struct aws_s3_client_config {
...
+     * The user must call aws_s3_meta_request_update_window() to increment the window and keep data flowing.
+     */
+    bool manual_window_management;
+
+    /**
+     * The starting size of each meta-request's flow-control window.
+     * Required if `manual_window_management` is true,
+     * ignored if `manual_window_management` is false.
+     */
+    size_t initial_window_size;
```
DmitriyMusatkin, line 236: what are the units for this? 
requests?

PR #203, opened 2022-08-20
Fix race condition in aws_s3_client_endpoint_release (#202)

grrtrr:
This fixes the race condition in `aws_s3_client_endpoint_release` reported in #202.

### DIFF
--- a/include/aws/s3/private/s3_client_impl.h
+++ b/include/aws/s3/private/s3_client_impl.h
@@ -32,7 +32,7 @@ enum aws_s3_connection_finish_code {
 };
 
 /* Callback for the owner of the endpoint when the endpoint has completely cleaned up. */
-typedef void(aws_s3_endpoint_shutdown_fn)(void *user_data);
+typedef void(aws_s3_endpoint_shutdown_fn)(struct aws_string *host_name, void *user_data);
 
 struct aws_s3_endpoint_options {
     /* URL of the host that this endpoint refers to. */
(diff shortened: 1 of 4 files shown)


TingDaoK (commented):
Thank you for the detailed debug process. The idea of the fix is right.

However, the fix could lead to a dead lock. The `aws_ref_count_release(&endpoint->ref_count);` will invoke the `s_s3_endpoint_ref_count_zero` within the lock. And lead to the shutdown callback of endpoint invoked from `s_s3_endpoint_http_connection_manager_shutdown_callback`. 

The shutdown callback is defualt to [here](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L733), which will grab the lock from the client again. And lead to a deadlock.

We need to fix this in another way.

TingDaoK:
Maybe we can move the removal from the hashtable to the shutdown callback.

If the client is the user of the endpoint, the shutdown callback from the client should be the one who take care of the shutdown process, which includes remove it from the hashtable. In this way, we can get rid of the two weird release function and things to check if it's owned by the client or not.

We will fix this later (It's aware by us, but we have other priorities to fix). If you are willing to fix this, we are happy to review.

grrtrr:
> If you are willing to fix this, we are happy to review.

Thank you for help with this. Based on the comments, I will work on a better solution and update #203.

grrtrr:
> the shutdown callback from the client should be the one who take care of the shutdown
This requires some reorganization. Simply adding `aws_hash_table_remove` in `s_s3_client_endpoint_shutdown_callback` will not work, since this allows an S3Endpoint reference with a `NULL` `http_connection_manager` to persist for some time (leading to the same segmentation faults as before).

grrtrr:
Added the failing implementation to document the problem - hash table entry and `http_connection_manager` need to be removed at the same time, or remove the hash table entry first.
The latter again requires a complicated solution of adding 1 more callback, and ensuring that this callback is always called without race condition - which is the starting point of #202.
Hence I think the best solution is to remove keeping entries in the hash table. The performance gains do not seem to justify the greatly increased complexity and with it, the potential for race conditions and segmentation faults.

PR #204, opened 2022-08-22
Expose additional configuration options for Http connection manager

waahm7:
*Issue #, if available:* https://github.com/awslabs/aws-crt-java/issues/502 & https://github.com/aws/aws-sdk-cpp/issues/1882

*Description of changes:*
- Exposes proxy options 
- Exposes KeepAlive options
- Exposes connect_timeout_ms
- Exposes monitoring options
- Exposes proxy_env_var_settings


include/aws/s3/private/s3_client_impl.h
```diff
@@ -183,6 +191,14 @@ struct aws_s3_client {
     aws_s3_client_shutdown_complete_callback_fn *shutdown_callback;
     void *shutdown_callback_user_data;
 
+    /*
+     * Optional.
+     * Configuration for using proxy from environment variable.
```
TingDaoK, line 196: `from environment variable.`?

Is this a copy/paste error between `proxy_env_var_settings`?

include/aws/s3/s3_client.h
```diff
@@ -200,6 +200,20 @@ struct aws_s3_client_config {
...
     aws_s3_client_shutdown_complete_callback_fn *shutdown_callback;
     void *shutdown_callback_user_data;
+
+    /*
+     * Optional.
+     * Configuration for using proxy from environment variable.
+     * Defaults to true
+     * Only works when proxy_options is not set.
+     */
+    struct aws_http_proxy_options *proxy_options;
```
TingDaoK, line 210: The meta request can be connected to a separate endpoint, maybe it makes sense to add the override of the proxy options from the meta request?

include/aws/s3/private/s3_client_impl.h
waahm7: Yes, fixed. Thanks.

source/s3_client.c
```diff
@@ -684,6 +683,7 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
                 .user_data = client,
                 .max_connections = aws_s3_client_get_max_active_connections(client, NULL),
                 .port = port,
+                .proxy_options = options->proxy_options,
```
waahm7, line 686: We don't need to make a deep copy of options as when the endpoint creates a connection manager, it will deep copy proxy_options to proxy_config. So we can simply pass it down to the connection manager.

include/aws/s3/s3_client.h
```diff
@@ -290,6 +306,32 @@ struct aws_s3_meta_request_options {
...
+     * Optional.
+     * If set to 0, default value is used.
+     */
+    uint32_t connect_timeout_ms;
+
+    /*
+     * Optional.
+     * Set keepalive true to periodically transmit messages for detecting a disconnected peer.
+     */
+    struct aws_s3_tcp_keep_alive_options tcp_keep_alive_options;
```
graebm, line 334: it's weird that we're adding these new options to `aws_s3_meta_request_options`

These options are used when creating an `aws_s3_endpoint` (which contains an `aws_http_connection_manager`). But then we store the `aws_s3_endpoint` in a hashtable and reuse them for later meta-requests. So if there are 2 meta-requests using the same endpoint, then the 2nd meta-request is going to use the `aws_s3_endpoint` that was created by the 1st meta-request. If the 1st meta-request didn't set any of these new options, but the 2nd meta-request did, then these new options will end up getting totally ignored.

If we keep these new options in `aws_s3_meta_request_options`, we need to be sure they're honored. Like, don't just look up pre-existing `aws_s3_endpoint` by URI, instead look up endpoints based on ALL these settings.

OR we go with another mechanism for configuring specific endpoints. Like, when creating the client, the user can pass in per-endpoint settings. Later on, when a meta-request is made, and we realize we need to create an `aws_s3_endpoint`, we have the settings to use for it

I don't know what level of configuration people are going to want though. Like, do they want to specify defaults? Like use these proxy options for ALL URI by default?  Or is there any chance they'd only want to use the extra settings for 1 request to a URI, but not other requests to that same URI?

waahm7: In the original two issues, customers are expecting to set these properties at the client level. S3 client in CPP (not the CRT-S3 client) also allowed customers to set these properties at a client level instead of an endpoint. I am not sure about how the other SDKs are handling it.
 
**Potential Solution 1 (Expose these options in the Client Config):**
I think we can also do something similar to S3 CPP client and move these properties to S3ClientConfig, which will resolve the issue of caching meta requests.

Issues:
- Customers will have to create different clients for different options.

**Potential Solution 2 (Expose these options per request basis):**
Current solution, customers will be able to configure different options for each endpoint. 

Issues:
- Caching (will need to discuss and decide on a solution)
- Customers will have to provide these properties for each meta request which might not be ideal if they don't change that frequently on per endpoint basis.

I am not sure either how frequently the customers will want to configure these properties. I am more inclined towards solution 1 as it is simple and matches the CPP S3 Client, however, this mainly depends on how frequently these options change.

graebm: I am also inclined towards option 1 over option 2. It seems much more natural to configure the client, instead of each individual request.

We could ALSO allow per-endpoint configuration, by letting the user submit some kind of per-endpoint-configuration list. Then if a request uses that particular endpoint, it will use that configuration.

include/aws/s3/s3_client.h
```diff
@@ -202,6 +202,22 @@ struct aws_s3_client_config {
     void *shutdown_callback_user_data;
 };
 
+/* Keepalive properties are TCP only.
+ * If interval or timeout are zero, then default values are used.
+ */
+struct aws_s3_tcp_keep_alive_options {
+
+    /* Set keepalive true to periodically transmit messages for detecting a disconnected peer.*/
+    bool keepalive;
```
DmitriyMusatkin, line 211: do we need bool here? wouldnt null keep alive indicate keep alive is on?

include/aws/s3/private/s3_client_impl.h
```diff
@@ -58,6 +58,32 @@ struct aws_s3_endpoint_options {
...
+     * Defaults to true
+     * Only works when proxy_options is not set.
+     */
+    struct proxy_env_var_settings *proxy_ev_settings;
+
+    /**
+     * Optional.
+     * If set to 0, default value is used.
+     */
+    uint32_t connect_timeout_ms;
```
DmitriyMusatkin, line 80: nit: from user perspective i hate seeing default value is used. can we mention what that default value is

include/aws/s3/private/s3_client_impl.h
```diff
@@ -58,6 +58,32 @@ struct aws_s3_endpoint_options {
...
+    /**
+     * Optional.
+     * Proxy configuration for http connection.
+     */
+    struct aws_http_proxy_options *proxy_options;
+
+    /*
+     * Optional.
+     * Configuration for using proxy from environment variable.
+     * Defaults to true
```
DmitriyMusatkin, line 71: what does defaults to true on pointer means?

include/aws/s3/private/s3_client_impl.h
```diff
@@ -58,6 +58,32 @@ struct aws_s3_endpoint_options {
 
     /* HTTP port override. If zero, determine port based on TLS context */
     uint16_t port;
+
+    /**
+     * Optional.
+     * Proxy configuration for http connection.
```
DmitriyMusatkin, line 64: can we add a more explicit comment on relationship between proxy options and ev proxy options and what happens when users set both

source/s3_client.c
DmitriyMusatkin: this sounds like potential for race condition bugs. 
creation is async, so if it returns before endpoints are created and user deletes options, well run into random bugs

include/aws/s3/private/s3_client_impl.h
waahm7: Thanks, fixed.

source/s3_client.c
waahm7: Fixed by copying proxy options to proxy config.

include/aws/s3/s3_client.h
waahm7: Do you mean null keep alive will indicate that keep alive is off? Since it is just a normal struct, it can't be null. We can turn it into a pointer and remove the keepalive. If the pointer is null, that would mean keepalive is off and vice versa. 

Otherwise, we need a keepalive variable so that users can just set keepalive to true and use default values for the rest. This is basically a partial struct from aws_socket_options which has two extra fields  enum aws_socket_type type and enum aws_socket_domain domain. I didn't want to allow users to override these two fields.

source/s3_client.c
```diff
@@ -304,6 +304,28 @@ struct aws_s3_client *aws_s3_client_new(
...
+        client->proxy_ev_settings = aws_mem_calloc(allocator, 1, sizeof(struct proxy_env_var_settings));
+        *client->proxy_ev_settings = *client_config->proxy_ev_settings;
+
+        if (client_config->proxy_ev_settings->tls_options) {
+            client->proxy_ev_tls_options = aws_mem_calloc(allocator, 1, sizeof(struct aws_tls_connection_options));
+            if (aws_tls_connection_options_copy(client->proxy_ev_tls_options, client->proxy_ev_settings->tls_options)) {
+                goto on_error;
+            }
+            client->proxy_ev_settings->tls_options = client->proxy_ev_tls_options;
+        }
```
waahm7, line 324: Is there a better way to deep copy proxy_ev_settings->tls_options? The problem is that in proxy_ev_settings, tls_options is declared as const. So if I just use that directly instead of using client->proxy_ev_tls_options, I get a bunch of warnings about passing const tls_options to other functions which expect a non-const object like aws_mem_release.

source/s3_client.c
```diff
@@ -421,6 +449,17 @@ struct aws_s3_client *aws_s3_client_new(
         aws_mem_release(client->allocator, client->tls_connection_options);
         client->tls_connection_options = NULL;
     }
+    if (client->proxy_config) {
+        aws_http_proxy_config_destroy(client->proxy_config);
+    }
```
waahm7, line 454: I am unable to write a test for the proxy_config override. As the struct proxy_config implementation detail is private in the aws-c-http repo, I can't access its members to assert.

source/s3_client.c
```diff
@@ -304,6 +304,34 @@ struct aws_s3_client *aws_s3_client_new(
         *((size_t *)&client_config->max_part_size) = client_config->part_size;
     }
 
+    if (client_config->proxy_options) {
+        client->proxy_config = aws_http_proxy_config_new_from_proxy_options(allocator, client_config->proxy_options);
+        if (client->proxy_config == NULL) {
+            goto on_error;
+        }
+    }
```
waahm7, line 312: I am unable to write a test for the proxy_config override. As the struct proxy_config implementation detail is private in the aws-c-http repo, I can't access its members to assert.

include/aws/s3/private/s3_client_impl.h
graebm: OBJECTION❗️

when we write the exact default value, it inevitably gets tweaked to fix someone's ticket and we forget to fix the N places we copy/pasted that value to docs

I specifically put this in our developer guide ;)

https://github.com/awslabs/aws-crt-python/blob/main/guides/dev/writing-code.md#use-none-as-the-default-value-for-optional-arguments
> In documentation, just say "a default value is used" instead of writing in the actual value, because the odds are good that the documentation will get out of sync with reality.

include/aws/s3/private/s3_client_impl.h
```diff
@@ -15,7 +15,9 @@
 #include <aws/common/mutex.h>
 #include <aws/common/ref_count.h>
 #include <aws/common/task_scheduler.h>
+#include <aws/http/connection.h>
 #include <aws/http/connection_manager.h>
+#include <aws/http/proxy.h>
```
graebm, line 20: Prefer to use forward-declarations instead of adding includes, whenever you can get away with it. This keeps down compile times in large C/C++ projects.

include/aws/s3/s3_client.h
```diff
@@ -150,6 +149,22 @@ enum aws_s3_checksum_algorithm {
     AWS_SCA_COUNT,
 };
 
+/* Keepalive properties are TCP only.
+ * If interval or timeout are zero, then default values are used.
+ */
+struct aws_s3_tcp_keep_alive_options {
```
graebm, line 155: why break these out into their own struct?

But I COULD see an argument for a new struct for ALL connection-specific options, like 
`struct aws_s3_client_connection_options`. The argument being: if we want to allow endpoint-specific options in the future, now we can just re-use this struct that we were already using.

I'm wishy washy on whether this is a good idea. We'd have to move 2 existing variables into this new struct: `aws_s3_client_config.tls_mode` and `aws_s3_client_config.tls_connection_options`

include/aws/s3/private/s3_client_impl.h
waahm7: I have removed the exact default value.

DmitriyMusatkin: I agree that it can be a pain to keep comments up to date, but I've also been on the other side of this, where i had to read through a bunch of code to find out what the default value is. Its been especially frustrating with java V1 sdks where you have multiple indirections before you can find the value. Java V2 is improving on this and a lot of classes mention default values. I love that pretty much every CDK options field mentions default value in the comment.

Anyways, this is not a hill im willing to die on right now. If we decide not to put defaults in comments as a team, im happy to disagree and commit.

include/aws/s3/private/s3_client_impl.h
waahm7: I have tried that. The problem with that is that we can't write tests for these variables that verify that allocation, deallocation, and assignment are working fine. As it is a private header, it won't be included in dependencies anyhow.

include/aws/s3/s3_client.h
waahm7: I have updated it to a pointer, and null means keep alive is false. I broke these out into their own struct because of their relationship to keep alive.

include/aws/s3/private/s3_client_impl.h
waahm7: Both options seem good to me. One other possible option might be that we could maintain an appendix of all the default values. It will be easier to update it at one location only.

DmitriyMusatkin (approved):

PR #208, opened 2022-09-02
Endpoint lifetime fix

TingDaoK:
*Issue #, if available:*

- The release of the endpoint will lead to touch the hashtable owned by client, which needs to be protected by lock.
- We have the issue that when the refcount of endpoint drops to zero, we need to grab the lock and remove the endpoint from hashtable. But there is a race between client grab the lock to find the endpoint from hashtable and release of endpoint to remove itself from hashtable.
- Related issue https://github.com/awslabs/aws-c-s3/issues/202

*Description of changes:*

- We cannot always make sure the lock be grab to release the refcount. So that, we schedule a task to always be safe to grab lock and check the refcount and decide to remove it from hashtable or not

### DIFF
--- a/include/aws/s3/private/s3_client_impl.h
+++ b/include/aws/s3/private/s3_client_impl.h
@@ -78,7 +78,13 @@ struct aws_s3_endpoint {
 
     /* If the endpoint registered to the hashtable owned by client, set the pointer to client for unregister from the
      * hashtable once endpoint started to clean itself up.  */
-    struct aws_s3_client *client;
+    struct {
+        /* Protected by the client lock */
+        struct aws_s3_client *client;
+        struct aws_task *task;
+        bool task_scheduled;
+        size_t num_released;
+    } async_release;
 
     void *user_data;
 };
--- a/include/aws/s3/private/s3_client_impl.h
+++ b/include/aws/s3/private/s3_client_impl.h
@@ -318,7 +318,7 @@ AWS_S3_API
 void aws_s3_client_unlock_synced_data(struct aws_s3_client *client);
 
 AWS_S3_API
-struct aws_s3_endpoint *aws_s3_endpoint_acquire(struct aws_s3_endpoint *endpoint);
+struct aws_s3_endpoint *aws_s3_endpoint_acquire_synced(struct aws_s3_endpoint *endpoint);
 
 AWS_S3_API
 void aws_s3_endpoint_release(struct aws_s3_endpoint *endpoint);
--- a/include/aws/s3/private/s3_client_impl.h
+++ b/include/aws/s3/private/s3_client_impl.h
@@ -76,9 +76,9 @@ struct aws_s3_endpoint {
     /* Callback for when this endpoint completely shuts down. */
     aws_s3_endpoint_shutdown_fn *shutdown_callback;
 
-    /* True, if the endpoint is created by client. So, it need to be thread safe to manage the refcount via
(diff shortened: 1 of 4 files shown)


source/s3_client.c
```diff
@@ -693,11 +722,18 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
                 error_occurred = true;
                 goto unlock;
             }
-            endpoint->handled_by_client = true;
+            struct aws_task *async_release_task = aws_mem_calloc(client->allocator, 1, sizeof(struct aws_task));
+            aws_task_init(async_release_task, s_endpoint_release_task, endpoint, "se_endpoint_refcount_task");
+
+            endpoint->async_release.client = client;
+            endpoint->async_release.task = async_release_task;
+            endpoint->async_release.task_scheduled = false;
```
graebm, line 730: trivial: should this endpoint initialization code go in in s3_endpoint.c instead of s3_client.c?

include/aws/s3/private/s3_client_impl.h
```diff
@@ -76,9 +76,18 @@ struct aws_s3_endpoint {
...
-    /* True, if the endpoint is created by client. So, it need to be thread safe to manage the refcount via
-     * `aws_s3_client_endpoint_release` */
-    bool handled_by_client;
+    /* For S3 client to handle the hashtable of endpoints. */
+    struct {
+        /* Point to the client. */
+        struct aws_s3_client *client;
+        /* If set, the release of endpoint will be scheduled to run within the task to prevent the data race with client
+         */
+        struct aws_task *task;
```
graebm, line 85: why create a new task just for this? why not re-use the client's existing `process_work_task`?

we could keep an aws_array_list of `struct aws_s3_endpoint*` to be released

source/s3_client.c
```diff
@@ -575,6 +575,35 @@ struct aws_s3_request *aws_s3_client_dequeue_request_threaded(struct aws_s3_clie
...
+    size_t num_released = 0;
+    /* BEGIN CRITICAL SECTION */
+    {
+        aws_s3_client_lock_synced_data(client);
+        endpoint->async_release.task_scheduled = false;
+        size_t current_ref = aws_atomic_load_int(&endpoint->ref_count.ref_count);
+        num_released = endpoint->async_release.num_released;
+        endpoint->async_release.num_released = 0;
+        AWS_ASSERT(num_released <= current_ref);
+        if (num_released == current_ref) {
```
graebm, line 593: count might have changed from another thread

source/s3_endpoint.c
```diff
@@ -209,26 +209,27 @@ void aws_s3_endpoint_release(struct aws_s3_endpoint *endpoint) {
...
-    AWS_PRECONDITION(client);
-    AWS_PRECONDITION(endpoint->handled_by_client);
-
-    /* BEGIN CRITICAL SECTION */
-    {
-        aws_s3_client_lock_synced_data(client);
-        /* The last refcount to release */
-        if (aws_atomic_load_int(&endpoint->ref_count.ref_count) == 1) {
-            aws_hash_table_remove(&client->synced_data.endpoints, endpoint->host_name, NULL, NULL);
+    if (endpoint->async_release.task) {
```
graebm, line 212: not a fan of endpoint having 2 "modes", where 1 mode is never used

just needless complexity

graebm:
I took another swing at fixing this over in a new PR: https://github.com/awslabs/aws-c-s3/pull/209

PR #216, opened 2022-09-27
s3_request: fix broken requests_in_flight counter

### LINKED ISSUE #214: s3_request: num_requests_in_flight not reduced due to nil client field
### Problem description

While debugging a problem with a drastic reduction in transfer speed, I noticed that _the `client->stats.num_requests_in_flight` count does not go back to 0 at the end of the transfer_.

The last `s3_client` statement logged before the shutdown was:
```console
[INFO] 2022-09-27 15:10:09.977 S3ClientStats [140233490298624] id=0x7f8aa4310180 Requests-in-flight(approx/exact):0/998  Requests-preparing:0  Requests-queued:0  Requests-network(get/put/default/total):0/0/0/0  
Requests-streaming-waiting:0  Requests-streaming:0  Endpoints(in-table/allocated):0/0
[DEBUG] 2022-09-27 15:10:09.977 S3Client [140233490298624] id=0x7f8aa4310180 Client shutdown progress: starting_destroy_executing=0  body_streaming_elg_allocated=0  process_work_task_scheduled=0  process_work_ta
sk_in_progress=0  num_endpoints_allocated=0 finish_destroy=1
[DEBUG] 2022-09-27 15:10:09.977 S3Client [140233490298624] id=0x7f8aa4310180 Client finishing destruction.
```
Please note the 998  `stats.num_requests_in_flight` count in `Requests-in-flight(approx/exact):0/998`.

The `stats.num_requests_in_flight` is used to regulate the number of new requests that are sent, the high count of exact requests-in-flight (in contrast to 0 approximate in-flight requests) caused the request rate to drop drastically.

The following _debug log statement_ brought clarity:
```c
// source/s3_request.c
static void s_s3_request_destroy(void *user_data) {
    struct aws_s3_request *request = user_data;

    if (request == NULL) {
        return;
    }

    struct aws_s3_meta_request *meta_request = request->meta_request;

    if (meta_request != NULL) {
        struct aws_s3_client *client = meta_request->client;

        if (client != NULL) {
            aws_s3_client_notify_request_destroyed(client, request);  // <== NOT CALLED WHEN client is NULL
        }
    }
    if (request->tracked_by_client) { // <=== ADDED THIS STATEMENT TO LOGS
        AWS_LOGF_ERROR(
            AWS_LS_S3_REQUEST,
            "id=%p REQUEST DESTROY meta=%p  client=%p",
            (void *)request,
            (void *)meta_request,
            (void *) (meta_request == NULL ? NULL : meta_request->client));
    }
//...
}
```

In the log generated by the transfer, there were 1337 (actual number) entries of the following type:
```console
[ERROR] 2022-09-27 15:19:24.323 S3Request [139957177939712] id=0x7f4a4d66af00 REQUEST DESTROY meta=0x7f489dd9b800  client=(nil)
```
According to the above code, this means that the `stats.num_requests_in_flight` was off due to the `client` field being `NULL`.
As a _result_, the transfer speed dropped drastically, from full speed to only a few requests per second.

### What to fix

The `client` field is needed to accurately maintain the `stats.num_requests_in_flight` field responsible for the request-speed-control.

If possible, find the code path that sets `meta_request->client` to `NULL` before destroying the request.
Otherwise, a different approach is needed for maintaining the counter.

grrtrr:
This fixes a bug in computing the exact requests-in-flight.

These were not decremented in some case, due to the `client` field being `NULL` when the request was destroyed.

Instead of decrementing `num_requests_in_flight` and scheduling follow-on work when the request is destroyed,
perform this task when calling `aws_s3_meta_request_finished_request` (which signals that the request is done
and will not be retried), using a wrapper function.

*Issue #, if available:* Resolves #214


DmitriyMusatkin:
Cancel tests seem to sporadically fail after this change. Im guessing this is due to removal of s_s3_client_schedule_process_work_synced call, which results in cancel not wrapping up pending requests in some cases.

grrtrr:
> Cancel tests seem to sporadically fail after this change. Im guessing this is due to removal of s_s3_client_schedule_process_work_synced call, which results in cancel not wrapping up pending requests in some cases.


@DmitriyMusatkin - that is correct. My suggested implementation in #214 was short-sighted and buggy, since it only decreased the `num_requests_in_flight` after a connection was made (in `aws_s3_client_notify_connection_finished`). When a request is cancelled, processing that in `aws_s3_client_update_connections_threaded` lacked the `s_s3_client_schedule_process_work_synced` (which is done in `aws_s3_client_notify_connection_finished` and in `s_s3_client_prepare_callback_queue_request` on error).

I will update the notes in #214 and submit an update to this PR after some more testing.

DmitriyMusatkin (approved):

PR #223, opened 2022-10-14
fix: support endpoint with port e.g. http://127.0.0.1:9888

fboranek:
*Issue #, if available:* https://github.com/aws/aws-sdk-cpp/issues/1844

*Description of changes:*

Fix get/put object when endpoint have custom port in S3CrtClient

Make a client with this config:

```cc
    auto endpoint = std::string{"http://127.0.0.1:9888"};

    Aws::S3Crt::ClientConfiguration configClient;
    configClient.endpointOverride = Aws::String{str.data(), str.size()};
    configClient.executor = executor;
    configClient.enableHostPrefixInjection = false;
    configClient.scheme =
        boost::starts_with(endpoint, "https") ? Aws::Http::Scheme::HTTPS : Aws::Http::Scheme::HTTP;
```

In the endpoint it is needed hostname in form "127.0.0.1:9888" to properly generate HTTP header `Host: 127.0.0.1:9888`, but it is not suited for hostname resolution.
So if it port defined, it must be for resolution port stripped.


TingDaoK:
There is an `endpoint` configuration that supports endpoint with port. https://github.com/awslabs/aws-c-s3/blob/main/include/aws/s3/s3_client.h#L395-L401

For https://github.com/aws/aws-sdk-cpp/issues/1844, you will need to add the convert from `Aws::Http::URI &uri` to `struct aws_uri *` at the `InitCommonCrtRequestOption` [here](https://github.com/aws/aws-sdk-cpp/blob/main/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp#L444-L472). I'll let C++ team know.

Closing this PR as it's not the right solution for the issue..

fboranek:
@TingDaoK I followed this path you suggested (in time of patch), but there was another bug. The hostname was resolved well but the request failed due to check Host in HTTP request != host (`127.0.0.1:9888 != localhost`). However i cannot recall details, just remember it was related to the Http header "Host: 127.0.0.1:9888".

Do you have test for custom enpoind with port?

TingDaoK:
We do have test for custom endpoint with port. https://github.com/awslabs/aws-c-s3/blob/main/tests/s3_mock_server_tests.c

The uri is [here](https://github.com/awslabs/aws-c-s3/blob/main/tests/s3_tester.c#L30).
You can find how we configure the tests around [here](https://github.com/awslabs/aws-c-s3/blob/main/tests/s3_tester.c#L1214-L1242)

I don't get the error you mentioned above, `Host in HTTP request != host (127.0.0.1:9888 != localhost)`. Maybe it's an issue in your application or C++ SDK? The CRT client don't check the header value.

PR #224, opened 2022-10-20
Validate checksum fix

TingDaoK:
- Support list of checksum algorithms to validate
- Changed the behavior that:
   - When all the parts of the object checksum have been validated, we still tell user that we did not validate the checksum
   - When some part of the object checksum validation failed, we tell user checksum validation failed, but we did not validate the checksum.

TODO: 
- Tests for checksum validation failed. <- We don't have mock server, we cannot control that...


include/aws/s3/private/s3_checksums.h
```diff
@@ -26,6 +26,18 @@ struct aws_s3_checksum {
     bool good;
 };
 
+struct aws_s3_checksum_config_storage {
```
DmitriyMusatkin, line 29: storage seems a bit weird here, do we use similar naming anywhere else?
can we use options suffix for the checksum config passed in and keep internal as checksum_config?

include/aws/s3/private/s3_checksums.h
```diff
@@ -48,6 +60,8 @@ struct aws_input_stream *aws_checksum_stream_new(
     struct aws_byte_buf *checksum_output);
 
 /**
+ * TODO: properly support chunked encoding.
```
DmitriyMusatkin, line 63: nit: should we doc more here on what "properly" means or create a ticket for it?

include/aws/s3/private/s3_request.h
```diff
@@ -63,6 +63,8 @@ struct aws_s3_request {
 
     /* running checksum of the respone to an individual get part http request */
     struct aws_s3_checksum *request_level_running_response_sum;
+    /* The algorithm used to validate the checksum */
+    uint32_t validation_algorithm;
```
DmitriyMusatkin, line 67: why uint32_t and not enum type?

include/aws/s3/s3_client.h
```diff
@@ -318,6 +318,20 @@ struct aws_s3_checksum_config {
...
+     * Ignored when validate_response_checksum is not set.
+     * If not set all the algorithms will be selected as default behavior.
+     *
+     * The list of algorithms for user to pick up when validate the checksum. Client will pick up the algorithm from the
+     * list with the priority based on performance, and the algorithm sent by server. The priority based on performance
+     * is [CRC32C, CRC32, SHA1, SHA256].
+     *
+     * If the response checksum was validated by client, the result will indicate which algorithm was picked.
+     */
+    struct aws_array_list *validate_checksum_algorithms;
```
DmitriyMusatkin, line 334: add note on who owns the mem for this and what are exact types expected as elements.

source/s3_auto_ranged_get.c
```diff
@@ -632,6 +638,15 @@ static void s_s3_auto_ranged_get_request_finished(
                 ++auto_ranged_get->synced_data.num_parts_completed;
 
                 if (!request_failed) {
+
+                    if (request->did_validate) {
+                        if (auto_ranged_get->validation_algorithm == AWS_SCA_NONE) {
```
DmitriyMusatkin, line 643: comment on what its trying to achieve? im not clear on what its trying to do

source/s3_meta_request.c
```diff
@@ -1520,3 +1523,22 @@ void aws_s3_meta_request_result_clean_up(
 
     AWS_ZERO_STRUCT(*result);
 }
+
+bool aws_s3_meta_request_check_response_checksum_algorithm(
```
DmitriyMusatkin, line 1527: nit: naming is a bit awkward here. aws_s3_meta_request_checksum_config_has_algorithm?

include/aws/s3/private/s3_checksums.h
TingDaoK: let's create a ticket for it

include/aws/s3/private/s3_request.h
TingDaoK: was trying to avoid include the `s3_checksum.h` here.

include/aws/s3/private/s3_checksums.h
```diff
@@ -128,4 +142,8 @@ int aws_checksum_update(struct aws_s3_checksum *checksum, const struct aws_byte_
  */
 AWS_S3_API
 int aws_checksum_finalize(struct aws_s3_checksum *checksum, struct aws_byte_buf *output, size_t truncate_to);
-#endif
+
+AWS_S3_API
+void checksum_config_init(struct checksum_config *config_storage, const struct aws_s3_checksum_config *config);
```
DmitriyMusatkin, line 147: nit: still says storage here

DmitriyMusatkin (approved):
fix and ship

PR #228, opened 2022-11-04
thread options weren't meant to be zeroed out

graebm:
I stumbled across this code when I searching for uses of aws_thread.

You're supposed to call `aws_default_thread_options()` to initialize `aws_thread_options`, not zero it out.


graebm:
Oh, looks like this change already made it in...

PR #229, opened 2022-11-04
Alternative "make request" function to avoid race conditions

graebm:
**Issue:**
The current API requires users to be VERY CAREFUL to avoid the following race condition:
> If user stores the pointer returned by `aws_s3_client_make_meta_request()`,
> and then accesses the stored pointer from callbacks,
> the callbacks can fire on another thread before this function has returned,
> so the stored pointer doesn't exist yet, and kaboom.

See the race condition being fixed here:
* https://github.com/aws/aws-sdk-cpp/pull/2025

**Description of changes:**
Introduce new API for creating meta-requests, which helps avoid race conditions in user code

### DIFF
--- a/include/aws/s3/private/s3_meta_request_impl.h
+++ b/include/aws/s3/private/s3_meta_request_impl.h
@@ -126,6 +126,7 @@ struct aws_s3_meta_request {
     void *user_data;
 
     /* Customer specified callbacks. */
+    aws_s3_meta_request_init_callback_fn *init_callback;
     aws_s3_meta_request_headers_callback_fn *headers_callback;
     aws_s3_meta_request_receive_body_callback_fn *body_callback;
     aws_s3_meta_request_finish_fn *finish_callback;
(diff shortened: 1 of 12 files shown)


include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -184,9 +185,6 @@ struct aws_s3_meta_request {
         /* Note: this needs to be first for using AWS_CONTAINER_OF with the nested structure. */
         struct aws_linked_list_node node;
 
-        /* True if this meta request is currently in the client's list. */
-        bool scheduled;
```
graebm, line 188: I'm not sure what this was ever used for

We have branches that handle a meta-request being re-added to the client, but that's not possible. I was touching code near those branches and was like "how do I handle this ... thing that can't actually happen"

so instead of thinking how I'd handle an impossible bit of state, I removed the state.

tests/s3_tester.c
```diff
@@ -396,7 +459,7 @@ static bool s_s3_tester_have_meta_requests_finished(void *user_data) {
     AWS_PRECONDITION(user_data);
     struct aws_s3_tester *tester = (struct aws_s3_tester *)user_data;
 
-    return tester->synced_data.meta_requests_finished > 0;
+    return tester->synced_data.meta_requests_finished;
```
graebm, line 462: There's no real change here. I changed the type elsewhere from a "1 bit bool" to an actual bool:
~uint32_t meta_requests_finished : 1;~ -> `bool meta_requests_finished;`

there is no need to optimize for space like this in testing-only structs

graebm:
This is a big change that no one is clamoring for. Closing for now...

PR #231, opened 2022-11-14
Mock s3 test

TingDaoK:
*Issue #, if available:*

- We lack of ability to test error cases, as we don't have a mock S3 server

*Description of changes:*

- Added a mock S3 server framework that support multipart upload for now.

TODO:
- supports all the type of request we made.
- add helper for test to write the response during runtime.

### DIFF
--- a/.builder/actions/mock_server_setup.py
+++ b/.builder/actions/mock_server_setup.py
@@ -19,14 +19,13 @@ class MockServerSetup(Builder.Action):
 
     def run(self, env):
         self.env = env
-        python_path = env.config['variables'].get('python')
+        python_path = sys.executable
         # install dependency for mock server
-        result = self.env.shell.exec(python_path,
-                                     '-m', 'pip', 'install', 'h11', 'trio')
-        if result.returncode != 0:
-            print(
-                "Mock server failed to setup, skip the mock server tests.", file=sys.stderr)
-            exit(-1)
+        self.env.shell.exec(python_path,
+                            '-m', 'pip', 'install', 'h11', 'trio', check=True)
+        # check the deps can be import correctly
+        self.env.shell.exec(python_path,
+                            '-c', 'import h11, trio', check=True)
 
         # set cmake flag so mock server tests are enabled
         env.project.config['cmake_args'].append(


.builder/actions/mock_server_setup.py
```diff
@@ -0,0 +1,40 @@
...
+            python_execute = "python"
+        else:
+            python_execute = "python3"
+
+        # install dependency for mock server
+        result = self.env.shell.exec(python_execute,
+                                     '-m', 'install', 'h11', 'trio')
+        if result.returncode != 0:
+            print("Mock server failed to setup, skip the mock server tests.")
+            return
```
TingDaoK, line 30: I am not sure which is better, error out or return..

graebm: Is it failing to install anywhere? If it works everywhere, then make it an error

If we just return: there's a chance that setup starts failing everywhere, but we never realize because it's a silent warning

If we make it an error: then you can end up spending waaaaay too much time trying to get it working on esoteric/old platforms like manylinux1.

.builder/actions/mock_server_setup.py
```diff
@@ -0,0 +1,40 @@
...
+    def run(self, env):
+        self.env = env
+        if os.name == 'nt':
+            python_execute = "python"
+        else:
+            python_execute = "python3"
+
+        # install dependency for mock server
+        result = self.env.shell.exec(python_execute,
+                                     '-m', 'install', 'h11', 'trio')
```
graebm, line 27: Use the string "{python}"
Builder will substitute the correct python path for this image

.builder/actions/mock_server_setup.py
```diff
@@ -0,0 +1,40 @@
...
+
+        # set cmake flag so mock server tests are enabled
+        env.project.config['cmake_args'].append(
+            '-DENABLE_MOCK_SERVER_TESTS=ON')
+
+        base_dir = os.path.dirname(os.path.realpath(__file__))
+        dir = os.path.join(base_dir, "..", "..", "tests", "mock_s3_server")
+        os.chdir(dir)
+
+        subprocess.Popen([python_execute, "mock_s3_server.py"])
```
graebm, line 40: add a comment that this is launching a background process that runs forever

graebm: maybe we could do something to try and clean this up afterwards?

maybe register an [atexit](https://docs.python.org/3/library/atexit.html) handler that tries to kill the background process?

.builder/actions/mock_server_setup.py
TingDaoK: current it's not failing from anywhere. But, two weeks ago, I think github ubuntu 20 image was failing to install trio for "I don't know" reason. Let's make it fail and heck around the image that doesn't support it.

.builder/actions/mock_server_setup.py
```diff
@@ -0,0 +1,43 @@
...
+    def run(self, env):
+        self.env = env
+        python_path = env.config['variables'].get('python')
+        # install dependency for mock server
+        result = self.env.shell.exec(python_path,
+                                     '-m', 'pip', 'install', 'h11', 'trio')
+        if result.returncode != 0:
+            print(
+                "Mock server failed to setup, skip the mock server tests.", file=sys.stderr)
+            exit(-1)
```
graebm, line 29: you can do `self.env.shell.exec(..., check=True)`

and then you don't need to add any extra code for checking the return code

.builder/actions/mock_server_setup.py
```diff
@@ -0,0 +1,43 @@
...
+class MockServerSetup(Builder.Action):
+    """
+    Set up this machine for running the mock server test
+
+    This action should be run in the 'pre_build_steps' or 'build_steps' stage.
+    """
+
+    def run(self, env):
+        self.env = env
+        python_path = env.config['variables'].get('python')
```
graebm, line 22: you had good luck using `sys.executable` instead here

.builder/actions/mock_server_setup.py
```diff
@@ -0,0 +1,43 @@
...
+
+        base_dir = os.path.dirname(os.path.realpath(__file__))
+        dir = os.path.join(base_dir, "..", "..", "tests", "mock_s3_server")
+        os.chdir(dir)
+
+        p = subprocess.Popen([python_path, "mock_s3_server.py"])
+
+        @atexit.register
+        def close_mock_server():
+            p.terminate()
```
graebm, line 43: neato

graebm (approved):

PR #235, opened 2022-11-18
Pause resume update

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Fix handling of pause after request completes.
Switched pause interface to return token object instead of serializing everything to a string.


include/aws/s3/s3.h
```diff
@@ -35,6 +35,7 @@ enum aws_s3_errors {
     AWS_ERROR_S3_RESUME_FAILED,
     AWS_ERROR_S3_OBJECT_MODIFIED,
     AWS_ERROR_S3_NON_RECOVERABLE_ASYNC_ERROR,
+    AWS_ERROR_S3_PAUSE_FAILED_REQUEST_COMPLETE,
```
TingDaoK, line 38: ~COMPLETE~ -> COMPLETED?

source/s3_auto_ranged_put.c
```diff
@@ -1184,72 +1154,52 @@ static void s_s3_auto_ranged_put_request_finished(
...
 
-        aws_byte_buf_clean_up(&result_string_buf);
-        aws_json_value_destroy(root);
+        /**
+         * Cancels the meta request using the PAUSED flag to avoid deletion of uploaded parts.
+         * This allows the client to resume the upload later, setting the persistable state in the meta request options.
+         */
+        aws_s3_meta_request_set_fail_synced(meta_request, NULL, AWS_ERROR_S3_PAUSED);
+    } else {
+        AWS_LOGF_DEBUG(AWS_LS_S3_META_REQUEST, "Cannot pause request since its already complete");
```
TingDaoK, line 1197: This should be ERROR level log.

include/aws/s3/s3_client.h
```diff
@@ -491,19 +492,95 @@ AWS_S3_API
...
-int aws_s3_meta_request_pause(struct aws_s3_meta_request *meta_request, struct aws_string **out_resume_token);
+int aws_s3_meta_request_pause(
+    struct aws_s3_meta_request *meta_request,
+    struct aws_s3_meta_request_resume_token **out_resume_token);
+
+struct aws_s3_upload_resume_token_options {
+    struct aws_byte_cursor upload_id;
+    size_t part_size;
+    size_t total_num_parts;
+};
```
TingDaoK, line 522: add some document about this option?

TingDaoK (approved):

include/aws/s3/s3_client.h
```diff
@@ -491,19 +492,100 @@ AWS_S3_API
...
+ * To resume an upload that was paused, supply the serialized resume token in the meta
+ * request options structure member aws_s3_meta_request_options.resume_token.
  * The upload can be resumed either from the same client or a different one.
- * Resume token is opaque with format varying based on operation.
- * Clients should not parse the token. For format details refer to pause method comments for a given operation.
  * Resume token will be set to null in case of failures.
+ * Note: similar to cancel pause does not cancel requests already in flight and
+ * and parts might complete after pause is requested.
+ * Note: pausing meta request before it starts will succeed with NULL token
+ * returned and will pause request. (resuming with null token is equivalent to
```
graebm, line 506: this is really weird, to return AWS_OP_SUCCESS, but then the token is NULL

do you think users will want to treat not-yet-started requests differently than OTHER paused requests?

maybe just return a token that knows it was never really started, and there are functions the user can query to learn that fact.

or report a unique error like `AWS_ERROR_S3_REQUEST_NOT_YET_STARTED`, instead of reporting success but leaving the token NULL

graebm: The not-yet-started case is very similar to the already-complete case.
Either case could come up when the user calls pause(), but one reports and ERROR and the other does not.

we should consider doing the same thing in both cases...

either both report a unique error
or both return a token, that the user can query to realize it's actually already complete, or actually it never really started.

source/s3_auto_ranged_put.c
```diff
@@ -1184,72 +1154,52 @@ static void s_s3_auto_ranged_put_request_finished(
...
+     *     upload
+     *   - after upload - return already completed error, token is NULL and
+     *     nothing to cancel
+     */
 
-        aws_json_value_add_to_object(
-            root,
-            aws_byte_cursor_from_c_str("total_num_parts"),
-            aws_json_value_new_number(meta_request->allocator, (double)auto_ranged_put->synced_data.total_num_parts));
+    if (!auto_ranged_put->synced_data.complete_multipart_upload_sent) {
```
graebm, line 1177: we can still get into weird scenarios where we've only *constructed* the final CompleteMultipartUpload request, but it hasn't actually finished yet...

we report error REQUEST_COMPLETED ... but that final request still might fail

once the meta-request is in this state, with a CompleteMultipartUpload in flight, it's basically unpausable. It must complete either successfully or unsuccessfully, but we're unable to pause it and resume later

This is a real pickle. I can't think of any synchronous solution. The only solution I can come up with is delivering the pause-token asynchronously, so that we can be 100% certain any pending requests get cancelled or finished before the pause token is delivered to the user

include/aws/s3/s3_client.h
```diff
@@ -491,19 +492,100 @@ AWS_S3_API
...
+};
+
+/**
+ * Create upload resume token from persisted data.
+ * Note: Data required for resume token varies per operation.
+ */
+AWS_S3_API
+struct aws_s3_meta_request_resume_token *aws_s3_meta_request_resume_token_new_upload(
+    struct aws_allocator *allocator,
+    const struct aws_s3_upload_resume_token_options *options);
```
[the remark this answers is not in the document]
DmitriyMusatkin: yes. serialization has been moved to be callers responsibility, so this function provides a way to reconstruct token

include/aws/s3/s3_client.h
```diff
@@ -491,19 +492,100 @@ AWS_S3_API
...
+ */
+AWS_S3_API
+size_t aws_s3_meta_request_resume_token_total_num_parts(struct aws_s3_meta_request_resume_token *resume_token);
+
+/*
+ * Num parts completed.
+ * Returns 0 for invalid tokens.
+ */
+AWS_S3_API
+size_t aws_s3_meta_request_resume_token_num_parts_completed(struct aws_s3_meta_request_resume_token *resume_token);
```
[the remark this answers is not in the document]
DmitriyMusatkin: fair enough. this should be a precondition instead

include/aws/s3/s3.h
```diff
@@ -35,6 +35,7 @@ enum aws_s3_errors {
     AWS_ERROR_S3_RESUME_FAILED,
     AWS_ERROR_S3_OBJECT_MODIFIED,
     AWS_ERROR_S3_NON_RECOVERABLE_ASYNC_ERROR,
+    AWS_ERROR_S3_PAUSE_FAILED_REQUEST_COMPLETED,
```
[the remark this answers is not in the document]
DmitriyMusatkin: hmm, i went back and forth between already complete and request completed. my only concern with already complete is that it does not really sound like an error on its own outside of pause context

include/aws/s3/s3_client.h
```diff
@@ -491,19 +492,100 @@ AWS_S3_API
...
+ * Note: pausing meta request after it completes will fail with AWS_ERROR_S3_PAUSE_FAILED_REQUEST_COMPLETE
  * @param meta_request pointer to the aws_s3_meta_request of the upload to be paused
- * @param resume_token outputs the json string with the state that can be used to resume the operation.
- * @return
+ * @param resume_token resume token
+ * @return error code.
  */
 AWS_S3_API
-int aws_s3_meta_request_pause(struct aws_s3_meta_request *meta_request, struct aws_string **out_resume_token);
+int aws_s3_meta_request_pause(
```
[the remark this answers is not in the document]
DmitriyMusatkin: i agree, but as originally defined old interface have a lot of issue. this api is only about 4 month old and java is the only binding we have defined it in and im not sure how many people manually use it, but i think we can afford to break backwards compat here to improve the interface

include/aws/s3/s3_client.h
DmitriyMusatkin: i actually had a separate error for it while i was playing around with it. 
my thinking on returning null is that null signifies basically that request should start from beginning, which is the case with pause where we haven't started processing it yet. but lets have a quick discussion about it on mon

source/s3_auto_ranged_put.c
DmitriyMusatkin: if we define the pause operation to work only during the window when crt is actively uploading parts, then i dont really see the issue with throwing error when we started complete. its just how operation works by definition and its synchronous.

lets discuss more in depth. i have a couple concerns with making it async, cause i dont think its a good user experience and as mentioned above i dont really see no supporting pause when complete started as a deal breaker, cause from customer perspective there is nothing to pause anymore, since all data is already uploaded

include/aws/s3/s3_client.h
```diff
@@ -491,19 +492,100 @@ AWS_S3_API
 void aws_s3_meta_request_cancel(struct aws_s3_meta_request *meta_request);
 
 /**
- * In order to pause an ongoing upload, call aws_s3_meta_request_pause(). It will return a resume token that can be
- * persisted and used to resume the upload. To resume an upload that was paused, supply the resume token in the meta
- * request options structure member aws_s3_meta_request_options.persistable_state.
+ * Note: pause is currently only supported on upload requests.
+ * In order to pause an ongoing upload, call aws_s3_meta_request_pause() that
+ * will return resume token. Token can be used to query the state of operation
+ * at the pausing time. Use aws_byte_buf_init_from_s3_resume_token to serialize token.
```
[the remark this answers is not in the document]
graebm: nevermind ... serialization is the user's responsibility now

include/aws/s3/s3_client.h
```diff
@@ -495,17 +495,22 @@ void aws_s3_meta_request_cancel(struct aws_s3_meta_request *meta_request);
...
+ * To resume an upload that was paused, supply resume token in the meta
  * request options structure member aws_s3_meta_request_options.resume_token.
  * The upload can be resumed either from the same client or a different one.
- * Resume token will be set to null in case of failures.
+ * Corner cases for resume upload are as follows:
+ * - upload is not MPU - fail with pause not supported error
+ * - pausing before MPU is created - NULL resume token returned. NULL resume
+ *   token is equivalent to restarting upload
+ * - pausing in the middle of part transfer - return resume token. scheduling of
+ *   new part uploads stops.
```
graebm, line 507: I'm curious about where work resumes in the common case that we're in the middle of uploading...

Say an upload has 4 parts, and at the moment the user calls pause(), the state is:
- part 1 has been completely uploaded
- part 2 is in the middle of uploading
- part 3 is scheduled, but hasn't sent any network traffic yet
- part 4 hasn't been touched

... when the user uses that resume token, will that next upload send:
- just part 4 (because the last session already scheduled 1,2,3)
- OR: parts 2, 3, 4 (because we re-send any parts we're not 100% sure completed)
- or something else...

whichever it is, make a comment to the reader to dispel any questions they might have.
Something like "On resuming, partially uploaded parts will be re-sent"
or if it doesn't work that way "You must want for the current upload operation to complete before using the resume token"

oh yeah, DO they need to avoid overlapping a pause with a resume? If we're re-uploading parts that were mid-upload when we paused ... we might actually have two overlapping requests for the same exact part !?!?

include/aws/s3/s3_client.h
```diff
@@ -398,12 +399,12 @@ struct aws_s3_meta_request_options {
...
     /**
      * Optional.
-     * For meta requests that support pause/resume (e.g. PutObject), the resume token returned by
+     * For meta requests that support pause/resume (e.g. PutObject), serialized resume token returned by
      * aws_s3_meta_request_pause() can be provided here.
      * Note: If PutObject request specifies a checksum algorithm, client will calculate checksums while skipping parts
      * from the buffer and compare them them to previously uploaded part checksums.
      */
-    const struct aws_byte_cursor *resume_token;
+    struct aws_s3_meta_request_resume_token *resume_token;
```
graebm, line 407: rambling: there's sort of an argument for our _acquire()/_release() functions taking `const` pointers, even though they do modify the object's internal state, just so we can pass around const references in cases like this

but none of our acquire/release functions currently do this. Just rambling....

include/aws/s3/s3_client.h
graebm: we talked
when resuming, the server tells us what parts it's already received. nice.

include/aws/s3/s3_client.h
```diff
@@ -495,17 +495,22 @@ void aws_s3_meta_request_cancel(struct aws_s3_meta_request *meta_request);
...
+ * Corner cases for resume upload are as follows:
+ * - upload is not MPU - fail with pause not supported error
+ * - pausing before MPU is created - NULL resume token returned. NULL resume
+ *   token is equivalent to restarting upload
+ * - pausing in the middle of part transfer - return resume token. scheduling of
+ *   new part uploads stops.
+ * - pausing after completeMPU started - return resume token. if s3 cannot find
+ *   find associated MPU id when resuming with that token and num of parts
+ *   uploaded equals to total num parts, then operation is no op. Otherwise
+ *   operation fails.
```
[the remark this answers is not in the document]
graebm: we talked. clarify here that, in the no-op case, they won't get any response headers

also clarify in the on-headers-callback that it won't fire in this case

include/aws/s3/private/s3_auto_ranged_put.h
```diff
@@ -25,6 +25,8 @@ struct aws_s3_auto_ranged_put {
     /* Initialized either during creation in resume flow or as result of create multipart upload during normal flow. */
     struct aws_string *upload_id;
 
+    struct aws_s3_meta_request_resume_token *resume_token;
```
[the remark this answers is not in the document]
DmitriyMusatkin: added comment. the only reason its not const is because we call release/acquire on it and as mentioned in another comment those require non-const unfortunately.

PR #238, opened 2022-12-06
meta request: support for should_continue cancellation callback

grrtrr:
Add a boolean `continue_callback`, which polls a request-cancellation flag set by the user. This matches the use-case of the `ContinueRequestHandler` / `ContinueRequest` / `ShouldContinue` boolean functions used by `aws-sdk-cpp`.

Passive polling is preferred over active cancellation, since the latter would require to store a reference to the s3 meta request. The polling function is evaluated at the following points during the lifetime of the meta request:
1. _Request preparation_ (`aws_s3_meta_request_prepare_request`).
2. _Request update_ (the `.update` handler invoked via `aws_s3_meta_request_update`).
3. _Http stream completion_ (`.on_complete` handler invoking `aws_s3_meta_request_send_request_finish_default`).
4. _Writing download data_ (`aws_s3_meta_request_stream_response_body_synced`).

If a `continue_callback` is provided and if it returns `false` at any of these 4 evaluation times, the meta request completes with the `error_code` `AWS_ERROR_S3_CANCELED`.

By using a boolean flag in the body of the function, the polling overhead can be kept minimal.


grrtrr:
Hi @jmklix why was this closed without comments. Is there an alternative that can be used and where is it documented?

As it stands, `S3Crt` requests can not be canceled at the moment.

TingDaoK:
with https://github.com/awslabs/aws-c-s3/pull/395, we now cancel all the in-flight requests. It's over-complicate the behavior for user to control for each requests. Unless you have a strong opinion with use case, I don't see a point to keep the requests in-flight when you decided to cancel the meta request.

PR #248, opened 2023-01-13
Fix error handling for meta request init

TingDaoK:
- add test for `aws_s3_meta_request_auto_ranged_put_new` and `aws_s3_meta_request_auto_ranged_get_new`
- Fix error handling for `aws_s3_meta_request_init_base`.


include/aws/s3/private/s3_auto_ranged_get.h
```diff
@@ -50,11 +50,15 @@ struct aws_s3_auto_ranged_get {
     struct aws_string *etag;
 };
 
+AWS_EXTERN_C_BEGIN
+
 /* Creates a new auto-ranged get meta request.  This will do multiple parallel ranged-gets when appropriate. */
-struct aws_s3_meta_request *aws_s3_meta_request_auto_ranged_get_new(
+AWS_S3_API struct aws_s3_meta_request *aws_s3_meta_request_auto_ranged_get_new(
```
DmitriyMusatkin, line 56: for team discussion: i really dont like idea of exporting internal symbols just so we can test those calls. we need better approach for this. maybe a separate test only visibility level

DmitriyMusatkin (approved):

PR #249, opened 2023-01-19
Fix Host header not including port

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
There is some overlap between meta request endpoint and Host header. Clarified the docs what happens depending on which one is used. 
Fix logic comparing the two if both are set with ports. 
Made Host header optional if endpoint is provided

### DIFF
--- a/include/aws/s3/s3_client.h
+++ b/include/aws/s3/s3_client.h
@@ -396,7 +396,7 @@ struct aws_s3_meta_request_options {
      * Optional.
      * Endpoint override for request. Can be used to override scheme and port of
      * the endpoint.
-     * There is some overlap between Host header and Endpoint and cornet cases
+     * There is some overlap between Host header and Endpoint and corner cases
      * are handled as follows:
      * - Only Host header is set - Host is used to construct endpoint. https is
      *   default with corresponding port
--- a/include/aws/s3/s3_client.h
+++ b/include/aws/s3/s3_client.h
@@ -395,7 +395,7 @@ struct aws_s3_meta_request_options {
     /**
      * Optional.
      * Endpoint override for request. Can be used to override scheme and port of
-     * the endpoint. 
+     * the endpoint.
      * There is some overlap between Host header and Endpoint and cornet cases
      * are handled as follows:
      * - Only Host header is set - Host is used to construct endpoint. https is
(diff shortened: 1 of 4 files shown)


include/aws/s3/s3_client.h
```diff
@@ -394,9 +394,16 @@ struct aws_s3_meta_request_options {
...
+     * Endpoint override for request. Can be used to override scheme and port of
+     * the endpoint.
+     * There is some overlap between Host header and Endpoint and cornet cases
+     * are handled as follows:
+     * - Only Host header is set - Host is used to construct endpoint. https is
+     *   default with corresponding port
+     * - Only endpoint is set - Host header is created from endpoint. Port and
+     *   Scheme from endpoint is used.
+     * - Both Host and Endpoint is set - Host header must match Authority of
+     *   Endpoint uri. Port and Scheme from endpoint is used.
```
TingDaoK, line 406: How will `Scheme` be used?

source/s3_client.c
```diff
@@ -724,8 +774,11 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
         }
     }
 
-    struct aws_byte_cursor host_header_value;
+    if (s_update_host_header_based_on_endpoint_override(client, message_headers, options->endpoint)) {
```
TingDaoK, line 777: trivial: maybe rename to something `apply_endpoint_override`? We don't necessarily update host header.

source/s3_client.c
```diff
@@ -768,14 +825,23 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
     {
         aws_s3_client_lock_synced_data(client);
 
-        struct aws_string *endpoint_host_name = aws_string_new_from_cursor(client->allocator, &host_header_value);
+        struct aws_uri host_uri;
+        if (aws_uri_init_parse(&host_uri, client->allocator, &host_header_value)) {
+            error_occurred = true;
+            goto unlock;
+        }
```
TingDaoK, line 832: We can skip this init again if the `options->endpoint` is set

source/s3_client.c
DmitriyMusatkin: nice. i like that name better. I was struggling to come up with a good name for this

include/aws/s3/s3_client.h
DmitriyMusatkin: scheme is used to determine whether its https or http. we can explicitly say that, but i though it was clear enough

TingDaoK (approved):

PR #252, opened 2023-01-26
More tests 3

TingDaoK:
- More tests:
   - Invalid response from server
   - Mismatch checksum from server
   - Different type of checksum
   - This should cover all those case for GetObject Requests   

- Fix bugs in the mock server
  - support HEAD of GetObject
  - Default to a length if `Range` header not exist


tests/s3_data_plane_tests.c
```diff
@@ -2675,50 +2675,57 @@ static int s_test_s3_round_trip_default_get_fc(struct aws_allocator *allocator,
     struct aws_byte_buf path_buf;
     AWS_ZERO_STRUCT(path_buf);
 
-    ASSERT_SUCCESS(aws_s3_tester_upload_file_path_init(
-        allocator, &path_buf, aws_byte_cursor_from_c_str("/prefix/round_trip/test_default_fc.txt")));
+    /* I know it's ugly, but I don't want to waste time to refact all those flexible checksum tests */
```
graebm, line 2678: ~to refact~ -> refactoring? to refactor?

but more importantly, I don't know what "it" is, why it's ugly, or what "all those" other tests are. please make this comment make sense to a newcomer

are you talking about the loop is ugly? a loop seems ok to me

but if you wanted N explicit tests, instead of 1 test that checks all algorithms, just make this a helper function, and set up N tests that call it

tests/s3_data_plane_tests.c
```diff
@@ -2675,50 +2675,57 @@ static int s_test_s3_round_trip_default_get_fc(struct aws_allocator *allocator,
...
+    for (int algorithm = AWS_SCA_INIT; algorithm <= AWS_SCA_END; ++algorithm) {
+        char object_path_sprintf_buffer[128] = "";
+        snprintf(
+            object_path_sprintf_buffer,
+            sizeof(object_path_sprintf_buffer),
+            "/prefix/round_trip/test_default_fc_%d.txt",
```
graebm, line 2684: is this file in source control?
if it is, could we use the checksum's name instead of an int?
maybe use aws_get_http_header_name_from_algorithm() to get the string?

graebm (approved):
fix and ship

tests/s3_data_plane_tests.c
TingDaoK: I would image we can have something like a helper to handle different type and different checksum algorithm instead of lot of really similar code around those flexible checksum tests and for loops with lot of code.

PR #254, opened 2023-01-31
Put Object Tests

TingDaoK:
*Issue #, if available:*

- Fixed a bug when resume a request that has multiple page of ListParts response, it hangs
- For edge case that resume after completed, we will still sent abortMPU.

*Description of changes:*

- Added tests for Resume upload with Mock server for edge cases
- Supports ListParts from mock server.
- Revamp some putObject code to be more clear
- Added tests to cover some PutObject error handling.


source/s3_request.c
```diff
@@ -92,5 +92,6 @@ static void s_s3_request_destroy(void *user_data) {
     aws_s3_request_clean_up_send_data(request);
     aws_byte_buf_clean_up(&request->request_body);
     aws_s3_meta_request_release(request->meta_request);
+
```
DmitriyMusatkin, line 95: nit: extra white space

source/s3_auto_ranged_put.c
```diff
@@ -502,6 +502,10 @@ static bool s_s3_auto_ranged_put_update(
                 if (auto_ranged_put->upload_id == NULL) {
                     goto no_work_remaining;
                 }
+                if (auto_ranged_put->base.synced_data.finish_result.error_code == AWS_ERROR_SUCCESS) {
+                    /* Not sending abort when success */
+                    goto no_work_remaining;
```
graebm, line 507: I'm have a really hard time following the flow of this function...

If we get here, what is it that finished successfully?

source/s3_auto_ranged_put.c
```diff
@@ -979,6 +991,8 @@ static void s_s3_auto_ranged_put_request_finished(
                 }
 
                 auto_ranged_put->synced_data.list_parts_completed = !has_more_results;
+                auto_ranged_put->synced_data.list_parts_sent =
+                    !has_more_results; /* Unset list_parts_send if there are more to send */
```
graebm, line 995: put the comment on its own line. If your trailing comment forces a line-wrap, it all just gets hard to read

add comments, or something, to explain how this will result in another ListParts request being sent out, using the continuation token.

also the two `bool = !otherbool;` in a row is a bit confusing. Maybe code it like you'd explain it.

```
if (has_more_results) {
    /* explain about how this results in another ListParts request */
    auto_ranged_put->synced_data.list_parts_sent = false;
} else {
    auto_ranged_put->synced_data.list_parts_completed = true;
}

source/s3_auto_ranged_put.c
TingDaoK: updated the comment. This function is truly hard to follow. Maybe we can change all those boolean about each request state to something better...

DmitriyMusatkin (approved):

PR #255, opened 2023-02-01
More S3 client test

TingDaoK:
- add tests for s3_client.c


source/s3_request_messages.c
```diff
@@ -266,7 +266,7 @@ struct aws_http_message *aws_s3_create_multipart_upload_message_new(
         false /*exclude_x_amz_meta*/);
 
     if (message == NULL) {
-        goto error_clean_up;
```
waahm7, line 269: trivial: I prefer the original implementation. This forces us to not do any allocation before `aws_s3_message_util_copy_http_message_no_body_filter_headers`. We are not doing any right now but it might change in future.

include/aws/s3/private/s3_request_messages.h
```diff
@@ -40,7 +40,7 @@ struct aws_http_message *aws_s3_message_util_copy_http_message_no_body_filter_he
 /* Copy headers from one message to the other and exclude specific headers.
  * exclude_x_amz_meta controls whether S3 user metadata headers (prefixed with "x-amz-meta) are excluded.*/
 AWS_S3_API
-int aws_s3_message_util_copy_headers(
```
waahm7, line 43: Do we really want to break API for this change?

source/s3_client.c
```diff
@@ -668,13 +668,9 @@ int s_apply_endpoint_override(
...
-    if (aws_http_headers_get(message_headers, g_host_header_name, &host_value)) {
-        AWS_LOGF_ERROR(
-            AWS_LS_S3_CLIENT,
-            "id=%p Cannot create meta s3 request; message provided in options does not have a 'Host' header.",
-            (void *)client);
-        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
-    }
+    int error = aws_http_headers_get(message_headers, g_host_header_name, &host_value);
+    AWS_ASSERT(!error);
+    (void)error;
```
waahm7, line 673:
```suggestion
    AWS_ASSERT(aws_http_headers_get(message_headers, g_host_header_name, &host_value));
```

source/s3_request_messages.c
```diff
@@ -979,12 +953,10 @@ int aws_s3_message_util_copy_headers(
             }
         }
 
-        if (aws_http_message_add_header(dest_message, header)) {
-            return AWS_OP_ERR;
-        }
+        error |= aws_http_message_add_header(dest_message, header);
```
waahm7, line 956: This is hard to read. It would be better to just do `AWS_ASSERT(aws_http_message_add_header(dest_message, header))`

source/s3_client.c
TingDaoK: AWS_ASSERT will be optimized out for release build.

waahm7: What about precondition if we are asserting that this will never fail?

source/s3_request_messages.c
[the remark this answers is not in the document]
TingDaoK: That's the first thing in the function, everything in the function is based on the message. 

As long as people in the future be careful about error handling, we should be good. Just make sure before any allocation, or the first allocation failed, just return, otherwise, clean up.

source/s3_client.c
TingDaoK: If anything went wrong, we can catch the error during our tests and CI, which basically are debug build. But, anything went wrong from there, will be a programming bug, instead of something can really happen. 

But, yeah, in theory, that should never fail. Just doing this to respect the return code.

waahm7: A fatal assert might be better. We do the same in Swift. If something slips through our tests, this will be a hard to debug bug because of this weird ignoring the error in production (which will cause failure somewhere else) but trying to catch it in debug.

TingDaoK: The logic here is host header has already been set by the code [above](https://github.com/awslabs/aws-c-s3/pull/255/files#diff-5ee13dd0ee005b828bde6d60ff89582070983e3cb7e5d92389b629a9850c27f8R651-R667). 

It should never ever error out.

waahm7 (approved):

PR #256, opened 2023-02-03
Fix proxy options copy in s3_new

waahm7:
*Issue #, if available:*

*Description of changes:*
- Use `aws_http_proxy_config_new_from_proxy_options_with_tls_info`
- Cleanup region


source/s3_client.c
```diff
@@ -365,6 +359,14 @@ struct aws_s3_client *aws_s3_client_new(
         }
     }
 
+    if (client_config->proxy_options) {
+        client->proxy_config = aws_http_proxy_config_new_from_proxy_options_with_tls_info(
```
TingDaoK, line 363: trivial: you can use `client_config->tls_mode == AWS_MR_TLS_ENABLED` instead of `client->tls_connection_options != NULL`

source/s3_client.c
```diff
@@ -451,6 +453,8 @@ struct aws_s3_client *aws_s3_client_new(
     return client;
 
 on_error:
+    aws_string_destroy(client->region);
```
TingDaoK, line 456: Good catch, I missed it from the `test_s3_client_create_error`, can you replace [here](https://github.com/awslabs/aws-c-s3/blob/main/tests/s3_data_plane_tests.c#L62) with the following code?
```
    struct aws_s3_client_config client_config = {
        .region = g_test_s3_region,
        .client_bootstrap = tester.client_bootstrap,
    };
```

source/s3_client.c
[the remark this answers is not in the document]
waahm7: Thanks, we can't handle that in binding because we will have to handle it in every binding. This code is already being used by sdk-cpp. I will document this.

source/s3_client.c
waahm7: We can't fail the test if we add bootstrap to S3 client.

PR #266, opened 2023-03-10
New `send_filepath` option, for faster uploads.

graebm:
**Issue:**
The Java SDK's S3 Transfer Manager, which uses aws-c-s3, is slower to upload a single large file than what we see with a raw C/C++ sample. Profiling reveals that the bottleneck is the file reading callbacks.

Java isn't necessarily slower than C at reading files, but there's a mismatch between the way Java's Reactive Stream API (aka [Flow](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html)), and the way aws-c-s3 asks for data, that leads to an extra copy. Since file reading is the bottleneck, that extra copy really has an impact on overall performance.

**Description of changes:**
Offer a new `send_filepath` option. When used, aws-c-s3 will handle opening the file and reading from it in whatever way is most efficient, without calling back into Java.

**More rambling:**
We did something similar in the aws-crt-python bindings, where it really did just seem that Python was slower than C (due to GIL? extra allocations? we didn't dig too hard). We worked around it the same way I'm doing here: letting the user [pass the filepath to C](https://github.com/awslabs/aws-crt-python/blob/fb2bf2c6404ca45b7b96e22e3fcc3dd39c3c309d/awscrt/s3.py#L173-L176) so we can avoid interacting with the Python interpreter while reading. Seems like a good idea to just add this trick to the core aws-c-s3 implementation so we don't need to keep doing it in every language binding.

Also, this opens the door to doing more tricks in aws-c-s3 to speed up file reading. Like maybe reading in parallel using multiple threads.


bretambrose:
> 
> Also, this opens the door to doing more tricks in aws-c-s3 to speed up file reading. Like maybe reading in parallel using multiple threads.
> 

Stating the obvious, but just want to call out if that ever happens it should be optional, since commodity storage hardware may perform worse.

source/s3_request_messages.c
```diff
@@ -918,6 +918,42 @@ struct aws_http_message *aws_s3_message_util_copy_http_message_no_body_filter_he
...
+    struct aws_allocator *allocator,
+    struct aws_http_message *base_message,
+    struct aws_byte_cursor filepath) {
+
+    bool success = false;
+    struct aws_string *filepath_str = NULL;
+    struct aws_input_stream *body_stream = NULL;
+    struct aws_http_message *message = NULL;
+
+    message = aws_s3_message_util_copy_http_message_no_body_filter_headers(allocator, base_message, NULL, 0, false);
```
TingDaoK, line 932: trivial: 
```suggestion
    /* Copy message and retain all headers */
    message = aws_s3_message_util_copy_http_message_no_body_filter_headers(allocator, base_message, NULL/*excluded_header_array*/, 0/*excluded_header_array_size*/, false/*exclude_x_amz_meta*/);
```

TingDaoK (approved):
LGTM

PR #268, opened 2023-03-16
Remove port part while comparing the host in s_apply_endpoint_override

shtripat:
*Issue #, if available:*

*Description of changes:*
This is in context with project mountpoint-s3 where other object storage systems like MinIO could be used in backend. In dev mode the `--endpoint-url` value passed could be of form `host[:port]` and during the comparison for checking the host value, we should chuck off the port part to make sure we dont hit the error
`Cannot create meta s3 request; host header value HOST does not match endpoint override HOST:PORT`

### DIFF
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -678,7 +678,12 @@ int s_apply_endpoint_override(
     struct aws_byte_cursor host_value;
     AWS_FATAL_ASSERT(aws_http_headers_get(message_headers, g_host_header_name, &host_value) == AWS_OP_SUCCESS);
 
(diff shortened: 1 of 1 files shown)


jamesbornholt:
Thanks for catching this! I was chatting to @DmitriyMusatkin today about it. I think we should probably fix this on the Mountpoint side rather than changing the CRT—we already have the port number since we're setting it in the `endpoint`, so we should just include it in the `Host` header too. I opened https://github.com/awslabs/mountpoint-s3/issues/159 to track that.

shtripat:
Closing as it would be taken care at mountpoint-s3 client side by PR https://github.com/awslabs/mountpoint-s3/pull/160

PR #271, opened 2023-03-29
telemetry API

TingDaoK:
- Support get the telemetry after each single request


include/aws/s3/s3_client.h
```diff
@@ -233,6 +283,15 @@ struct aws_s3_client_config {
...
     void *shutdown_callback_user_data;
 
+    /**
+     * Optional.
+     * To get telemetry metrics when a single request finishes.
+     * If set the request will keep track the metrics from `aws_s3_request_metrics`, and fire the callback when the
+     * request finishes receiving response.
+     */
+    aws_s3_meta_request_telemetry_fn *telemetry_callback;
+    void *telemetry_callback_user_data;
```
waahm7, line 293: I think it will be multiple callbacks like part start , part end etc. `telemetry_callback` is not a good name, maybe something like `on_requst_finished_callback`

include/aws/s3/s3_client.h
```diff
@@ -142,6 +142,56 @@ typedef void(aws_s3_meta_request_shutdown_fn)(void *user_data);
...
+    /* X-AMZ-REQUEST-ID Header value. */
+    struct aws_byte_cursor request_id;
+    /* The IP address of the request connected to */
+    struct aws_byte_cursor ip_address;
+
+    struct {
+        /* The total time for the single request started sending to finish (response received or failed). */
+        uint64_t total_duration_time_ns;
+        /* The time for the request started sending to start receiving response. */
+        uint64_t response_latency_time_ns;
```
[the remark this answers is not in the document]
graebm: also, we need some way to indicate "data not available", but 0 could be a real value for the durations...

we could either say "0 for timestamps means it didn't happen"
or use -1 if we go with `int64_t`

include/aws/s3/s3_client.h
```diff
@@ -142,6 +142,56 @@ typedef void(aws_s3_meta_request_shutdown_fn)(void *user_data);
...
+        uint32_t response_status;
+        /* HTTP Headers of the response received. */
+        struct aws_http_headers *response_headers;
+        /* The request line: method, path and query of the request.  */
+        struct aws_byte_cursor request_line;
+    } req_resp_info_metrics;
+
+    struct {
+        /* The pointer ID to the connection that request was made from */
+        size_t connection_id;
```
graebm, line 178: also useful: whether this was the 1st or 40th request on a given connection. maybe named: connection_request_count

include/aws/s3/private/s3_request.h
```diff
@@ -23,6 +24,62 @@ enum aws_s3_request_flags {
...
 
+/**
+ * Information sent in the meta_request telemetry callback.
+ */
+struct aws_s3_request_metrics {
+    struct aws_allocator *allocator;
+
+    struct {
+        /* The time stamp when the request started by S3 client, which is prepared time by the client */
+        uint64_t start_timestamp_ns;
```
JonathanHenson, line 35: Is it a time stamp? That implies I can convert this number to a calendar date and time. Is that the case or is it from the monotonic clock?

TingDaoK: They are from the monotonic clock, I'll add document about it, thanks!

include/aws/s3/private/s3_util.h
```diff
@@ -144,6 +144,8 @@ AWS_S3_API
 extern const struct aws_byte_cursor g_range_header_name;
 
 extern const struct aws_byte_cursor g_if_match_header_name;
+AWS_S3_API
+extern const struct aws_byte_cursor g_request_id_header_name;
```
DmitriyMusatkin, line 148: do we really want to expose symbols for this header value?
if we need those headers in test should we just make them constant exposed in the header instead of externs?

include/aws/s3/s3_client.h
```diff
@@ -653,12 +675,119 @@ struct aws_s3_meta_request *aws_s3_meta_request_acquire(struct aws_s3_meta_reque
...
+ * Release a reference.
+ * When the reference count drops to 0, this object will be cleaned up.
+ * It's OK to pass in NULL (nothing happens).
+ * Always returns NULL.
+ */
+AWS_S3_API
+struct aws_s3_request_metrics *aws_s3_request_metrics_release(struct aws_s3_request_metrics *metrics);
+
+/*************************************  Getters for s3 request metrics ************************************************/
+/* Ge the request ID from aws_s3_request_metrics, AWS_ERROR_S3_METRIC_DATA_NOT_AVAILABLE will be raised when data
```
DmitriyMusatkin, line 705: nit: typo

include/aws/s3/s3_client.h
```diff
@@ -653,12 +675,119 @@ struct aws_s3_meta_request *aws_s3_meta_request_acquire(struct aws_s3_meta_reque
...
+ * unavailable */
+AWS_S3_API
+int aws_s3_request_metrics_get_request_id(
+    const struct aws_s3_request_metrics *metrics,
+    struct aws_byte_cursor *request_id);
```
DmitriyMusatkin, line 710: should we explicitly name the out variables? ex out_request_id

include/aws/s3/private/s3_request.h
```diff
@@ -23,6 +24,69 @@ enum aws_s3_request_flags {
...
+        uint64_t receiving_duration_ns;
+    } time_metrics;
+
+    struct {
+        /* Response status code for the request */
+        int response_status;
+        /* HTTP Headers of the response received. */
+        struct aws_http_headers *response_headers;
+        /* The request line: method, path and query of the request. */
+        struct aws_byte_buf request_path_query;
```
DmitriyMusatkin, line 71: should we add the whole url? with virtual host addressing, path will not have bucket

include/aws/s3/private/s3_request.h
```diff
@@ -23,6 +24,69 @@ enum aws_s3_request_flags {
...
+        struct aws_byte_buf ip_address;
+        /* The pointer to the connection that request was made from */
+        void *connection_id;
+        /* The aws_thread_id_t to the thread that request ran on */
+        aws_thread_id_t thread_id;
+        /* The stream-id, which is the idex when the stream was activated. */
+        size_t stream_id;
+        /* CRT error code when the aws_s3_request finishes. */
+        int error_code;
+    } crt_info_metrics;
```
DmitriyMusatkin, line 85: from the doc in the header its confusing what those metrics are? do we send them per request?

include/aws/s3/private/s3_request.h
```diff
@@ -23,6 +24,69 @@ enum aws_s3_request_flags {
     AWS_S3_REQUEST_FLAG_ALWAYS_SEND = 0x00000004,
 };
 
+/**
+ * Information sent in the meta_request telemetry callback.
+ */
+struct aws_s3_request_metrics {
```
DmitriyMusatkin, line 30: should we expose extra things like num_parts, bytes_transferred?

include/aws/s3/private/s3_request.h
```diff
@@ -23,6 +24,69 @@ enum aws_s3_request_flags {
...
+        struct aws_http_headers *response_headers;
+        /* The request line: method, path and query of the request. */
+        struct aws_byte_buf request_path_query;
+    } req_resp_info_metrics;
+
+    struct {
+        /* The IP address of the request connected to */
+        struct aws_byte_buf ip_address;
+        /* The pointer to the connection that request was made from */
+        void *connection_id;
```
DmitriyMusatkin, line 78: exposing void * is just leaking implementation detail that id is just pointer. sjold we do size_t for it instead or something else?

include/aws/s3/private/s3_request.h
TingDaoK: I don't get why it's confusing. You mean the intern structures?

include/aws/s3/private/s3_request.h
TingDaoK: I am not sure, the virtual host address seems similar to the ip address, and bucket is kind provided by user?

include/aws/s3/private/s3_request.h
DmitriyMusatkin: maybe i dont understand what `request_path_query` is, but how would it include num parts?

include/aws/s3/private/s3_request.h
DmitriyMusatkin: doc mentions that this is meta_request level metrics, but some of the metrics seem to be at the request level (ex. ip_address).

include/aws/s3/private/s3_request.h
DmitriyMusatkin: virtual host address is just bucket being subdomain. bucket is provided by user, but then user needs to do mapping to figure out complete endpoint this request is addressing. adding whole url would just make it simpler for user which request metric is associated with

include/aws/s3/s3_client.h
```diff
@@ -404,6 +415,17 @@ struct aws_s3_meta_request_options {
...
+    /**
+     * Optional.
+     * To get telemetry metrics when a single request finishes.
+     * If set the request will keep track the metrics from `aws_s3_request_metrics`, and fire the callback when the
+     * request finishes receiving response.
+     *
+     * Note: the callback will be invoked multiple times from different thread.
+     * TODO: do we want a separate user data for telemetry callback?
+     */
+    aws_s3_meta_request_telemetry_fn *telemetry_callback;
```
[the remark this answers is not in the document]
waahm7: I think we should also mention that the data is not valid beyond the lifetime of the callback. They should acquire/release a reference if required.

include/aws/s3/s3_client.h
```diff
@@ -653,12 +675,121 @@ struct aws_s3_meta_request *aws_s3_meta_request_acquire(struct aws_s3_meta_reque
...
+
+/* Ge the start time stamp when request was sent to the network channel from aws_s3_request_metrics.
+ * AWS_ERROR_S3_METRIC_DATA_NOT_AVAILABLE will be raised if the request was end before it gets sent. */
+AWS_S3_API
+int aws_s3_request_metrics_get_send_start_timestamp_ns(
+    const struct aws_s3_request_metrics *metrics,
+    uint64_t *out_send_start_time);
+
+AWS_S3_API
+int aws_s3_request_metrics_get_send_end_timestamp_ns(
```
waahm7, line 737: I think we should add docs for all the functions and mention if they will be always available or not. Since the actual struct is now private, we should copy paste the documentation here as well.

include/aws/s3/private/s3_request.h
```diff
@@ -23,6 +24,69 @@ enum aws_s3_request_flags {
...
+
+        /* The time stamp when the request started to be encoded. 0 means data not available. Timestamp
+         * are from `aws_high_res_clock_get_ticks` */
+        uint64_t send_start_timestamp_ns;
+        /* The time stamp when the request finished to be encoded. 0 means data not available.
+         * Timestamp are from `aws_high_res_clock_get_ticks` */
+        uint64_t send_end_timestamp_ns;
+        /* The time duration for the request from start encoding to finish encoding (send_end_timestamp_ns -
+         * send_start_timestamp_ns). When send_end_timestamp_ns is 0, means data not available. */
+        uint64_t sending_duration_ns;
```
waahm7, line 52: Did we decide on https://github.com/awslabs/aws-c-s3/pull/271#discussion_r1152314703? I think it would be nice to be consistent with http layer.

source/s3_request.c
```diff
@@ -59,6 +75,19 @@ void aws_s3_request_clean_up_send_data(struct aws_s3_request *request) {
...
     aws_signable_destroy(request->send_data.signable);
     request->send_data.signable = NULL;
+    if (request->send_data.metrics) {
+        /* invoke callback */
+        struct aws_s3_meta_request *meta_request = request->meta_request;
+        struct aws_s3_request_metrics *metric = request->send_data.metrics;
+        aws_high_res_clock_get_ticks(&metric->time_metrics.end_timestamp_ns);
+        metric->time_metrics.total_duration_ns =
+            metric->time_metrics.end_timestamp_ns - metric->time_metrics.start_timestamp_ns;
+        /* End the timestamp */
```
waahm7, line 85: I am confused about this comment. Is this misplaced?

source/s3_request.c
```diff
@@ -59,6 +75,19 @@ void aws_s3_request_clean_up_send_data(struct aws_s3_request *request) {
...
+        if (meta_request->telemetry_callback) {
+            meta_request->telemetry_callback(meta_request, metric, meta_request->user_data);
```
waahm7, line 87: Is this the right place to trigger this callback? We also call this function in `aws_s3_request_setup_send_data`

TingDaoK: I think yes.

https://github.com/awslabs/aws-c-s3/blob/main/include/aws/s3/private/s3_request.h#L84

The send_data should be around for each time the request is sent, so, if any error or retries happens for the aws_s3_request, we will still get the metrics for that time the real http request sent/to be sent.

include/aws/s3/private/s3_request.h
TingDaoK: For put parts, it will have the part number in it, but, for multipart get, it is not.

I'll add the part number to be easier for the user

include/aws/s3/private/s3_request.h
TingDaoK: In my opinion, the reason we used int64_t to have the data unavailable easy to check, but we provide the getters with error, so I guess it will make more sense for those value to be unsigned, as they should never less than zero.

include/aws/s3/private/s3_request.h
TingDaoK: So, should it be the host header of the request? I can add that

include/aws/s3/private/s3_request.h
TingDaoK: But, for multipart get, we were actually doing ranged-get, which is different from get parts. I guess it can be confused to the user as well.

include/aws/s3/private/s3_request.h
TingDaoK: changed the getter to return a size_t

source/s3_meta_request.c
```diff
@@ -974,6 +984,21 @@ static int s_s3_meta_request_incoming_headers(
...
+        if (s3_metrics->req_resp_info_metrics.response_headers == NULL) {
+            s3_metrics->req_resp_info_metrics.response_headers = aws_http_headers_new(meta_request->allocator);
+        }
+
+        for (size_t i = 0; i < headers_count; ++i) {
+            const struct aws_byte_cursor *name = &headers[i].name;
+            const struct aws_byte_cursor *value = &headers[i].value;
+
+            aws_http_headers_add(s3_metrics->req_resp_info_metrics.response_headers, *name, *value);
+        }
```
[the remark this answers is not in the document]
TingDaoK: oh, the headers from callback will be invoked repeatedly here and is only part of the whole headers of the response. So, it's better to create our own copy. And we did the same thing couple line down there.

waahm7 (approved):

include/aws/s3/s3_client.h
```diff
@@ -661,12 +686,130 @@ struct aws_s3_meta_request *aws_s3_meta_request_acquire(struct aws_s3_meta_reque
...
+
+/* Get the start time stamp when request was sent to the network channel from aws_s3_request_metrics.
+ * AWS_ERROR_S3_METRIC_DATA_NOT_AVAILABLE will be raised if the request ended before it gets sent. */
+AWS_S3_API
+int aws_s3_request_metrics_get_send_start_timestamp_ns(
+    const struct aws_s3_request_metrics *metrics,
+    uint64_t *out_send_start_time);
+
+AWS_S3_API
+int aws_s3_request_metrics_get_send_end_timestamp_ns(
```
graebm, line 748: trivial: comment each getter, similar to how the underlying struct members are commented.
If there's some boilerplate that applies to every single getter (like how they're using the high resolution clock) you could just have that once, like on the forward declaration of `struct aws_s3_request_metrics`. Ultimately that will go on the class declaration when we bind it to Java

include/aws/s3/s3_client.h
```diff
@@ -661,12 +686,130 @@ struct aws_s3_meta_request *aws_s3_meta_request_acquire(struct aws_s3_meta_reque
...
+    const struct aws_s3_request_metrics *metrics,
+    uint64_t *out_receive_end_time);
+
+AWS_S3_API
+int aws_s3_request_metrics_get_receiving_duration_ns(
+    const struct aws_s3_request_metrics *metrics,
+    uint64_t *out_receiving_duration);
+
+AWS_S3_API
+int aws_s3_request_metrics_get_response_status(const struct aws_s3_request_metrics *metrics, int *out_response_status);
```
graebm, line 773: trivial/debatable: ~status~ -> `status_code`

"status" is just so vague

include/aws/s3/s3_client.h
```diff
@@ -661,12 +686,130 @@ struct aws_s3_meta_request *aws_s3_meta_request_acquire(struct aws_s3_meta_reque
...
+
+AWS_S3_API
+int aws_s3_request_metrics_get_connection_id(const struct aws_s3_request_metrics *metrics, size_t *out_connection_id);
+
+/* Get the thread ID of the thread that request was made from */
+AWS_S3_API
+int aws_s3_request_metrics_get_thread_id(const struct aws_s3_request_metrics *metrics, aws_thread_id_t *out_thread_id);
+
+AWS_S3_API
+int aws_s3_request_metrics_get_request_stream_id(const struct aws_s3_request_metrics *metrics, size_t *out_stream_id);
```
graebm, line 807: shouldn't stream_id be uint32_t?

include/aws/s3/s3_client.h
```diff
@@ -661,12 +686,130 @@ struct aws_s3_meta_request *aws_s3_meta_request_acquire(struct aws_s3_meta_reque
...
+
+/* Get the AWS CRT error code from request metics. */
```
graebm, line 809:
```suggestion
/* Get the AWS CRT error code from request metrics. */
```

include/aws/s3/s3_client.h
```diff
@@ -661,12 +686,130 @@ struct aws_s3_meta_request *aws_s3_meta_request_acquire(struct aws_s3_meta_reque
...
+
+AWS_S3_API
+int aws_s3_request_metrics_get_response_headers(
+    const struct aws_s3_request_metrics *metrics,
+    struct aws_http_headers **out_response_headers);
+
+AWS_S3_API
+void aws_s3_request_metrics_get_request_path_query(
+    const struct aws_s3_request_metrics *metrics,
+    struct aws_byte_cursor *out_request_path_query);
```
graebm, line 783: trivial: I would use aws_string for these string properties, instead of aws_byte_cursor/aws_byte_buf

source/s3_request.c
```diff
@@ -95,3 +124,259 @@ static void s_s3_request_destroy(void *user_data) {
...
+
+    (void)err;
+    aws_ref_count_init(&metrics->ref_count, metrics, s_s3_request_metrics_destroy);
+
+    return metrics;
+}
+struct aws_s3_request_metrics *aws_s3_request_metrics_acquire(struct aws_s3_request_metrics *metrics) {
+    if (!metrics) {
+        return NULL;
+    }
```
graebm, line 173:
```suggestion
    AWS_PRECONDITION(metrics);
```

only destroy() and release() functions should tolerate NULL

source/s3_request.c
```diff
@@ -95,3 +124,259 @@ static void s_s3_request_destroy(void *user_data) {
...
+
+int aws_s3_request_metrics_get_receiving_duration_ns(
+    const struct aws_s3_request_metrics *metrics,
+    uint64_t *receiving_duration) {
+    AWS_PRECONDITION(metrics);
+    AWS_PRECONDITION(receiving_duration);
+    if (metrics->time_metrics.receive_end_timestamp_ns == 0 || metrics->time_metrics.receive_start_timestamp_ns == 0) {
+        return aws_raise_error(AWS_ERROR_S3_METRIC_DATA_NOT_AVAILABLE);
+    }
+    *receiving_duration = metrics->time_metrics.receiving_duration_ns;
```
graebm, line 284: wouldn't it be less error_prone to continue storing times as int64_t internally, so we could simply use -1 to indicate date not available?
```suggestion
    if (metrics->time_metrics.receiving_duration_ns < 0) {
        return aws_raise_error(AWS_ERROR_S3_METRIC_DATA_NOT_AVAILABLE);
    }
    *receiving_duration = (uint64_t)metrics->time_metrics.receiving_duration_ns;
```

vs doing this thing where 0 maaay be valid from some values, so we better consult two other values

like we were doing in that "Range: 0-0" bug we just had

source/s3_request.c
waahm7: I thought acquire also treats NULL https://github.com/awslabs/aws-c-http/pull/359#discussion_r789280477.

graebm: Huh, I was unaware of this

a quick ctrl-shift-f shows that acquire() functions' tolerance for NULL is wildly inconsistent throughout the codebase :|

whatever, sure, tolerate NULL I could see it both ways
PRO: consistent with release()
CON: the more functions you have that tolerate NULL, the sketchier your codebase gets, as an unexpected NULL gets blows up much later than it might have otherwise

don't super care to spend the time arguing

include/aws/s3/s3_client.h
TingDaoK: Why? I think we have the metrics as the object holding all the data, so this is the getter to fetch the data from the object, will it be better to use `aws_byte_cursor`?

graebm: aws_string is null-terminated, which makes is easier if the user wants to pass it along to things like printf() or other old-fashioned C calls that expect a null-terminated string

waahm7: Additionally, it is easier to reason about the lifetime of aws_string. With a byte_cursor, users must also ensure that the underlying memory is not freed until the lifetime of the byte_cursor.

graebm: I was still proposing the same sketchy lifetime as we'd get with aws_byte_cursor, where we just give them a pointer to the internal data and warn them about lifetimes

source/s3_request.c
```diff
@@ -95,3 +124,285 @@ static void s_s3_request_destroy(void *user_data) {
...
+    *send_end_time = metrics->time_metrics.send_end_timestamp_ns;
+    return AWS_OP_SUCCESS;
+}
+
+int aws_s3_request_metrics_get_sending_duration_ns(
+    const struct aws_s3_request_metrics *metrics,
+    uint64_t *sending_duration) {
+    AWS_PRECONDITION(metrics);
+    AWS_PRECONDITION(sending_duration);
+    if (metrics->time_metrics.send_end_timestamp_ns < 0 || metrics->time_metrics.send_start_timestamp_ns < 0) {
```
graebm, line 263:
```suggestion
    if (metrics->time_metrics.sending_duration_ns < 0) {
```

source/s3_request.c
```diff
@@ -95,3 +124,285 @@ static void s_s3_request_destroy(void *user_data) {
...
+    *receive_end_time = metrics->time_metrics.receive_end_timestamp_ns;
+    return AWS_OP_SUCCESS;
+}
+
+int aws_s3_request_metrics_get_receiving_duration_ns(
+    const struct aws_s3_request_metrics *metrics,
+    uint64_t *receiving_duration) {
+    AWS_PRECONDITION(metrics);
+    AWS_PRECONDITION(receiving_duration);
+    if (metrics->time_metrics.receive_end_timestamp_ns < 0 || metrics->time_metrics.receive_start_timestamp_ns < 0) {
```
graebm, line 299:
```suggestion
    if (metrics->time_metrics.receiving_duration_ns < 0) {
```

source/s3_meta_request.c
```diff
@@ -1060,6 +1085,40 @@ static int s_s3_meta_request_incoming_body(
...
+
+    s3_metrics->crt_info_metrics.stream_id = http_metrics->stream_id;
+
+    /* Also related metrics from the request/response. */
+    s3_metrics->crt_info_metrics.connection_id = (void *)connection->http_connection;
+    const struct aws_socket_endpoint *endpoint = aws_http_connection_get_remote_endpoint(connection->http_connection);
+    struct aws_byte_cursor remote_address = aws_byte_cursor_from_c_str(endpoint->address);
+
+    request->send_data.metrics->crt_info_metrics.ip_address =
+        aws_string_new_from_cursor(request->allocator, &remote_address);
```
graebm, line 1116: trivial:
```suggestion
    request->send_data.metrics->crt_info_metrics.ip_address =
        aws_string_new_from_c_str(request->allocator, endpoint->address);
```

include/aws/s3/s3_client.h
```diff
@@ -661,12 +686,172 @@ struct aws_s3_meta_request *aws_s3_meta_request_acquire(struct aws_s3_meta_reque
...
+
+/*************************************  Getters for s3 request metrics ************************************************/
+/**
+ * Get the request ID from aws_s3_request_metrics.
+ * If unavailable, AWS_ERROR_S3_METRIC_DATA_NOT_AVAILABLE will be raised.
+ **/
+AWS_S3_API
+int aws_s3_request_metrics_get_request_id(
+    const struct aws_s3_request_metrics *metrics,
+    struct aws_byte_cursor *out_request_id);
```
graebm, line 723: make this a string like the other strings

also, warn about its lifetime

graebm (approved):
fix & ship

include/aws/s3/s3_client.h
```diff
@@ -412,6 +425,18 @@ struct aws_s3_meta_request_options {
...
 
+    /**
+     * Optional.
+     * To get telemetry metrics when a single request finishes.
+     * If set the request will keep track of the metrics from `aws_s3_request_metrics`, and fire the callback when the
+     * request finishes receiving response.
+     * See `aws_s3_meta_request_telemetry_fn`
+     *
+     * Notes:
+     * - The callback will be invoked multiple times from different threads.
```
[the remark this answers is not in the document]
TingDaoK: yes

include/aws/s3/s3_client.h
```diff
@@ -412,6 +425,18 @@ struct aws_s3_meta_request_options {
      */
     aws_s3_meta_request_progress_fn *progress_callback;
 
+    /**
+     * Optional.
+     * To get telemetry metrics when a single request finishes.
```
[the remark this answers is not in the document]
TingDaoK: yes

include/aws/s3/s3_client.h
```diff
@@ -661,12 +686,174 @@ struct aws_s3_meta_request *aws_s3_meta_request_acquire(struct aws_s3_meta_reque
...
+/**
+ * Get the path and query of the request.
+ * If unavailable, AWS_ERROR_S3_METRIC_DATA_NOT_AVAILABLE will be raised.
+ * If available, out_request_path_query will be set to a string. Be warned this string's lifetime is tied to the metrics
+ * object.
+ */
+AWS_S3_API
+void aws_s3_request_metrics_get_request_path_query(
+    const struct aws_s3_request_metrics *metrics,
+    const struct aws_string **out_request_path_query);
```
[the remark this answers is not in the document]
TingDaoK: I'll add the type of request in a follow up PR to be more clear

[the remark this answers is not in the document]
TingDaoK: https://github.com/awslabs/aws-c-s3/pull/287

PR #276, opened 2023-04-24
add multipart upload threshold

TingDaoK:
- add multipart upload threshold configure
- API-break change: Update the part size public interface from `size_t` to `uint64_t`


include/aws/s3/private/s3_client_impl.h
```diff
@@ -192,6 +192,12 @@ struct aws_s3_client {
      * to meta requests for use. */
     const size_t max_part_size;
 
+    /* The size threshold in bytes for when to use multipart uploads for a AWS_S3_META_REQUEST_TYPE_PUT_OBJECT meta
+     * request. Uploads over this size will automatically use a multipart upload strategy, while uploads smaller than
+     * this threshold will use a single connection to upload the whole object. If not set, `part_size` will be used as
+     * threshold. */
+    const uint64_t multipart_upload_threshold;
```
DmitriyMusatkin, line 199: if part sizes are size_t, shouldnt this also be size_t?

source/s3_client.c
```diff
@@ -305,6 +305,10 @@ struct aws_s3_client *aws_s3_client_new(
         *((size_t *)&client->max_part_size) = (size_t)s_default_max_part_size;
     }
 
+    if (client_config->multipart_upload_threshold != 0) {
+        *((uint64_t *)&client->multipart_upload_threshold) = client_config->multipart_upload_threshold;
```
DmitriyMusatkin, line 309: i know we are doing it in other cases, but do we really need all the casting magic here?

source/s3_client.c
```diff
@@ -1009,7 +1013,9 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
 
                     client_max_part_size = g_s3_min_upload_part_size;
                 }
-                if (content_length <= client_part_size) {
+                uint64_t multipart_upload_threshold =
```
DmitriyMusatkin, line 1016: corner case: what if multipart threshold is less than a part? should it fail or send 1 part?

include/aws/s3/private/s3_client_impl.h
```diff
@@ -192,6 +192,12 @@ struct aws_s3_client {
      * to meta requests for use. */
     const size_t max_part_size;
 
+    /* The size threshold in bytes for when to use multipart uploads for a AWS_S3_META_REQUEST_TYPE_PUT_OBJECT meta
```
DmitriyMusatkin, line 195: nit: can we add info on what happens if content length exactly equals threshold

source/s3_client.c
TingDaoK: send one part of multipart upload as Java SDK did

include/aws/s3/private/s3_client_impl.h
DmitriyMusatkin: part size and threshold were long in java v1 tm? and looks like proposal is to have them both be long going forward. its probably better to have the same type for all part related config options, and java can cast if really needed

DmitriyMusatkin (approved):

DmitriyMusatkin (approved):

PR #285, opened 2023-05-10
Support unknown content-length for put

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Support unknown content length for put


source/s3_auto_ranged_put.c
```diff
@@ -1102,61 +1189,71 @@ static void s_s3_auto_ranged_put_request_finished(
...
-                meta_request->progress_callback(meta_request, &progress, meta_request->user_data);
-            }
+
             /* BEGIN CRITICAL SECTION */
             {
                 aws_s3_meta_request_lock_synced_data(meta_request);
 
                 ++auto_ranged_put->synced_data.num_parts_completed;
 
+                if (request_is_noop) {
```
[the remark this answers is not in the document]
DmitriyMusatkin: checked the code. metrics will not be collected

source/s3_auto_ranged_put.c
```diff
@@ -447,10 +502,19 @@ static bool s_s3_auto_ranged_put_update(
...
+            if (auto_ranged_put->has_content_length) {
+                if (auto_ranged_put->synced_data.num_parts_completed != auto_ranged_put->synced_data.total_num_parts) {
+                    goto has_work_remaining;
+                }
+            } else {
+                if ((!auto_ranged_put->prepare_data.is_body_stream_at_end) ||
+                    auto_ranged_put->synced_data.num_parts_completed != auto_ranged_put->synced_data.num_parts_sent) {
+                    goto has_work_remaining;
+                }
             }
```
graebm, line 518: maybe continue using `total_num_parts` but add a bool whether or not total_num_parts is final

```
        /* If content length is not known, this becomes true after we read EOF from the body stream */
        bool total_num_parts_known;
```

I'm not really sure if this simplifies or confuses things, just throwing the idea out there

include/aws/s3/private/s3_auto_ranged_put.h
```diff
@@ -69,11 +75,18 @@ struct aws_s3_auto_ranged_put {
         struct aws_s3_paginated_operation *list_parts_operation;
         struct aws_string *list_parts_continuation_token;
 
+        /* Note: total num parts is known only if content-length is known,
+        otherwise it will be 0. */
         uint32_t total_num_parts;
         uint32_t num_parts_sent;
+        uint32_t num_parts_prepared;
```
graebm, line 82: "prepared" is a bit ambiguous. We increment this in the middle of the "prepare request" pipeline. There's still more work before we're done preparing, like async signing. I'd expect something like this to be incremented after the "prepare request" pipeline is finished, right before the callback fires

Also, an aws_s3_request can be prepared multiple times, in the case of retry (we skip reading if its not the first time we've prepared it)

maybe?
```suggestion
        /* Number of "sent" parts we've finished reading the body for
         * (does not include skipped parts in the case of pause/resume) */
        uint32_t num_parts_read;
```

graebm (commented):
not done reviewing ... but heading to lunch

source/s3_client.c
```diff
@@ -1028,63 +1026,75 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
...
-                if ((content_length % part_size) > 0) {
-                    ++num_parts;
+                    if ((content_length % part_size) > 0) {
+                        ++num_parts;
+                    }
                 }
 
                 return aws_s3_meta_request_auto_ranged_put_new(
                     client->allocator, client, part_size, content_length, num_parts, options);
             } else {
```
graebm, line 1088: trivial: the if statement was like 100 lines ago
```suggestion
            } else { /* else using resume token */
```

source/s3_client.c
```diff
@@ -1028,63 +1026,75 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
...
-                /* dont pass part size and total num parts. constructor will pick it up from token */
+                if (!content_length_found) {
+                    AWS_LOGF_ERROR(
+                        AWS_LS_S3_META_REQUEST,
+                        "Could not create auto-ranged-put meta request; content_length must be specified.");
```
graebm, line 1092: trivial: mention "resume" somewhere
"Content-Length must be specified for resume"
??

graebm (commented):
not 100% finished, but more comments

source/s3_auto_ranged_put.c
```diff
@@ -15,6 +15,18 @@
 static const struct aws_byte_cursor s_upload_id = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("UploadId");
 static const size_t s_complete_multipart_upload_init_body_size_bytes = 512;
 static const size_t s_abort_multipart_upload_init_body_size_bytes = 512;
+/* For unknown length body we no longer know the number of parts. to avoid
+ * resizing arrays for etags/checksums too much, those array start out with
+ * capacity specified by the constant bellow. Note: constant value is an
+ */
+static const uint32_t s_unknown_length_default_num_parts = 256;
```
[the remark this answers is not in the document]
DmitriyMusatkin: im fine with 32 as well and we can always tweak it later.

include/aws/s3/s3_client.h
```diff
@@ -52,6 +52,16 @@ enum aws_s3_meta_request_type {
...
      * The PutObject request will be split into MultiPart uploads that are executed in parallel
      * to improve throughput, when possible.
+     * Note: put object supports both known and unknown body length. The client
+     * relies on Content-Length header to determine length of the body.
+     * Request with unknown content length are always sent using multipart
+     * upload regardless of final number of parts and do have the following limitations:
+     * - multipart threshold is ignored and all request are made through mpu,
+     *   even if they only need one part
+     * - pause/resume is not supported
+     * - meta request will throw error if checksum header is provider (due to
```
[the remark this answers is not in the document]
DmitriyMusatkin: discussed offline and we will backlog it

include/aws/s3/private/s3_request.h
```diff
@@ -190,6 +190,16 @@ struct aws_s3_request {
...
+
+    /* When true, this request does not represent a useful http request and
+     * must not be send, but should still go through the rest of the flow. Those
+     * requests can occur when request is optimistically created during update,
+     * but cannot be prepared. ex. when put has no content length, requests will
+     * be scheduled as regular to ensure fair distribution against other
+     * requests, but can also result in requests for uploading data after the end
+     * of the stream (those requests will use below flag to indicate that they
+     * should not be sent). */
+    uint32_t is_noop : 1;
```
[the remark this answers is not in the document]
DmitriyMusatkin: discussed offline. yes its possible and ill go a bit more in depth in the other related comment on that.

source/s3_auto_ranged_put.c
```diff
@@ -73,11 +85,25 @@ static bool s_process_part_info(const struct aws_s3_part_info *info, void *user_
...
+     * element make sure to init any new elements to zeroed values. */
+    size_t current_num_parts = aws_array_list_length(&auto_ranged_put->synced_data.etag_list);
+    if (info->part_number > current_num_parts) {
+        struct aws_byte_buf empty_buf = {0};
+        struct aws_string *null_etag = NULL;
+
+        for (size_t i = info->part_number; i > current_num_parts; --i) {
+            aws_array_list_set_at(&auto_ranged_put->encoded_checksum_list, &empty_buf, i - 1);
+            aws_array_list_set_at(&auto_ranged_put->synced_data.etag_list, &null_etag, i - 1);
+        }
```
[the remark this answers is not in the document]
DmitriyMusatkin: its actually on purpose to avoid underflowing iterator variable in cases where first part was not completed before (i.e. case where array is empty and we receive data for any part other than 0).

source/s3_auto_ranged_put.c
```diff
@@ -800,23 +866,71 @@ static int s_s3_auto_ranged_put_prepare_request(
...
+                        auto_ranged_put->prepare_data.is_body_stream_at_end =
+                            aws_s3_meta_request_body_has_no_more_data(meta_request);
+
+                        struct aws_byte_buf checksum_buf = {0};
+                        aws_array_list_set_at(
+                            &auto_ranged_put->encoded_checksum_list, &checksum_buf, request->part_number - 1);
+
+                        struct aws_string *null_etag = NULL;
+                        aws_array_list_set_at(
+                            &auto_ranged_put->synced_data.etag_list, &null_etag, request->part_number - 1);
```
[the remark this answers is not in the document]
DmitriyMusatkin: added comment. other code sets the etag/checksum based on the data returned from s3 for pause/resume flows, and there we can have gaps in completed parts. here its just zeroing out values for the new parts as they are prepared to avoid any garbage values during error flows

source/s3_auto_ranged_put.c
```diff
@@ -800,23 +866,71 @@ static int s_s3_auto_ranged_put_prepare_request(
...
+                                                               meta_request,
+                                                               auto_ranged_put->prepare_data.num_parts_read_from_stream,
+                                                               request->part_number - 1)) {
                     goto message_create_failed;
                 }
-                auto_ranged_put->prepare_data.num_parts_read_from_stream = request->part_number - 1;
 
-                aws_byte_buf_init(&request->request_body, meta_request->allocator, request_body_size);
+                if (aws_s3_meta_request_body_has_no_more_data(meta_request)) {
+                    request->is_noop = true;
```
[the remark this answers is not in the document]
DmitriyMusatkin: Yeah, current behavior of read is the expected limitation/bug of how we do streaming and arguably i dont think fixing it should be a part of this pr. This pr relies on existing assumption that clients should either fill up buffer to max or fill it as much as possible and indicate to us that eof has been reached. We can consider fixing that behavior in a separate pr, but i think it also becomes a moot point after async streaming is finished. 

Refactored this part a little bit to make logic clearer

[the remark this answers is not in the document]
graebm: Cool. Yeah, I'll fix it up to be efficient when merging this in the async-streaming work. No need to agonize here. It's all private functions

source/s3_auto_ranged_put.c
```diff
@@ -1102,61 +1259,71 @@ static void s_s3_auto_ranged_put_request_finished(
...
-                     * later, so just get rid of the quotes now. */
-                    etag = aws_strip_quotes(meta_request->allocator, etag_within_quotes);
+                        error_code = AWS_ERROR_S3_MISSING_ETAG;
+                    } else {
+                        /* The ETag value arrives in quotes, but we don't want it in quotes when we send it back up
+                         * later, so just get rid of the quotes now. */
+                        etag = aws_strip_quotes(meta_request->allocator, etag_within_quotes);
+                    }
+                }
+                if (error_code == AWS_ERROR_SUCCESS && meta_request->progress_callback != NULL) {
```
TingDaoK, line 1287: Oh, we need to update the documentation about progress callback. https://github.com/awslabs/aws-c-s3/blob/main/include/aws/s3/s3_client.h#L422-L428

It says it only supports for copyObject, that's not true.

source/s3_auto_ranged_put.c
```diff
@@ -800,23 +887,90 @@ static int s_s3_auto_ranged_put_prepare_request(
...
+                bool successfully_read_data_from_stream = false;
+                if (!aws_s3_meta_request_body_has_no_more_data(meta_request)) {
+                    aws_byte_buf_init(&request->request_body, meta_request->allocator, request_body_size);
+                    if (aws_s3_meta_request_read_body(meta_request, &request->request_body)) {
+                        goto message_create_failed;
+                    }
 
-                if (aws_s3_meta_request_read_body(meta_request, &request->request_body)) {
-                    goto message_create_failed;
+                    successfully_read_data_from_stream = request->request_body.len != 0;
```
TingDaoK, line 909: what if the input stream is not end, but just stalled (no data available at the moment, but probably will have more data later)?

source/s3_auto_ranged_put.c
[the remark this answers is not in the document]
graebm: I like what you did with it 👍

source/s3_auto_ranged_put.c
DmitriyMusatkin: discussed in one of the comments above. 
thats the current interface contract we have, user needs to always fill in the buffer we give them even if they dont have data readily available, forcing them to block, etc... 
I think we all agree that its a bad contract, but changing that contract is not part of this PR. Async streaming will resolve ehta

source/s3_auto_ranged_put.c
```diff
@@ -800,23 +866,71 @@ static int s_s3_auto_ranged_put_prepare_request(
...
+                        aws_s3_meta_request_lock_synced_data(meta_request);
+
+                        ++auto_ranged_put->synced_data.num_parts_prepared;
+
+                        auto_ranged_put->prepare_data.is_body_stream_at_end =
+                            aws_s3_meta_request_body_has_no_more_data(meta_request);
+
+                        struct aws_byte_buf checksum_buf = {0};
+                        aws_array_list_set_at(
+                            &auto_ranged_put->encoded_checksum_list, &checksum_buf, request->part_number - 1);
```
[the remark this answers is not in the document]
graebm: yeah, good move, taking it to `synced_data

it used to be a C array allocated once, so different threads were free to access their particular index

but now it's a dynamic array, so there's a risk that one thread resizes and messes up another thread

source/s3_auto_ranged_put.c
graebm: aws-c-s3 already suffers from that bug (the current code assumes that every attempt to read will completely fill the buffer)

it will be fixed when I do async-streaming

TingDaoK: Hmmm, but we are not error out or anything to enforce the assumption, it will continue and probably upload the file with skipping part number with current impl, right?

source/s3_auto_ranged_put.c
```diff
@@ -346,20 +351,38 @@ static void s_s3_meta_request_auto_ranged_put_destroy(struct aws_s3_meta_request
...
+static bool s_should_skip_scheduling_more_parts_based_on_flags(
+    const struct aws_s3_auto_ranged_put *auto_ranged_put,
+    uint32_t flags) {
+    if ((flags & AWS_S3_META_REQUEST_UPDATE_FLAG_CONSERVATIVE) != 0) {
+        uint32_t num_parts_in_flight =
+            (auto_ranged_put->synced_data.num_parts_sent - auto_ranged_put->synced_data.num_parts_completed);
+
+        /* Because uploads must read from their streams serially, we try to limit the amount of in flight
+         * requests for a given multipart upload if we can. */
+        return num_parts_in_flight > 0;
```
graebm, line 380: Isn't this going to limit us to only working on 1 part at a time? Like, we won't even start reading the 2nd part until the 1st part is 100% complete, having received an HTTP response?

That's going to be real slow. Once a part is done reading, we want to immediately start reading the next part, it's fine to have dozens of parts incomplete because they're still doing their HTTP transfer

source/s3_auto_ranged_put.c
```diff
@@ -1102,61 +1260,71 @@ static void s_s3_auto_ranged_put_request_finished(
...
+                if (request_is_noop) {
+                    ++auto_ranged_put->synced_data.num_parts_noop;
+                }
+
                 AWS_LOGF_DEBUG(
                     AWS_LS_S3_META_REQUEST,
                     "id=%p: %d out of %d parts have completed.",
                     (void *)meta_request,
                     auto_ranged_put->synced_data.num_parts_completed,
                     auto_ranged_put->synced_data.total_num_parts);
```
TingDaoK, line 1312: We only increase the `total_num_parts` when we successfully read data, which means if the request is_noop, the total num parts will NOT count it. But the `num_parts_completed` will still be increased for noop requests. That will be confusing.

source/s3_auto_ranged_put.c
TingDaoK: Oh, if this is an existing bug, then, yeah, sure. :live-with-it

source/s3_auto_ranged_put.c
DmitriyMusatkin: this is existing flow, i just pulled it into a separate function. 

our scheduler does 2 passes - first conservative and then regular. conservative pass always picks at most 1 request from each meta request to schedule. regular pass picks up as many requests as possible from meta request and then proceeds to next one

source/s3_auto_ranged_put.c
DmitriyMusatkin: fair call out. moved total num parts to always increment

source/s3_auto_ranged_put.c
DmitriyMusatkin: yeah, we currently dont enforce that assumption, which is a bit scary. we can probably put in a fatal if it has not read the whole body

tests/CMakeLists.txt
```diff
@@ -106,6 +106,8 @@ add_net_test_case(test_s3_upload_part_message_without_content_md5)
 add_net_test_case(test_s3_create_multipart_upload_message_with_content_md5)
 add_net_test_case(test_s3_complete_multipart_message_with_content_md5)
 add_net_test_case(test_s3_put_object_double_slashes)
+add_net_test_case(test_s3_put_object_no_content_length)
+add_net_test_case(test_s3_put_object_single_part_no_content_length)
```
graebm, line 110: make sure we have tests that:
- create a 0-size Object
- create a noop part, after creating 1+ normal parts

DmitriyMusatkin: test for 0 size object added. it uncovered some issues that were fixed.
existing tests already hit path for noop part, due to tests having at most 3 parts and optimistic scheduling was set at min of 5

source/s3_auto_ranged_put.c
```diff
@@ -447,10 +524,19 @@ static bool s_s3_auto_ranged_put_update(
...
             }
 
-            /* There is one more request to send after all of the parts (the complete-multipart-upload) but it can't be
-             * done until all of the parts have been completed.*/
-            if (auto_ranged_put->synced_data.num_parts_completed != auto_ranged_put->synced_data.total_num_parts) {
-                goto has_work_remaining;
+            /* There is one more request to send after all the parts (the complete-multipart-upload) but it can't be
+             * done until all the parts have been completed.
+             * TODO: slightly different handling right now between known and
+             * unknown content length. can this be unified?*/
```
waahm7, line 530: I can't think of something better, maybe we can just remove the todo?

source/s3_auto_ranged_put.c
```diff
@@ -988,6 +999,17 @@ static int s_s3_auto_ranged_put_prepare_request(
         }
         case AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_COMPLETE_MULTIPART_UPLOAD: {
 
+            if (!auto_ranged_put->has_content_length &&
+                auto_ranged_put->prepare_data.num_parts_read_from_stream == 0) {
+                AWS_LOGF_ERROR(
+                    AWS_LS_S3_META_REQUEST,
+                    "id=%p Could not create complete multipart upload for meta request with zero content length.",
```
graebm, line 1005: Error scolds the user for choosing to use Multipart Upload, but that's an implementation detail.

The user chose not to specify Content-Length, so let's phrase the error around that

Maybe something like:
> Set "Content-Length: 0" to upload an empty object. 0 byte meta requests lacking a Content-Length header are not currently supported.

graebm: Add add code comment about how CompleteMultipartUpload will error if there are no parts

We can ultimately support this when we shift to buffering the first part to determine whether MPU is necessary

waahm7 (approved):

graebm (approved):
Fix & Shipit

source/s3.c
```diff
@@ -37,7 +37,7 @@ static struct aws_error_info s_errors[] = {
     AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_OBJECT_MODIFIED, "The object modifed during download."),
     AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_NON_RECOVERABLE_ASYNC_ERROR, "Async error received from S3 and not recoverable from retry."),
     AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_METRIC_DATA_NOT_AVAILABLE, "The metric data is not available, the requests ends before the metric happens."),
-    AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_INVALID_CONTENT_LENGTH, "Request has unexpected content length.")
+    AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_INCORRECT_CONTENT_LENGTH, "Request body length must either match Content-Length header or be non-zero."),
```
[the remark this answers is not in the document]
DmitriyMusatkin: im fine with that. removed the new error i added and its just using unsupported operation now

djchapm:
Hey - been waiting a long time for something like this - thanks for implementing.  But experiencing performance problem - can you tell me what happens between the time I complete my request, and the AsyncUpload actually completes?  I'm using AWS SDK V2 S3 API to do Async upload with CRT Async Client - works without specifying upload size now due to this update, just there's a huge lag once I'm done.  Not sure where to go with this or who to ask.  Ends up being much slower start to finish than our old workaround for unknown content size.  Thanks!

graebm:
Please create an issue. This pull-request is a weird place to have a long debugging conversation

Please explain your workload:
- approximate size of objects you're uploading
- EC2 instance type
- what did you set target throughput to?
- did you customize part size?
- where is the data coming from? is it coming fast or slow? is it coming in a smooth stream, or coming in bursts from some other source?

graebm:
But to explain what's going on under the hood...

since we don't know how much data will be sent, we assume it will be large and treat it as a [Multipart Upload](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html). We send CreateMultipartUpload, then send an UploadPart for each 8MiB of data, then send CompleteMultipartUpload when it's done. If you're sending files smaller than 8MiB lower performance is expected. We have ideas on how to optimize this, but it adds a lot of complexity. It's something we hope to tackle in the future.

TingDaoK:
If it's the lag after the send has ended, I would check the integration between how you create the input and how it's passed from Java to C. As in C, we check the input_stream status to complete the MPU.

djchapm:
Thanks for the quick response - will create an issue and if you think it's unrelated to aws-c-s3 then we can ship it to s3-transfer-manager.

PR #286, opened 2023-05-11
Fix handling of content-encoding header for default request

waahm7:
*Issue #, if available:*
https://github.com/aws/aws-sdk-java-v2/issues/3750

*Description of changes:*
- If a content-encoding header is already present, append it to `aws-chunked` instead of replacing it completely.
- This worked for multipart upload request because we don't replace it for createMPU request, however, for default request it was broken.


source/s3_request_messages.c
```diff
@@ -754,11 +756,21 @@ struct aws_input_stream *aws_s3_message_util_assign_body(
...
+            /* set Content-Encoding header. If the header already exists, append the exisiting value to aws-chunked
              */
-            if (aws_http_headers_set(headers, g_content_encoding_header_name, g_content_encoding_header_aws_chunked)) {
+            struct aws_byte_cursor content_encoding_header_cursor;
+            aws_byte_buf_init_copy_from_cursor(
+                &content_encoding_header_buf, allocator, g_content_encoding_header_aws_chunked);
+            if (aws_http_headers_get(headers, g_content_encoding_header_name, &content_encoding_header_cursor) ==
+                AWS_OP_SUCCESS) {
+                aws_byte_buf_append_byte_dynamic(&content_encoding_header_buf, ',');
+                aws_byte_buf_append_dynamic(&content_encoding_header_buf, &content_encoding_header_cursor);
```
TingDaoK, line 767: > that applied the encodings MUST generate a Content-Encoding
   header field that lists the content codings in the order in which
   they were applied

https://www.rfc-editor.org/rfc/rfc7231.html#section-3.1.2.2

We should add aws-chunked to the end, right?

waahm7: I followed the example from https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html. I will try applying it to the end and see if that works.

source/s3_request_messages.c
```diff
@@ -754,11 +756,29 @@ struct aws_input_stream *aws_s3_message_util_assign_body(
         if (checksum_config->location == AWS_SCL_TRAILER) {
             /* aws-chunked encode the payload and add related headers */
 
-            /* set Content-Encoding header. TODO: the aws-chunked should be appended to the existing content encoding.
+            /* set Content-Encoding header. If the header already exists, append the exisiting value to aws-chunked
              */
-            if (aws_http_headers_set(headers, g_content_encoding_header_name, g_content_encoding_header_aws_chunked)) {
+            struct aws_byte_cursor content_encoding_header_cursor;
```
TingDaoK, line 761: Add a comment that we checked that the existing value cannot be `aws-chunked` [here](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L744-L760). 

And, debatable: is it worth to create a meta-request level of content-encoding header, so that we don't need to do this append every time when we create a single request?

waahm7: Thanks, I have updated the comment. I thought about adding it at meta-request level but this value is only used in one particular case which makes it a bit confusing to add it at the meta request level.

TingDaoK (approved):

PR #287, opened 2023-05-12
Add request type to telemetry

TingDaoK:
follow up PR for https://github.com/awslabs/aws-c-s3/pull/271#discussion_r1190024156


include/aws/s3/s3_client.h
```diff
@@ -66,6 +66,36 @@ enum aws_s3_meta_request_type {
...
+    AWS_S3_REQUEST_TYPE_AUTO_RANGED_PUT_LIST_PARTS,
+    AWS_S3_REQUEST_TYPE_AUTO_RANGED_PUT_CREATE_MULTIPART_UPLOAD,
+    AWS_S3_REQUEST_TYPE_AUTO_RANGED_PUT_PART,
+    AWS_S3_REQUEST_TYPE_AUTO_RANGED_PUT_ABORT_MULTIPART_UPLOAD,
+    AWS_S3_REQUEST_TYPE_AUTO_RANGED_PUT_COMPLETE_MULTIPART_UPLOAD,
+
+    /* Auto Ranged GET request types */
+    AWS_S3_REQUEST_TYPE_AUTO_RANGE_GET_HEAD_OBJECT,
+    AWS_S3_REQUEST_TYPE_AUTO_RANGE_GET_PART,
+    AWS_S3_REQUEST_TYPE_AUTO_RANGE_GET_INITIAL_MESSAGE,
```
jamesbornholt, line 84: unused?

include/aws/s3/s3_client.h
```diff
@@ -66,6 +66,36 @@ enum aws_s3_meta_request_type {
...
+
+    /* The default type. Same as default meta request */
+    AWS_S3_REQUEST_TYPE_DEFAULT,
```
jamesbornholt, line 87: nitpicky, but should this be first?

jamesbornholt (commented):
This looks like what we need, thanks!

include/aws/s3/s3_client.h
```diff
@@ -66,6 +66,36 @@ enum aws_s3_meta_request_type {
     AWS_S3_META_REQUEST_TYPE_MAX,
 };
 
+/**
+ * The type of S3 request made. Used by metrics.
+ */
+enum aws_s3_request_type {
+
+    /* Auto Ranged PUT request types */
+    AWS_S3_REQUEST_TYPE_AUTO_RANGED_PUT_LIST_PARTS,
```
graebm, line 75: Maybe simply use the [S3 API name](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html).

I'm not sure it's a great idea to use our private subclass names as part of the public API. Frees us up to change the names of things, or move to some other design scheme (break up subclasses differently, change to some other design pattern, etc)

```suggestion
    AWS_S3_REQUEST_TYPE_LIST_PARTS,
```

include/aws/s3/s3_client.h
```diff
@@ -66,6 +66,36 @@ enum aws_s3_meta_request_type {
...
+    AWS_S3_REQUEST_TYPE_AUTO_RANGED_PUT_CREATE_MULTIPART_UPLOAD,
+    AWS_S3_REQUEST_TYPE_AUTO_RANGED_PUT_PART,
+    AWS_S3_REQUEST_TYPE_AUTO_RANGED_PUT_ABORT_MULTIPART_UPLOAD,
+    AWS_S3_REQUEST_TYPE_AUTO_RANGED_PUT_COMPLETE_MULTIPART_UPLOAD,
+
+    /* Auto Ranged GET request types */
+    AWS_S3_REQUEST_TYPE_AUTO_RANGE_GET_HEAD_OBJECT,
```
graebm, line 82: weird that it's "RANGED" for put but "RANGE" for get (note missing "D")

include/aws/s3/s3_client.h
graebm: I'm not 100% sure how this data is going to be used.
Would it be better to just report a string in the metrics, instead of an enum?
Would it be OK if the string changed in some future version?
Is it better that we indicate what's going on under the hood, which is liable to change. Or is it better to be stable in what we report, and keep out details that might change in the future?

I don't know. maybe talk with someone who's thinking about using this

TingDaoK: I am not sure as well. I did this mainly because:
- I think we already expose the metrics as some inspect of the internal implementation. So, those private objects are as well part of the inspect of how we actually doing the job under the hood
- We are not matching the S3 api one by one. Like get the object size from copy, get object part, which is a ranged get, etc? 

BUT, I think it makes more sense to keep public stuff align with the public doc from S3, and user knows the meta request were made. So, it's okay if copy/put shared the same API calls. And, keep us free to change in the future.

jamesbornholt: The S3 API name is all we really want to know. We already know what type of meta request it is from the `aws_s3_meta_request` pointer the telemetry callback gets; we just don't have a way to distinguish e.g. HeadObject and GetObject calls in auto_ranged_get (so we can report on their latency separately).

include/aws/s3/s3_client.h
```diff
@@ -68,6 +68,26 @@ enum aws_s3_meta_request_type {
...
+
+    /* S3 APIs */
+    AWS_S3_REQUEST_TYPE_HEAD_OBJECT,
+    AWS_S3_REQUEST_TYPE_GET_OBJECT,
+    AWS_S3_REQUEST_TYPE_LIST_PARTS,
+    AWS_S3_REQUEST_TYPE_CREATE_MULTIPART_UPLOAD,
+    AWS_S3_REQUEST_TYPE_UPLOAD_PART,
+    AWS_S3_REQUEST_TYPE_ABORT_MULTIPART_UPLOAD,
+    AWS_S3_REQUEST_TYPE_COMPLETE_MULTIPART_UPLOAD,
+    AWS_S3_REQUEST_TYPE_MULTIPART_COPY,
```
DmitriyMusatkin, line 86: nit: for copy we use uploadPartCopy api

DmitriyMusatkin (approved):

PR #290, opened 2023-05-18
Async streaming 1st pass

graebm:
First pass at async streaming of the request body.

Merging to`async-stream` feature branch, not ready for `main` yet.

Relies on new classes: `aws_future` and `aws_async_stream`. These haven't been reviewed yet, but give feedback here if the APIs are confusing.

Description of changes:
- Internally, always read from `aws_async_stream`
    - Wrap old-fashioned `aws_input_stream` in an `aws_async_stream` if necessary.
- Change vtable->prepare_request() now returns an `aws_future`
    - This used to be synchronous, but now it may be async
    - In some cases this has been chopped up into MANY async steps, often a whole tree of async, which really increases the codebase's complexity 😬
- New conventions for async stuff
    - Async functions return `aws_future`, instead of taking a callback.
        - `aws_future` gives the user more control over how their callback is invoked.
    - ~Use `_async()` as suffix for functions that return `aws_future`~ (nobody liked this)
    - Use `_async_ctx` as suffix for the struct where data is stored until the operation completes.
        - Prefer to create a unique struct for each unique operation, even if their contents are identical. Hopefully helps future coders keep track of which step belongs to what overall operation.
    - Use `operation` as term for all work involved in fulfilling an async future.
        - Wanted to use the term "task", which is common in async/await land, but `aws_task` is already a thing.

Future work:
- Add tests
- Get `aws_future` and `aws_async_stream` reviewed
- Add docs and graphs to help devs understand all this additional complexity


source/s3_default_meta_request.c
```diff
@@ -208,20 +219,82 @@ static bool s_s3_meta_request_default_update(
...
+
     if (meta_request_default->content_length > 0 && request->num_times_prepared == 0) {
         aws_byte_buf_init(&request->request_body, meta_request->allocator, meta_request_default->content_length);
 
-        if (aws_s3_meta_request_read_body(meta_request, &request->request_body)) {
-            return AWS_OP_ERR;
-        }
+        /* Kick off the async read */
+        struct aws_future *read_future = aws_s3_meta_request_read_body(meta_request, &request->request_body);
+        aws_future_register_callback(read_future, s_s3_default_prepare_request_on_read_done, request_prep);
```
waahm7, line 245: Debatable: I'd prefer that we release the 'future' here and allow 'aws_future_register_callback' to acquire the reference before the callback, and then release it after the callback. This would eliminate the need to release it within the callback.

waahm7 (commented):
Great job at simplifying things using aws_future and aws_async_stream! I have a few minor suggestions, but overall, this is some truly outstanding work.

include/aws/s3/s3_client.h
```diff
@@ -380,11 +380,20 @@ struct aws_s3_meta_request_options {
...
      */
     struct aws_byte_cursor send_filepath;
 
+    /**
+     * Optional
+     * If set, the request body comes from this async stream.
+     * Use this when outgoing data will be produced in asynchronous chunks.
+     * If set, the `message` body-stream and `send_filepath` are ignored.
+     */
+    struct aws_async_stream *send_async_stream;
```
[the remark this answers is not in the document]
graebm: Changed it to be an error if more than 1 source is set.

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -56,15 +58,19 @@ struct aws_s3_meta_request_vtable {
...
         struct aws_s3_meta_request *meta_request,
         struct aws_s3_request *request,
         aws_s3_meta_request_prepare_request_callback_fn *callback,
         void *user_data);
 
-    /* Given a request, prepare it for sending (ie: creating the correct HTTP message, reading from a stream (if
-     * necessary), signing it, computing hashes, etc.) */
-    int (*prepare_request)(struct aws_s3_meta_request *meta_request, struct aws_s3_request *request);
+    /* Given a request, asynchronously prepare it for sending
+     * (creating the correct HTTP message, reading from a stream (if necessary), signing it, computing hashes, etc.).
```
TingDaoK, line 71: Already exist, but still: This vtable function doesn't sign the request, after the function returns. I guess remove it will be more clear?

source/s3_default_meta_request.c
graebm: We talked. The future callback no longer passes a pointer to itself. This makes the ownership model more obvious.  Now the consumer MUST store a pointer to the future or they won't have access to it from the callback. Now the consumer generally releases the reference to that future when they clean up everything else stored in their `async_ctx` struct.

include/aws/s3/s3_client.h
```diff
@@ -380,11 +380,20 @@ struct aws_s3_meta_request_options {
 
     /**
      * Optional.
-     * If set, this file is sent as the request body, and the `message` body-stream is ignored.
+     * If set, this file is sent as the request body.
      * This can give better performance than sending data using the body-stream.
+     * If set, the `message` body-stream and `send_async_stream` are ignored.
```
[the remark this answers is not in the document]
graebm: Fixed. Now it's an error if more than one is set, there's no priority to worry about

source/s3_meta_request.c
```diff
@@ -596,48 +609,36 @@ static void s_s3_meta_request_prepare_request_task(struct aws_task *task, void *
...
 
-    AWS_LOGF_ERROR(
-        AWS_LS_S3_META_REQUEST,
-        "id=%p Could not prepare request %p due to error %d (%s).",
-        (void *)meta_request,
-        (void *)request,
-        error_code,
-        aws_error_str(error_code));
+    int error_code = aws_future_get_error(preparation_future);
+    preparation_future = aws_future_release(preparation_future);
```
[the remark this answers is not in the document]
graebm: this is just the thing where release() functions return NULL
we didn't NEED to null out the pointer, but a lot of the aws-c-s3 code nulls out pointers once they're deleted so I stuck with the style

anyway, I changed this, now the future is released along with anything else stored in the async_ctx, and I don't bother nulling out pointers because it's all going to die 2 lines later

source/s3_auto_ranged_put.c
```diff
@@ -684,244 +743,545 @@ static int s_skip_parts_from_stream(
...
 
-        size_t request_body_size = s_compute_request_body_size(meta_request, part_index + 1);
+    for (; skip_ctx->part_index < skip_ctx->skip_until_part_number; ++skip_ctx->part_index) {
 
-        if (temp_body_buf.capacity != request_body_size) {
-            // reinit with correct size
-            aws_byte_buf_clean_up(&temp_body_buf);
-            if (aws_byte_buf_init(&temp_body_buf, meta_request->allocator, request_body_size)) {
-                return AWS_OP_ERR;
+        /* unless we just re-entered loop due to async read completion, kick off a new async read */
```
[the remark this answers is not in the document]
graebm: Agree it's confusing.
I tried writing this ... a few different ways. This is the best I could come up with. Suggestions welcom 🤷‍♀️

include/aws/s3/private/s3_meta_request_impl.h
graebm: Ooh, thanks, good point
I guess I still didn't understand the full flow after spending like a week editing this code 😅

source/s3_auto_ranged_put.c
```diff
@@ -684,244 +743,545 @@ static int s_skip_parts_from_stream(
...
+                aws_byte_buf_init(temp_body_buf, meta_request->allocator, request_body_size);
+            } else {
+                // reuse buffer
+                aws_byte_buf_reset(temp_body_buf, false);
+            }
+
+            read_part_future = aws_s3_meta_request_read_body(skip_ctx->meta_request, &skip_ctx->temp_body_buf);
+
+            /* the read may or may not complete immediately */
+            if (aws_future_register_callback_if_not_done(read_part_future, s_skip_parts_from_stream_loop, skip_ctx)) {
```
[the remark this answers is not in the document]
graebm: This is the worst-named function in the aws_future API
I am 1000% open to suggestions

I used to call it `aws_future_is_done_else_register_callback()` and it would return whether the future was done. But then everywhere I used it, I wanted to early-out if it was NOT done, and putting a `!`  or `== false` made it even harder to read, so I flipped what made it return true, and renamed it and 🤢

source/s3_default_meta_request.c
```diff
@@ -208,20 +219,82 @@ static bool s_s3_meta_request_default_update(
...
+    } else {
+        /* Don't need to read body, jump directly to the last step */
+        s_s3_default_prepare_request_finish(request_prep, AWS_ERROR_SUCCESS);
+    }
+
+    return preparation_future;
+}
+
+/* Completion callback for reading the body stream */
+static void s_s3_default_prepare_request_on_read_done(struct aws_future *read_future, void *user_data) {
```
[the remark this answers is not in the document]
graebm: I changed the API, now there is only `(void *user_data)`

source/s3_auto_ranged_get.c
```diff
@@ -336,17 +334,18 @@ static bool s_s3_auto_ranged_get_update(
     return work_remaining;
 }
 
-/* Given a request, prepare it for sending based on its description. */
-static int s_s3_auto_ranged_get_prepare_request(
-    struct aws_s3_meta_request *meta_request,
-    struct aws_s3_request *request) {
-    AWS_PRECONDITION(meta_request);
+/* Given a request, prepare it for sending based on its description.
+ * Currently, this is actually synchronous. */
```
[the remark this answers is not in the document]
graebm: Yeah, everyone agreed this was more confusing than helpful. I'm stripping ~_async~ off the names

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -45,26 +46,32 @@ typedef void(aws_s3_meta_request_prepare_request_callback_fn)(
...
 };
 
 struct aws_s3_meta_request_vtable {
     /* Update the meta request.  out_request is required to be non-null. Returns true if there is any work in
      * progress, false if there is not. */
     bool (*update)(struct aws_s3_meta_request *meta_request, uint32_t flags, struct aws_s3_request **out_request);
 
+    /* Run vtable->prepare_request() on the meta-request's event loop.
+     * We do this because body streaming is slow, and we don't want it on our networking threads.
+     * The callback may fire on any thread (an async sub-step may run on another thread). */
```
DmitriyMusatkin, line 64: more a question for later: do we have any cases where firing it on any thread is not what we want? interrupting other work on that thread? numa?

source/s3_auto_ranged_get.c
```diff
@@ -397,18 +396,24 @@ static int s_s3_auto_ranged_get_prepare_request(
...
         (void *)request,
         request->part_number);
 
-    return AWS_OP_SUCCESS;
+    success = true;
 
-message_alloc_failed:
-
-    return AWS_OP_ERR;
+finish:;
```
DmitriyMusatkin, line 409: nit: extra ;

source/s3_auto_ranged_get.c
```diff
@@ -397,18 +396,24 @@ static int s_s3_auto_ranged_get_prepare_request(
...
+    struct aws_future *future = aws_future_new(meta_request->allocator, AWS_FUTURE_VALUELESS);
+    if (success) {
+        aws_future_set_valueless(future);
```
DmitriyMusatkin, line 412: do you need to explicitly set valueless on a future that was created valueless?

source/s3_client.c
```diff
@@ -979,12 +999,11 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
                 return NULL;
             }
 
-            struct aws_input_stream *input_stream = aws_http_message_get_body_stream(options->message);
-
-            if ((input_stream == NULL) && (options->send_filepath.len == 0)) {
+            if (body_source_count == 0) {
```
DmitriyMusatkin, line 1002: do we need similar check for default case?

source/s3_request_messages.c
```diff
@@ -835,6 +836,49 @@ struct aws_input_stream *aws_s3_message_util_assign_body(
...
+    }
+
+    /* If user provides HTTP message with aws_input_stream body, wrap that in an async stream */
+    struct aws_input_stream *request_body = aws_http_message_get_body_stream(message);
+    if (request_body) {
+        return aws_async_stream_new_from_input_stream(allocator, request_body);
+    }
+
+    /* Otherwise, no body provided, just create empty async-stream */
+    struct aws_byte_cursor empty_cursor = {.len = 0};
```
DmitriyMusatkin, line 874: nit: why not just {0}? to avoid ptr having garbage?

source/s3_auto_ranged_put.c
```diff
@@ -684,244 +743,545 @@ static int s_skip_parts_from_stream(
...
+            message_future = s_s3_prepare_upload_part_async(request);
+            break;
+        case AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_COMPLETE_MULTIPART_UPLOAD:
+            message_future = s_s3_prepare_complete_multipart_upload_async(request);
+            break;
+        case AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_ABORT_MULTIPART_UPLOAD:
+            message_future = s_s3_prepare_abort_multipart_upload_async(request);
+            break;
+        default:
+            AWS_FATAL_ASSERT(0);
```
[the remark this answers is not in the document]
graebm: I added a TODO

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -45,26 +46,32 @@ typedef void(aws_s3_meta_request_prepare_request_callback_fn)(
     void *user_data);
 
 struct aws_s3_prepare_request_payload {
+    struct aws_allocator *allocator;
     struct aws_s3_request *request;
     aws_s3_meta_request_prepare_request_callback_fn *callback;
     void *user_data;
     struct aws_task task;
+    struct aws_future *preparation_future; /* aws_future<void> this operation is waiting on */
```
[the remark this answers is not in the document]
graebm: good point, "operation" in SDK-land usually refers to the the modeled request and everything around it

changed it to "job" and it reads pretty well

include/aws/s3/private/s3_meta_request_impl.h
graebm: we'll see how it shakes out in profiling

I'm currently just using the future's vanilla [register_callback()](https://github.com/awslabs/aws-c-io/blob/cbea39a7f40b1892bc4bf148234c805b7cc14192/include/aws/io/future.h#L64-L71) function, which fires the callback immediately when the future completes. I'm assuming we want to react as quickly as possible after the read completes.

We could also use the future's [register_event_loop_callback()](https://github.com/awslabs/aws-c-io/blob/cbea39a7f40b1892bc4bf148234c805b7cc14192/include/aws/io/future.h#L91-L103) function, so that the callback is always scheduled to run on the same IO thread. But if there's some other long-running thing happening on that thread, this would add delay in how quickly we can react to a slow read() call finishing up.

source/s3_auto_ranged_get.c
graebm: it's actually necessary, because I declare a variable on the next line.
See: https://stackoverflow.com/questions/8384388/variable-declaration-after-goto-label

source/s3_auto_ranged_get.c
graebm: yes, in this design, you got a FATAL_ASSERT if you called a getter or setter that didn't align with the enum passed to new()

but in the latest design it's type-safe and you can't compile the wrong getter/setter.

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -107,6 +114,9 @@ struct aws_s3_meta_request {
     /* Initial HTTP Message that this meta request is based on. */
     struct aws_http_message *initial_request_message;
 
+    /* Async stream for meta request's body */
+    struct aws_async_stream *send_async_body;
```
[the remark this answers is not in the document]
graebm: renamed to `request_body_async_stream`
still a bit awkward, but wanted to be clear that:
- it's not the response's body
- it's not an `aws_http_stream`
- it's not an `aws_input_stream` (it's an `aws_async_input_stream`)

source/s3_request_messages.c
graebm: If you use designated initializers, it zeroes out anything that you don't explicitly set, which is why I didn't set `.ptr = 0`

I'm pretty sure some compilers warn about `= {0}`. I'll give it a test, but if it doesn't work I'll use good old AWS_ZERO_STRUC(empty_cursor);

source/s3_client.c
graebm: Good question. Let's chat tomorrow

source/s3_request_messages.c
graebm: holy crap it worked! beautiful!

DmitriyMusatkin: nice. i was misremembering the standard on this one - setting any element in the struct does set other elements to 0. {0} initializer is standard since c89 and probably earlier so we should be fine using it everywhere

PR #295, opened 2023-05-25
Move total_num_parts to non synced data.

waahm7:
*Description of changes:*
- Make total_num_parts non async. total_num_parts is now defined at the time of creation and never changes. For no content_length case it will be zero.
- Replace some hacks around running total_num_parts with different hacks that don't rely on total_num_parts. 

### DIFF
--- a/include/aws/s3/private/s3_auto_ranged_put.h
+++ b/include/aws/s3/private/s3_auto_ranged_put.h
@@ -52,7 +52,8 @@ struct aws_s3_auto_ranged_put {
     struct {
         /*
          * Start index of skipping parts.
-         * This is used to keep track of how many parts have been read from input steam and where to try to start skipping parts from.
+         * This is used to keep track of how many parts have been read from input steam and where to try to start
+         * skipping parts from.
          */
         uint32_t part_index_for_skipping;
     } prepare_data;
--- a/include/aws/s3/private/s3_auto_ranged_put.h
+++ b/include/aws/s3/private/s3_auto_ranged_put.h
@@ -33,7 +33,7 @@ struct aws_s3_auto_ranged_put {
 
     /* Note: total num parts is known only if content-length is known,
     otherwise it will be 0. */
-    uint32_t total_num_parts;
+    uint32_t total_num_parts_from_content_length;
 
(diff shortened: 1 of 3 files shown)


source/s3_auto_ranged_put.c
```diff
@@ -1298,19 +1289,13 @@ static struct aws_future *s_s3_prepare_complete_multipart_upload(struct aws_s3_r
...
-    struct aws_s3_auto_ranged_put *auto_ranged_put = meta_request->impl;
 
     int error_code = aws_future_get_error(complete_mpu_prep->skipping_future);
     if (error_code != AWS_ERROR_SUCCESS) {
         s_s3_prepare_complete_multipart_upload_finish(complete_mpu_prep, error_code);
         return;
     }
 
-    /* Skipping was successful */
-    auto_ranged_put->prepare_data.num_parts_read_from_stream = auto_ranged_put->synced_data.total_num_parts;
```
waahm7, line 1312: Couple of options:
- Use size of etag list
- Only update it if content_length is available
- since num_parts_read_from_stream is only used for skipping parts, we can also just not update it.

graebm: let's just give this a name indicating it's only used for skipping. hard to think of a good one:
part_index_for_skipping
skipping_progress
skipping_resumes_at

waahm7: Thanks, updated it to part_index_for_skipping.

include/aws/s3/private/s3_auto_ranged_put.h
```diff
@@ -31,6 +31,10 @@ struct aws_s3_auto_ranged_put {
     uint64_t content_length;
     bool has_content_length;
 
+    /* Note: total num parts is known only if content-length is known,
+    otherwise it will be 0. */
+    uint32_t total_num_parts;
```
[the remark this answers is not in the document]
waahm7: I have renamed it to `total_num_parts_from_content_length` as I wanted to keep the word `total` and make it concise. If you don’t like it, we can change it again.

graebm (approved):

PR #297, opened 2023-05-31
Feature/presignedurl

eherozhao:
*Issue #, if available:*
The current S3 library doesn't have a support to use presigned url to ranged get object.

*Description of changes:*
- Updated the code to support presigned url for ranged_get.
- Added a test to validate the change.
- Added presigned url in `test_helper.py` since aws-c-s3 doesn't gave an api to generate presigned url and formatted the script.
- Updated tests/CMakeList and test_helper python script so that the customers don't have to manually run `export` and execute python3 script for ctest.


DmitriyMusatkin:
can you provide more info on how you are expecting it will work? 
presigned_url string is passed to meta request, but its only used to determine whether signing needs to be short circuited
is the expectation that user will override authority and path of the meta request with presigned url values? added test does not seem to be doing it

eherozhao:
> can you provide more info on how you are expecting it will work? presigned_url string is passed to meta request, but its only used to determine whether signing needs to be short circuited is the expectation that user will override authority and path of the meta request with presigned url values? added test does not seem to be doing it

Thanks for the question. That is a great question. At the beginning I was making this change with the change on aws-sdk-cpp since I also had to change on that side to generate a httpRequest with presigned url and pass it to userData->request so the new httpRequest can be used here with my change to pybass the signing process. Without the sdk-cpp change, it made this part of change a bit unrelated to my initial goal since like you said, all of my change here are just short circuiting the signing. So all in all, this change is to bypass the signing config whenever the presigned url is saved in options and customers can just pass in a httprequest created by the presigned url. (I will rephrase the description above to make it more clear.) As for the test, it was just to make sure the meta request was still working properly with this change.

include/aws/s3/s3_client.h
```diff
@@ -403,6 +403,9 @@ struct aws_s3_meta_request_options {
      * be used. If not NULL, these options will override the client options. */
     const struct aws_signing_config_aws *signing_config;
 
+    /* The presigned url*/
+    const char *presigned_url;
```
TingDaoK, line 407: Correct me if I am wrong:
Looks like all you need is to skip the signing with presigned url. 

I was going to suggest that leave `signing_config` to be NULL, but as a meta request, we cannot change the client level setting. 

But, as a workaround, you can pass anonymous credentials into the `signing_config`, which should skip the signing for you. It is a workaround without any code change to the lib

eherozhao: > Correct me if I am wrong: Looks like all you need is to skip the signing with presigned url.
> 
> I was going to suggest that leave `signing_config` to be NULL, but as a meta request, we cannot change the client level setting.
> 
> But, as a workaround, you can pass anonymous credentials into the `signing_config`, which should skip the signing for you. It is a workaround without any code change to the lib

May I ask how to pass anoymous credentials to the `signing_config`? 

Based on my investigation, the `aws_s3_meta_request_sign_request_default` is the function creates signing_config and pass it to meta_request. https://github.com/awslabs/aws-c-s3/pull/297/files#diff-436560b33218855638d58d1c14ecca658afd2c8ec65f2b525e21c6d829a7068aR755
so we have to bypass the following eventually which was what I did in the PR.
```
.sign_request = aws_s3_meta_request_sign_request_default,
```

TingDaoK:
```
    struct aws_signing_config_aws config;
    AWS_ZERO_STRUCT(config);
    struct aws_credentials *credentials = aws_credentials_new_anonymous(allocator);
    config.credentials = credentials;
```

TingDaoK: That function does NOT create the `signing_config`, it fetches config from meta request or client. https://github.com/awslabs/aws-c-s3/pull/297/files#diff-436560b33218855638d58d1c14ecca658afd2c8ec65f2b525e21c6d829a7068aR701-R716

```
if (meta_request->cached_signing_config != NULL) {
        signing_config = meta_request->cached_signing_config->config;
    } else if (client->cached_signing_config != NULL) {
        signing_config = client->cached_signing_config->config;
    } else {
        AWS_LOGF_DEBUG(
            AWS_LS_S3_META_REQUEST,
            "id=%p: No signing config present. Not signing request %p.",
            (void *)meta_request,
            (void *)request);
        on_signing_complete(NULL, AWS_ERROR_SUCCESS, user_data);
        return;
}
```

eherozhao: so if I understand right, your suggestion
```
 struct aws_signing_config_aws config;
    AWS_ZERO_STRUCT(config);
    struct aws_credentials *credentials = aws_credentials_new_anonymous(allocator);
    config.credentials = credentials;
```
should be put inside the `else` block before calling `on_signing_complete`. Am I right?

tests/s3_data_plane_tests.c
```diff
@@ -1294,6 +1294,72 @@ static int s_test_s3_put_object_with_part_remainder(struct aws_allocator *alloca
...
+
+    ASSERT_SUCCESS(aws_http_message_add_header(message, host_header));
+    ASSERT_SUCCESS (aws_http_message_set_request_method(message, aws_http_method_get));
+    ASSERT_SUCCESS (aws_http_message_set_request_path(message, g_test_presigned_url));
+
+    struct aws_s3_meta_request_options options;
+    AWS_ZERO_STRUCT(options);
+    options.type = AWS_S3_META_REQUEST_TYPE_GET_OBJECT;
+    options.message = message;
+    options.presigned_url = aws_string_c_str(presigned_url_string);
```
TingDaoK, line 1344: So, you can test like:
```suggestion
    struct aws_signing_config_aws signing_config;
    AWS_ZERO_STRUCT(signing_config);
    aws_s3_init_default_signing_config(&signing_config, g_test_s3_region, NULL);
    struct aws_credentials *credentials = aws_credentials_new_anonymous(allocator); 
    signing_config.credentials = credentials;
    options.signing_config = &signing_config;
    
```

eherozhao: I changed to this with a header <aws/auth/credentials.h> and a release `aws_credentials_release(credentials);`, but it threw the following error:
```
6: Fatal error condition occurred in /Users/haozhao/Downloads/simurgh/aws-c-s3/source/s3_util.c:278: credentials_provider
46: Exiting Application
46: [INFO] [2023-06-07T16:57:13Z] [000000016fa53000] [event-loop] - id=0x158104110: default timeout 100s, and max events to process per tick 100
46: [TRACE] [2023-06-07T16:57:13Z] [000000016fa53000] [event-loop] - id=0x158104110: waiting for a maximum of 100s 0ns
46: ################################################################################
46: Stack trace:
46: ################################################################################
46: 1   aws-c-s3-tests                      0x0000000100b01c8c aws_backtrace_print + 188
46: 2   aws-c-s3-tests                      0x0000000100adcdac aws_fatal_assert + 104
46: 3   aws-c-s3-tests                      0x0000000100a2af18 aws_s3_init_default_signing_config + 112
46: 4   aws-c-s3-tests                      0x00000001009cf3e0 s_test_s3_get_object_presigned_url + 1740
46: 5   aws-c-s3-tests                      0x00000001009c45c8 s_aws_run_test_case + 336
46: 6   aws-c-s3-tests                      0x00000001009c4b9c test_s3_get_object_presigned_url + 32
46: 7   aws-c-s3-tests                      0x00000001009b14a0 main + 1364
46: 8   dyld                                0x0000000100fd508c start + 520
1/1 Test #46: test_s3_get_object_presigned_url ...Subprocess aborted***Exception:   0.26 sec

```
it pointed to s3_util.c line 278 `void aws_s3_init_default_signing_config` so I assume the credential provider is not allowed to be `Null`?

Also, I think for our use case, we also need the logic that skipping signing config whenever presigned_url was identified. If not, we have to pass the "empty" signing config from sdk cpp or crt cpp side and I am not sure that is workable due to the untouchable Client class.

TingDaoK: well... Then.
```suggestion
    struct aws_signing_config_aws signing_config;
    AWS_ZERO_STRUCT(signing_config);
    struct aws_credentials_provider *provider = aws_credentials_provider_new_anonymous(allocator, NULL);
    aws_s3_init_default_signing_config(&signing_config, g_test_s3_region, provider);
    options.signing_config = &signing_config;
```

All I wanted to say is you can skip signing without changing aws-c-s3. :)

eherozhao:
```
struct aws_signing_config_aws signing_config;
AWS_ZERO_STRUCT(signing_config);
struct aws_credentials_provider_shutdown_options shutdown_options;
AWS_ZERO_STRUCT(shutdown_options);

struct aws_credentials_provider *provider = aws_credentials_provider_new_anonymous(allocator, &shutdown_options);
aws_s3_init_default_signing_config(&signing_config, g_test_s3_region, provider);
options.signing_config = &signing_config;
```
works but I didn't know whether this way of skipping signing config is something the team allows.

PR #300, opened 2023-06-02
S3 Endpoint Resolver

waahm7:
*Issue #, if available:*

*Description of changes:*
- Adds an opt-in aws-c-s3-endpoint-resolver.


include/aws/s3/s3_endpoint_resolver/s3_endpoint_resolver.h
```diff
@@ -0,0 +1,49 @@
+#ifndef AWS_S3_ENDPOINT_RESOLVER_H
+#define AWS_S3_ENDPOINT_RESOLVER_H
+
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/s3/s3.h>
+#include <aws/sdkutils/endpoints_rule_engine.h>
```
graebm, line 10:
```suggestion
```

in public headers,  avoid `#include`ing other headers, prefer to forward-declare anywhere you can get away with it

it keeps build times down for us and our users now and forever

source/s3_endpoint_resolver/aws_s3_endpoint_resolver_partition.c
```diff
@@ -0,0 +1,102 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates.
+ * All Rights Reserved. SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/s3/s3_endpoint_resolver/s3_endpoint_resolver.h>
+
+/**
+ * This file is generated using a Python script.
```
graebm, line 9: trivial: may as well mention the script by name/path

include/aws/s3/s3_endpoint_resolver/s3_endpoint_resolver.h
```diff
@@ -0,0 +1,49 @@
...
+
+AWS_PUSH_SANE_WARNING_LEVEL
+extern const char aws_s3_endpoint_resolver_partitions[];
```
DmitriyMusatkin, line 13: do we need it customer header? or can we have it in private header?

include/aws/s3/s3_endpoint_resolver/s3_endpoint_resolver.h
```diff
@@ -0,0 +1,49 @@
...
+extern const char aws_s3_endpoint_rule_set[];
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

include/aws/s3/s3_endpoint_resolver/s3_endpoint_resolver.h
```diff
@@ -0,0 +1,49 @@
...
+
+AWS_EXTERN_C_BEGIN
+
+/**
+ * Creates a new S3 endpoint resolver
+ */
+AWS_S3_API
+struct aws_s3_endpoint_resolver *aws_s3_endpoint_resolver_new(struct aws_allocator *allocator);
```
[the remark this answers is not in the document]
waahm7: Thanks, changed to generic type.

include/aws/s3/s3_endpoint_resolver/s3_endpoint_resolver.h
waahm7: Thanks, fixed.

source/s3_endpoint_resolver/aws_s3_endpoint_resolver_partition.c
waahm7: Thanks, updated.

include/aws/s3/s3_endpoint_resolver/s3_endpoint_resolver.h
waahm7: Thanks, fixed.

scripts/update_s3_endpoint_resolver_artifacts.py
```diff
@@ -0,0 +1,91 @@
...
+    elif c == '\n':
+        return '\\n'
+    elif c == '\r':
+        return '\\r'
+    elif c == '\t':
+        return '\\t'
+    elif c == '\v':
+        return '\\v'
+    else:
+        return c
```
graebm, line 29: trivial: a dict would be faster

scripts/update_s3_endpoint_resolver_artifacts.py
```diff
@@ -0,0 +1,91 @@
...
+
+if __name__ == '__main__':
+    session = boto3.session.Session()
+    s3 = session.client('s3', region_name='us-east-1')
+    bucket_name = 'aws-c-s3-endpoint-artifacts'
+
+    generate_c_file_from_json(
+        s3,
+        bucket_name,
+        's3-endpoint-rule-set.json',
```
graebm, line 82: opinion: might be useful for local debugging to allow just a source-filepath to be passed, rather than require that the script pull it from an S3 bucket

or have 2 scripts, one that just reads a src and writes to a dst, and another that pulls specific files from S3 and runs the script on them

source/s3_endpoint_resolver/s3_endpoint_resolver.c
```diff
@@ -0,0 +1,34 @@
...
+#include <aws/s3/s3_endpoint_resolver.h>
+#include <aws/sdkutils/endpoints_rule_engine.h>
+#include <aws/sdkutils/partitions.h>
+
+struct aws_endpoints_rule_engine *aws_s3_endpoint_resolver_new(struct aws_allocator *allocator) {
+    struct aws_endpoints_ruleset *ruleset = NULL;
+    struct aws_partitions_config *partitions = NULL;
+    struct aws_endpoints_rule_engine *rule_engine = NULL;
+
+    struct aws_byte_cursor ruleset_cursor = aws_byte_cursor_from_c_str(aws_s3_endpoint_rule_set);
```
graebm, line 15: utterly trivial: from_c_str() means an O(N) search to the end of the c-string

you could make these extern variables `aws_string` and define them using `AWS_STATIC_STRING_FROM_LITERAL()` or... who cares this is a call once thing

tests/s3_endpoint_resolver_tests.c
```diff
@@ -0,0 +1,51 @@
...
+#    include <aws/s3/s3_endpoint_resolver.h>
+#    include <aws/sdkutils/endpoints_rule_engine.h>
+
+AWS_TEST_CASE(test_s3_endpoint_resolver_create_destroy, s_test_s3_endpoint_resolver_create_destroy)
+static int s_test_s3_endpoint_resolver_create_destroy(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+    aws_s3_library_init(allocator);
+
+    struct aws_endpoints_rule_engine *rule_engine = aws_s3_endpoint_resolver_new(allocator);
+    aws_endpoints_rule_engine_release(rule_engine);
```
graebm, line 18:
```suggestion
    ASSERT_NOT_NULL(rule_engine);
    aws_endpoints_rule_engine_release(rule_engine);
```

tests/s3_endpoint_resolver_tests.c
```diff
@@ -0,0 +1,51 @@
...
+}
+
+AWS_TEST_CASE(test_s3_endpoint_resolver_resolve_endpoint, s_test_s3_endpoint_resolver_resolve_endpoint)
+static int s_test_s3_endpoint_resolver_resolve_endpoint(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+    aws_s3_library_init(allocator);
+
+    struct aws_endpoints_rule_engine *rule_engine = aws_s3_endpoint_resolver_new(allocator);
+    struct aws_endpoints_request_context *context = aws_endpoints_request_context_new(allocator);
+    ASSERT_SUCCESS(aws_endpoints_request_context_add_string(
```
graebm, line 31:
```suggestion
    struct aws_endpoints_rule_engine *rule_engine = aws_s3_endpoint_resolver_new(allocator);
    ASSERT_NOT_NULL(rule_engine);
    struct aws_endpoints_request_context *context = aws_endpoints_request_context_new(allocator);
    ASSERT_NOT_NULL(context);
    ASSERT_SUCCESS(aws_endpoints_request_context_add_string(
```

graebm (approved):
fix & ship

tests/s3_endpoint_resolver_tests.c
graebm: or just remove test since next one is also very simple

source/s3_endpoint_resolver/s3_endpoint_resolver.c
waahm7: Thanks, fixed by generating a byte cursor directly.

CMakeLists.txt
```diff
@@ -75,9 +79,10 @@ file(GLOB S3_HEADERS
 
 file(GLOB S3_SRC
         ${AWS_S3_ROOT_SRC}
-        ${AWS_S3_MODEL_SRC}
-        ${AWS_S3_EXTERNAL_SRC}
         )
+if (AWS_ENABLE_S3_ENDPOINT_RESOLVER)
+    list(APPEND S3_SRC ${AWS_S3_ENDPOINT_RESOLVER_SRC})
```
TingDaoK, line 84: I don't understand how this works, just throw out my question:
So, this will only compile the `source/s3_endpoint_resolver/*.c` when the cmake config is set. What about the header file? Will Cmake not compile the header files as well?

waahm7: Yes, it will only include the source when the CMake configuration is set. The header file is always included to simplify things. However, we warn users not to use it unless it's enabled.

include/aws/s3/s3_endpoint_resolver.h
```diff
@@ -0,0 +1,27 @@
...
+AWS_PUSH_SANE_WARNING_LEVEL
+
+struct aws_endpoints_request_context;
+struct aws_endpoints_rule_engine;
+AWS_EXTERN_C_BEGIN
+
+/**
+ * Creates a new S3 endpoint resolver.
+ * Warning: Before using this header, you have to enable it by
+ * setting AWS_ENABLE_S3_ENDPOINT_RESOLVER=ON
```
TingDaoK, line 20: I would be more specific about this is a CMake config, instead of like env var or something.

waahm7: Thanks, fixed.

rvansa:
Hi @waahm7 , is there any information available how this should be used? I am trying to direct CRT to use a `localstack` endpoint and this seems to be relevant; I can't find any API that would accept `struct aws_endpoints_rule_engine`. Thanks!

waahm7:
@rvansa This is just a helper to resolve s3 endpoints. You need to override the endpoint at https://github.com/awslabs/aws-c-s3/blob/main/include/aws/s3/s3_client.h#L815.

rvansa:
@waahm7 Thanks, I found that it is overridden here; I had a bit of hope that this would be intended for a client-wide override as in `Aws::S3::S3ClientConfiguration.endpointOverride`. The problem with `aws_s3_meta_request_options` is that the override is not exposed in all API, e.g. when trying to use `aws_s3_initiate_list_objects`. But there's more problems with that - a quick hacking revealed that the virtual host style starts conflicting when I actually try to list the buckets.

I am wondering how mature CRT really is - from docs I got the impression that it is the high-performant basis of SDK, but even with `Aws::S3Crt::S3CrtClient` the code seems to use CURL. And after couple of hacks, getting CRT to talk to `localstack` instance I found that the XML parser [can't parse simple XML response](https://github.com/awslabs/aws-c-common/issues/1165). Not speaking of not-implemented features like SSO token refresh.

waahm7:
> the code seems to use CURL.

We have our own HTTP client and don't use Curl. Curl is only used in some integration tests for sanity checks as it is easier to use and compare against.
> this would be intended for a client-wide override 

CRT S3 Client is designed to sit under the SDKs. The SDKs resolve the endpoint, construct an HTTP request, and pass that as a `meta_request`. The `http_request` normally has a host header which is used to connect to the endpoint. Overriding the host header or the `endpoint_override` should work, but there is no client-level override AFAIK. These knobs exist in the SDK, and the CRT S3 Client is a fairly low-level API.

This client is more suited for high-performance GetObject and PutObject operations only. For generic S3 usage like listObjects, the recommendation is normally to use an SDK. Any reasons you are trying to use `aws-c-s3` directly over an SDK?

DmitriyMusatkin:
Just a couple things to note:  aws_s3_initiate_list_objects is a private function (as its defined in a private header). it does support endpoint override, but in a different way from how meta requests do it. 

As waahm7 mentioned, aws-c-s3 is not meant to be a full featured sdk, but rather a library for other sdks to implement on top of. You can implement any s3 api on top of aws-c-s3 by using default meta. request type, but that does require manual construction of http request. Some SDKs like java do it. Some SDKs, like cpp you mentioned, only bind out put and get - mostly because default was not quite ready yet when they integrated and there has not been any pressing performance reasons to integrate with non-data plane functions. 

Note: this is library backing official AWS SDKs. We dont test against anything other than S3 and thus cant guarantee that the client will work against all third party implementations. With that said, we are happy to address any misses in the generic core functionality as long as they are not specific to 3p clients

rvansa:
I feel a little guilty hijacking a merged PR for a discussion like this, but thanks a lot for these answers, folks!

About the curl - yes, CRT does not use curl, I was trying to say that S3Client in SDK uses C++ for listing objects... The answer from Dmitry explains that (I've used only listing buckets updating the [S3 Hello World](https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/cpp/example_code/s3/hello_s3/hello_s3.cpp) - should have probably tried [S3 CRT demo](https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/cpp/example_code/s3-crt/s3-crt-demo.cpp) as well).

> Any reasons you are trying to use aws-c-s3 directly over an SDK?

My main concern is how bloated SDK is. Being able to communicate with AWS is just one feature of my library (roughly 0.5 MB), but statically linking S3 Hello World against SDK yields 57 MB with symbols, 17 MB without. With CRT I can get to ~acceptable size. And the system dependencies (libstdc++, libcurl...) are annoying if the library/app using it is not distributed through package manager as well (not up to me...).

>  aws_s3_initiate_list_objects is a private function

Good point; I started off [the only example I could find](https://github.com/awslabs/aws-c-s3/blob/main/samples/s3/s3-ls.c) and missed this. So this private is there just for testsuite, right? Together with the fact that not performance-sensitive calls are made straight through per-SDK HTTP library it starts making sense...

> We dont test against anything other than S3 and thus cant guarantee that the client will work against all third party implementations.

Understandable; you can't test everything. For time reasons I'll be using SDK now, though I hope to revisit CRT. I'll file issues if I find other problems.

PR #301, opened 2023-06-02
Tests for async streaming (seems to work!)

graebm:
- Add basic tests for async streaming of the request body.
- When async streaming, don't let more than 1 part at a time be in the "prepare" stage.
    -  since `aws_async_input_stream` doesn't allow overlapping read() calls.
- Stop wrapping synchronous streams in an async stream.
    - Now that we treat async streaming slightly different than synchronous, it's simpler to just have them be separate things.

Future work:
- more tests
- benchmarking / optimization


source/s3_meta_request.c
```diff
@@ -1588,7 +1596,38 @@ struct aws_future_bool *aws_s3_meta_request_read_body(
...
+    struct aws_stream_status status = {.is_end_of_stream = false, .is_valid = true};
+    while ((buffer->len < buffer->capacity) && !status.is_end_of_stream) {
+        /* Read from stream */
+        if (aws_input_stream_read(synchronous_stream, buffer) != AWS_OP_SUCCESS) {
+            aws_future_bool_set_error(synchronous_read_future, aws_last_error());
+            goto synchronous_read_done;
+        }
+
+        /* Check if stream is done */
+        if (aws_input_stream_get_status(synchronous_stream, &status) != AWS_OP_SUCCESS) {
```
TingDaoK, line 1621: trivial: do we also need to check `is_valid`?

tests/s3_tester.c
```diff
@@ -231,6 +233,25 @@ static void s_s3_test_meta_request_telemetry(
...
+    void *user_data) {
+
+    (void)meta_request;
+    AWS_ASSERT(meta_request);
+    AWS_ASSERT(progress);
+    AWS_ASSERT(user_data);
+
+    struct aws_s3_meta_request_test_results *meta_request_test_results = user_data;
+
+    aws_atomic_fetch_add(&meta_request_test_results->total_bytes_uploaded, (size_t)progress->bytes_transferred);
```
TingDaoK, line 248: worth to leave a comment about this `total_bytes_uploaded` only works for MPU, so that we won't surprise us again in the future.

TingDaoK (approved):

source/s3_meta_request.c
graebm: ehhhhhh

if read() fails, or get_status() fails I get an error-code I can report, but if `is_valid` is false I have to make up an error code

we never check `is_valid` in aws-c-http. if it's invalid, it will tell me the next time I try to read()

waahm7 (approved):

DmitriyMusatkin (approved):

PR #316, opened 2023-06-15
Not retry on TLS failure

TingDaoK:
- TLS negotiation failure not worth to retry, it's expensive and not really be different by retry.


graebm (approved):
The code changes looks solid, but please edit the PR description or add a comment in code,  explaining WHY we're making this change

Issue #317, opened 2023-06-16
S3-Transfer-Manager Streaming


djchapm:
### Describe the bug

I've been testing combination of latest aws-c-s3 update regarding #285 and the ability to stream data of unknown size to S3 using a publisher.

I have 2 concerns aside from errors I'm getting - 

1. For small loads this works great, but starts breaking down somewhere after 128MB at least in my setup.  I am not sure if the issue is with this project or the java v2 sdk for s3-transfer-manager - feel free to redirect.
2. As the size increases, there is a jump in time to complete the upload - at 128M I've seen the completion step is anywhere from ~4s to 18s,  

**Platform**: Macbook Pro, Ventura 13.4, 32GB, Apple M1 Max ARM64 processor.
**Aws CLI**: aws-cli/2.12.1 Python/3.11.4 Darwin/22.5.0 source/arm64
**TargetThroughput**: 20.0Gbps
**Minimum Part Size**: 1000000L (I think this causes issues after 128M)

I've mocked up data which is just a string of 128 Bytes that I send over and over in a ByteBuffer.

Errors include failed response (400) from awssdk, missing checksums for parts, and SIGSEGV on libobjc.A.dylib... I've also received a SIGABRT which doesn't even give me a dump.

Attaching a simple Java project to test - configure whichever credentials and a bucket name and execute - as you increase num lines you'll start to see issues.  I create the crt log in whatever your work directory happens to be.

### Expected Behavior

Files uploaded successfully and can manage at least 100G S3 Objects.
Closing the stream doesn't grow with file size so drastically as it does now.

### Current Behavior

Crashes, Failed uploads, heavy delay on completing the upload.

### Reproduction Steps

[s3AsyncTest.zip](https://github.com/awslabs/aws-c-s3/files/11776325/s3AsyncTest.zip)

### aws-c-s3 version used

aws.sdk.version 2.20.79, aws.crt.version 0.22.1

### Compiler and version used

openjdk 17.0.3 2022-04-19

### Operating System and version

Darwin US10MAC44VWYPKH 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:52:24 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T6000 arm64 arm Darwin

TingDaoK:
Hey @djchapm , thanks for reporting the issue.

I did reproduce the crash and 400 sometimes and it's a bug from our side, and I have a possible fix for it.

[...]

PR #318, opened 2023-06-19
use aws_array_list<aws_byte_buffer*> for encoded_checksum_list

TingDaoK:
*Issue #, if available:*

- https://github.com/awslabs/aws-c-s3/issues/317
- We used the array list of `struct aws_byte_buf` to store the checksum. And point to the checksum buffer in the list while we read from the stream and calculate the checksum. However, while we are using the buffer, the other thread from complete part callback can allocate a new part when we finish reading from the other part and may result in expand the array_list, which frees the old memory we are using, and lead to a crash.
  - [here](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_auto_ranged_put.c#L1189) is where the list may grow
  - [here](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_auto_ranged_put.c#L1234) is where we read from the list and pass it down to the input stream read

*Description of changes:*
- use array list of `struct aws_byte_buf*` instead. However, the downside is we need to allocate the buffer every time we create a new one, which will add N allocation time, and may impact the performance (?)

TODO:
- the checksum buffer will be initialized and updated by the `aws_chunk_stream` while we read from the input. So, if we use dynamic array with pre-allocated buffer, there will be a race condition between the expansion of the array to copy the buffer around and the buffer gets updated from the other thread. To prevent that:
  - a lock when we initialize the checksum buffer [here](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_chunk_stream.c#L72), which I don't like by adding lock around.
  - Instead of `aws_chunk_stream` initialize the buffer, it just fill the given one without reallocate, which sounds like a workable solution if we have performance issue from extra allocation by this PR.


DmitriyMusatkin (approved):

source/s3_auto_ranged_put.c
```diff
@@ -1231,10 +1237,19 @@ static void s_s3_prepare_upload_part_finish(struct aws_s3_prepare_upload_part_jo
...
+            request->part_number - 1);
         /* Clean up the buffer in case of it's initialized before and retry happens. */
-        aws_byte_buf_clean_up(checksum_buf);
+        checksum_buf = *checksum_buf_pointer;
+        if (checksum_buf) {
+            aws_byte_buf_clean_up(checksum_buf);
+        } else {
+            checksum_buf = aws_mem_calloc(meta_request->allocator, 1, sizeof(struct aws_byte_buf));
+            *checksum_buf_pointer = checksum_buf;
+        }
```
graebm, line 1252: this code is accessing `synced_data.encoded_checksum_list` without holding the lock

also, it's so damned confusing, it took me 20+ minutes to wrap my head around why we need to sometimes clean_up(), and other times allocate (but not aws_byte_buf_init())

I'm torn between suggesting comments that explaining what's going on here

or suggesting ways to make this code less confusing (stop allocating the byte-buf that the weird stream will initialize later)

maybe we can move this 1 block of code around to make more sense:
1) delete everything here.  `s_s3_prepare_upload_part_finish()` should not touch `synced_data.encoded_checksum_list`
2) move the `aws_mem_calloc(..., sizeof(struct aws_byte_buf))` up into the function above this, where we're already holding a lock to resize `synced_data.encoded_checksum_list`. But instead of setting NULL, set the allocated byte-buf
3) move the `aws_byte_buf_clean_up(checksum_buf);` up into `s_s3_prepare_upload_part()` in the same else block with the comment `/* Not the first time preparing request (e.g. retry).`

source/s3_auto_ranged_put.c
```diff
@@ -936,15 +942,15 @@ static void s_skip_parts_from_stream_loop(void *user_data) {
             goto on_done;
         }
 
-        struct aws_byte_buf checksum_buf;
+        struct aws_byte_buf *checksum_buf;
         aws_array_list_get_at(&auto_ranged_put->synced_data.encoded_checksum_list, &checksum_buf, skip_job->part_index);
 
         // compare skipped checksum to previously uploaded checksum
```
graebm, line 948:
```suggestion
        // if previously uploaded part had a checksum, compare it to what we just skipped
```

source/s3_auto_ranged_put.c
```diff
@@ -936,15 +942,15 @@ static void s_skip_parts_from_stream_loop(void *user_data) {
...
-        if (checksum_buf.len > 0 && s_verify_part_matches_checksum(
-                                        meta_request->allocator,
-                                        *temp_body_buf,
-                                        meta_request->checksum_config.checksum_algorithm,
-                                        checksum_buf)) {
+        if (checksum_buf && s_verify_part_matches_checksum(
```
graebm, line 949: this code makes me realize that the array-list entries could be in 3 states:
- NULL because ListParts says they didn't have a checksums before
- Allocated (but uninitialized) because we've read the part, but haven't calculated a checksum yet (or we're not doing checksums at all)
- Allocated, initialized, and filled

Is this an issue? That there are 2 "invalid" states (NULL or len==0) now instead of just 1?

source/s3_auto_ranged_put.c
DmitriyMusatkin: +1, but i think we should address it in 2 steps. let fix the issue first with the way it currently works and lets discuss how to refactor it better and we can do that in follow up pr

graebm: yeah, me and Dengke just had a chat and arrived at the same conclusion.
I can revisit this when I let the user pass in checksum

PR #322, opened 2023-06-30
Use 1 array for part info, instead of separate etag array and checksum array

graebm:
**Issue:**
I wanted to track more info about each part. We already had separate arrays for the parts' etags and checksums, and the code to deal with them was always finicky since parts may arrive out of order and the array-list API is clunky.

**Description of changes:**
Use a single array-list to track all part info, instead of separate arrays for etags, checksums, size, etc


source/s3_auto_ranged_put.c
```diff
@@ -1739,10 +1706,11 @@ static void s_s3_auto_ranged_put_request_finished(
...
                         ++auto_ranged_put->synced_data.num_parts_successful;
 
-                        /* ETags need to be associated with their part number, so we keep the etag indices
-                         * consistent with part numbers. This means we may have to add padding to the list in the
-                         * case that parts finish out of order. */
-                        aws_array_list_set_at(&auto_ranged_put->synced_data.etag_list, &etag, part_index);
+                        /* Store part's ETag */
+                        struct aws_s3_put_part_info *part = NULL;
+                        aws_array_list_get_at(&auto_ranged_put->synced_data.part_list, &part, part_index);
+                        AWS_ASSERT(part != NULL);
```
TingDaoK, line 1712: maybe worth to add `AWS_ASSERT(part->etag == NULL);`

TingDaoK (approved):

include/aws/s3/private/s3_copy_object.h
```diff
@@ -32,7 +32,7 @@ struct aws_s3_copy_object {
 
     /* Members to only be used when the mutex in the base type is locked. */
     struct {
-        struct aws_array_list etag_list;
```
DmitriyMusatkin, line 35: nit: does it actually make sense to use aws_s3_put_part_info here(comment about list actual type would also be nice)?
copy for example would never need checksums. 
should name at least be more generic?

source/s3_auto_ranged_put.c
```diff
@@ -119,10 +119,15 @@ static int s_s3_auto_ranged_put_pause(
...
 
-static bool s_process_part_info(const struct aws_s3_part_info *info, void *user_data) {
+static bool s_process_part_info_synced(const struct aws_s3_part_info *info, void *user_data) {
     struct aws_s3_auto_ranged_put *auto_ranged_put = user_data;
+    struct aws_s3_meta_request *meta_request = &auto_ranged_put->base;
 
-    struct aws_string *etag = aws_strip_quotes(auto_ranged_put->base.allocator, info->e_tag);
+    ASSERT_SYNCED_DATA_LOCK_HELD(auto_ranged_put);
+
+    struct aws_s3_put_part_info *part = aws_mem_calloc(meta_request->allocator, 1, sizeof(struct aws_s3_put_part_info));
```
DmitriyMusatkin, line 128: something to consider: part list is already by value, so instead of allocating and then setting data into array, we can ensure list size, get a ptr to element at given part index and update its fields

source/s3_auto_ranged_put.c
```diff
@@ -421,29 +410,21 @@ static void s_s3_meta_request_auto_ranged_put_destroy(struct aws_s3_meta_request
...
-         ++etag_index) {
-        struct aws_string *etag = NULL;
+    for (size_t part_index = 0; part_index < aws_array_list_length(&auto_ranged_put->synced_data.part_list);
+         ++part_index) {
 
-        aws_array_list_get_at(&auto_ranged_put->synced_data.etag_list, &etag, etag_index);
-        aws_string_destroy(etag);
+        struct aws_s3_put_part_info *part;
+        aws_array_list_get_at(&auto_ranged_put->synced_data.part_list, &part, part_index);
+        if (part) {
```
DmitriyMusatkin, line 418: super minor nit: within this pr both if (a) and if (a != NULL) are used. i know we are not super consistent with it in general, but i think we should prefer one style. from my experience c-s3 does != NULL more often

include/aws/s3/private/s3_copy_object.h
graebm: [UploadPartCopy](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html#API_UploadPartCopy_ResponseSyntax) CAN actually generate a checksum, and we could use them some day...

The reason I changed s3_copy_object.h/c to use the new `aws_s3_put_part_info` is that it also needs to create a CompleteMultipartUpload request, and I changed the function that creates it to take a "part list" instead of separate "etag list" and "checksum list". So both `s3_copy_object.h` and `s3_auto_ranged_put.h` need to agree on how they're storing part info. I didn't realize we needed this in copy ... so agree the name isn't super generic. But basically it's "the data you'll need for CompleteMultipartUpload later"

The reason the name isn't super generic is that I started with `aws_s3_part_info` and discovered that name was already being used by the ListParts call. I could just rename that one to `aws_s3_list_part_info` and steal the nice short generic name for use here, or think of a more generic name like `aws_s3_upload_part_info`??? `aws_s3_multipart_info`???

graebm: aws_s3_mpu_part_info !
good name
thanks

source/s3_auto_ranged_put.c
graebm: Remember how Dengke needed to change the byte-buffers to be heap-allocated instead of by-value because the list resizes, but we pass pointers around?: https://github.com/awslabs/aws-c-s3/pull/318

Anyway, I didn't change the weird way we store checksum, it still needs to be on the heap. But I figured I'd just put this whole struct on the heap (with the byte-buffer by-value within it). I took advantage of the on-the-heap nature in another place too: see [line 883](https://github.com/awslabs/aws-c-s3/pull/322/files#diff-1b77484ee8493ab0ab8653a326f4bb0ac42a569e63261469b653ab544d27d99aR883) of the s_skip_parts_from_stream_loop()

source/s3_auto_ranged_put.c
graebm: good point, I'll go explicit

DmitriyMusatkin (approved):

PR #327, opened 2023-07-06
Allow user to review checksums before multipart upload completes

graebm:
**Issue:**
It is very hard for the user to do end-to-end checksum verification for multipart uploads.

A user might buffer data in memory before passing it into the S3 client. The S3 client will compute checksums on whatever data is passed in, but what if this buffered data suffers a bit flip before hand? The S3 client would simply compute the checksum of the corrupted data.

It's easy if the object is uploaded in 1 request via PutObject. The user can simply compare their own checksum with the appropriate response header (e.g. [x-amz-checksum-crc32](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_ResponseSyntax)).

But it's hard for multipart upload, where the object's final checksum is calculated differently. The algorithm is [described here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html#large-object-checksums). The algorithm is basically: checksum of all parts' checksums concatenated together, then add "-14" (if there were 14 parts).

Currently, to verify multipart upload a user would need to reproduce the part-splitting logic of aws-c-s3 (which might change in the future), and reproduce S3's final checksum algorithm (I'm not sure if this could ever change?).

**Description of Changes:**
Add a `review_upload_callback` allowing users to "review" the part boundaries and part checksums calculated by aws-c-s3, and cancel the multipart upload if they don't agree.

**Design Considerations:**
This design isn't perfect, so it's marked "experimental/unstable" for now.

If you're using CRC32, there's a trick where you can [combine part-part checksums](https://stackoverflow.com/questions/6491834/is-it-possible-to-do-crc-32-calculation-in-splits) to get the whole-object checksum. The customer asking for this feature is taking this approach, so we're providing this to unblock them.

But other algorithms (e.g. SHA-256) don't allow this trick. The user would need to re-review all their data, calculating a checksum per part based on the part-boundaries provide by this callback, plus a whole-object checksum to compare against their original. Or the user would need to reproduce aws-c-s3's part-splitting logic as they stream data the first time, them from this callback they could confirm that their part-splitting logic matches ours. But if aws-c-s3's logic ever changed, they'd need to re-write their code to adjust.

My first instinct was to let the user provide the per-part checksums, via a callback that occurred after streaming data from each part. But I realized that, to truly guarantee end-to-end integrity, the user would still need to anticipate the part boundaries and reproduce aws-c-s3's part splitting logic. Or the user would need to keep each part's data buffered until this checksum callback occurred and re-scan it then. Also, this part-part callback based on recent reads would be hard to fit in a possible future where we read data for multiple parts in parallel.


include/aws/s3/s3_client.h
```diff
@@ -487,6 +550,14 @@ struct aws_s3_meta_request_options {
      */
     aws_s3_meta_request_telemetry_fn *telemetry_callback;
 
+    /**
+     * Optional.
+     * Callback for reviewing an upload before it completes.
+     * WARNING: experimental/unstable
+     * See `aws_s3_upload_review_fn`
+     */
+    aws_s3_meta_request_upload_review_fn *upload_review_callback;
```
TingDaoK, line 559: If we going to support single part for the same functionality, how can we cancel the put?

For a single part, there will only be request, and if we invoke the callback as we read through the body, which has already been sent to the server and request has basically finished. We will need to delete the object. Will that be an overkill? 

Just bring it up as the naming is `upload_review`. If we only gonna support MPU for this, we should name it as mpu

graebm: my think was: to make it work with singlepart, we'd need to very carefully fire this callback as part of the input_stream_read()

graebm: Or just do checksum calculation when we initially stream data from the user, which would be way simpler, but we'd need benchmarks to prove that it doesn't degrade performance

TingDaoK (approved):

source/s3_auto_ranged_put.c
```diff
@@ -1323,22 +1323,76 @@ static struct aws_future_http_message *s_s3_prepare_complete_multipart_upload(st
...
+    }
+
+    aws_s3_meta_request_unlock_synced_data(meta_request);
+    /* END CRITICAL SECTION */
+
+    /* Invoke callback */
+    int error_code = AWS_ERROR_SUCCESS;
+    if (meta_request->upload_review_callback(meta_request, &review, meta_request->user_data) != AWS_OP_SUCCESS) {
+        error_code = aws_last_error_or_unknown();
+    }
```
waahm7, line 1378: Are we expecting the user to set a proper error code? I think we should at least replace it with a relevant error in the event of an unknown case. Also, we should mention in the documentation of the callback how to raise the error.

waahm7 (approved):

source/s3_auto_ranged_put.c
graebm: I had the same thought when I wrote this
Users ARE expected to properly raise errors.
The documentation is bad because we didn't want to add 10 lines explaining proper error handling to every single callback.

I went back and forth on what to do... but ended up adding a short 3 line explanation to every callback in s3_client.h

graebm: user's aren't just expected to raise errors properly, it's required
otherwise they'll end up with a random error code
if no other error has EVER been raised on this thread, they'll end up with AWS_ERROR_UNKNOWN
but if any other error has EVER been raised on this thread, that's the error code they'll end up seeing

IMHO: the whole aws_raise_error() system is very very fragile

Issue #328, opened 2023-07-07
Fail to download full folder with 15K images and 7K text files


dbickson:
### Describe the bug

We are following the example by compiling the s3 demo copy program and running it on ubuntu 20.04 using the command line:
 ```nohup time aws-c-s3/build/samples/s3/s3 cp s3://vl-sample-dataset-kitti/Kitti/ ~/Kitti --region us-east-2 &```

The program runs fine without error but the number of received file is around 11K namely half the files are missing. 

```
# Kitti is the folder of received files with s3 c demo
ubuntu@ip-172-31-30-217:~/Kitti$ du -sh .
6.1G	.
ubuntu@ip-172-31-30-217:~/Kitti$ find . -type f | wc
  11647   11647  391999

# Kitti2 is the data downloaded using aws s3 sync
ubuntu@ip-172-31-30-217:/mnt/data/crtsdk$ find ~/Kitti2/ | wc
  22487   22487 1161674
ubuntu@ip-172-31-30-217:/mnt/data/crtsdk$ du -sh ~/Kitti2
12G	/home/ubuntu/Kitti2
```

### Expected Behavior

All files should be copied locally

### Current Behavior

Only 50\% of the files are copied, there is no error.
The machine has enough disk space for the copy:
```
ubuntu@ip-172-31-30-217:~$ df -k .
Filesystem     1K-blocks      Used Available Use% Mounted on
/dev/root      304681132 245898460  58766288  81% /
ubuntu@ip-172-31-30-217:~$
```
Attached below are the running logs, of listing aws via the aws-c-s3 (number of files is correctly 22k, listing via aws s3 command (again 22k) and the full output of the run that copied 11k files. 
We have opened the bucket permissions so you can try on your own.

[logs.zip](https://github.com/awslabs/aws-c-s3/files/11975388/logs.zip)

[s3_exec.zip](https://github.com/awslabs/aws-c-s3/files/11975420/s3_exec.zip)

### Reproduction Steps

 ```nohup time aws-c-s3/build/samples/s3/s3 cp s3://vl-sample-dataset-kitti/Kitti/ ~/Kitti --region us-east-2 &```

### Additional Information/Context

Interestingly, when looking at the number of download printouts, 
 grep "download: s3://vl-sample-dataset-kitti/Kitti/raw" nohup.out | sort -u | wc
I see somehitng between 34K to 44K printouts. Maybe due to multithreading?

 

### aws-c-s3 version used

[...]

PR #330, opened 2023-07-07
Sample improve, exit when error happens instead of raise error

TingDaoK:
*Issue #, if available:*

- The callback for list object doesn't raise error to be caught by the operation. The user needs to keep record of the error happens. 
- https://github.com/awslabs/aws-c-s3/issues/328

*Description of changes:*

- As for the sample, do raise error from callback, just exit with error print out.


source/s3_list_objects.c
```diff
@@ -142,7 +142,7 @@ static bool s_on_list_bucket_result_node_encountered(
         }
 
         if (ret_val && operation_data->on_object) {
-            ret_val |= operation_data->on_object(&fs_wrapper.fs_info, operation_data->user_data);
+            ret_val &= operation_data->on_object(&fs_wrapper.fs_info, operation_data->user_data);
```
graebm, line 145: trivial: rename ~ret_val~ -> `keep_going` or something, so we don't confuse it with an `int` again

samples/s3/s3-cp.c
```diff
@@ -665,15 +678,19 @@ static int s_kickoff_get_object(
...
     struct aws_string *mode = aws_string_new_from_c_str(cp_app_ctx->app_ctx->allocator, "wb");
     transfer_ctx->output_sink = aws_fopen_safe(file_path, mode);
-    aws_string_destroy(mode);
-    aws_string_destroy(file_path);
-
     if (!transfer_ctx->output_sink) {
-        return AWS_OP_ERR;
+        fprintf(
+            stderr,
+            "Failure when open file %s with error %s\n",
```
graebm, line 687:
```suggestion
            "Failure to open file %s, error %s\n",
```

source/s3_list_objects.c
graebm, line 145: Have you looked into our other uses of the XML parser, in our real code, to see if we're "properly" using `user_data` to report errors?

graebm: or just do `keep_going = ` instead of this bit nonsense

PR #343, opened 2023-08-21
retry for request skewed

TingDaoK:
*Issue #, if available:*

- Serve as a short term solution for https://github.com/aws/aws-sdk-java-v2/issues/4122

*Description of changes:*

- For long term, we should not let the request sitting in the queue for toooooo long, we need to design a better way, the options I think of:
    -    Move the signing to the connection thread (right before sending the request)?
    -    some way to limits the request to be sent to the queue (Don't sign the request until we have available connection)
    -    Resign the request if we find out the request has been in the queue for too long (server requests less then 15 mins.)


source/s3_auto_ranged_put.c
```diff
@@ -322,7 +317,7 @@ static int s_s3_auto_ranged_put_request_type(struct aws_s3_request *request) {
 
 static struct aws_s3_meta_request_vtable s_s3_auto_ranged_put_vtable = {
     .update = s_s3_auto_ranged_put_update,
-    .send_request_finish = s_s3_auto_ranged_put_send_request_finish,
+    .send_request_finish = aws_s3_meta_request_send_request_finish_default,
```
graebm, line 320: Sorry to comment on pre-existing code, but I get REALLY REALLY confused about the `vtable->send_request_finish()` vs `vtable->finished_request()` callbacks. Can we give them names that make it clear what each is doing? We can do this as a separate PR. Maybe:

request_complete() / request_utterly_complete()
request_complete_p1_maybe_retry()/ request_complete_p2_no_retry()

???

source/s3.c
```diff
@@ -38,6 +38,7 @@ static struct aws_error_info s_errors[] = {
     AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_NON_RECOVERABLE_ASYNC_ERROR, "Async error received from S3 and not recoverable from retry."),
     AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_METRIC_DATA_NOT_AVAILABLE, "The metric data is not available, the requests ends before the metric happens."),
     AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_INCORRECT_CONTENT_LENGTH, "Request body length must match Content-Length header."),
+    AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_REQUEST_TIME_TOO_SKEWED, "Response code indicates the request signed too long ago before server received it."),
```
graebm, line 41: Server rejected request due to RequestTimeTooSkewed

source/s3_copy_object.c
```diff
@@ -60,7 +60,7 @@ static int s_s3_copy_object_request_type(struct aws_s3_request *request) {
 
 static struct aws_s3_meta_request_vtable s_s3_copy_object_vtable = {
     .update = s_s3_copy_object_update,
-    .send_request_finish = aws_s3_meta_request_send_request_finish_handle_async_error,
+    .send_request_finish = aws_s3_meta_request_send_request_finish_default,
```
graebm, line 63: Every vtable.send_request_finish needs to call `aws_s3_meta_request_send_request_finish_default()` now, right?

If so, let's NOT have that be a vtable function we customize for the GET/PUT/DEFAULT/COPY vtables

Is any request-type doing things differently in there? If not, just remove the vtable function altogether, and we don't need to stress about how it's confusing with the other similarly named vtable function.

source/s3_util.c
```diff
@@ -62,9 +62,11 @@ const struct aws_byte_cursor g_user_agent_header_product_name =
     AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("CRTS3NativeClient");
 
 const struct aws_byte_cursor g_s3_internal_error_code = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("InternalError");
-const struct aws_byte_cursor g_s3_slow_down_error_code = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("SlowDown");
+
+const struct aws_byte_cursor s_s3_slow_down_error_code = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("SlowDown");
+const struct aws_byte_cursor s_s3_time_too_skewed = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("RequestTimeTooSkewed");
 /* The special error code as Asynchronous Error Codes */
-const struct aws_byte_cursor g_s3_internal_errors_code = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("InternalErrors");
+const struct aws_byte_cursor s_s3_internal_errors_code = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("InternalErrors");
```
graebm, line 69: trivial: can we put "InternalErrors" next to "InternalError"?

I realize they were already separated, but this change makes them even further apart

or just don't even have these variables at the top of the file, just type `"InternalErrors"` in the 1 place it's actually used in code

source/s3_copy_object.c
graebm: ok, that explains it

How about on_request_complete_pre_retry / on_request_complete_post_retry?

source/s3_auto_ranged_put.c
graebm: resolving this, since it's not really related to the changes in this PR

source/s3_copy_object.c
graebm: whatever

source/s3_meta_request.c
```diff
@@ -1152,40 +1152,86 @@ static void s_s3_meta_request_send_request_finish(
...
+            return false;
+        case AWS_S3_META_REQUEST_TYPE_PUT_OBJECT:
+            if (meta_request->vtable->get_request_type &&
+                meta_request->vtable->get_request_type(request) == AWS_S3_REQUEST_TYPE_COMPLETE_MULTIPART_UPLOAD) {
+                /* Only handle the async error for complete multipart upload. */
+                return true;
+            }
+            return false;
+        default:
+            return false;
```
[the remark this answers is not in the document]
TingDaoK: Let's keep it for now, we will need to update the check later when SDKs have an agreement on the behavior.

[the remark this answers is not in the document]
graebm: I know I've overheard some conversations in the last few weeks about how this behavior applies to more operations than the SDK org realized. I vote we go with:
> 2. default to true, and only return false for requests whose response body is an unmodeled blob (just GetObject ... right?)

source/s3_meta_request.c
```diff
@@ -1183,40 +1183,63 @@ static void s_s3_meta_request_send_request_finish(
...
-            error_code = AWS_ERROR_S3_NON_RECOVERABLE_ASYNC_ERROR;
+        if (error_code_from_body == AWS_ERROR_UNKNOWN || error_code_from_body == AWS_ERROR_SUCCESS) {
+            /* We don't find anything more useful from the body */
+            return error_code_from_status;
         }
-        return error_code;
     }
+    /* Return what  */
+    return error_code_from_body;
 }
```
graebm, line 1237: I still find this hard to follow. Here's my suggestion
```
static int s_s3_meta_request_error_code_from_response(struct aws_s3_request *request) {
    AWS_PRECONDITION(request);

    int error_code_from_status = s_s3_meta_request_error_code_from_response_status(request->send_data.response_status);

    /* Response body might be XML with an <Error><Code> inside.
     * The is very likely when status-code is bad.
     * In some cases, it's even possible after 200 OK. */
    int error_code_from_xml = AWS_ERROR_SUCCESS;
    if (error_code_from_status != AWS_ERROR_SUCCESS || s_should_check_for_error_despite_200_OK(request)) {
        if (request->send_data.response_body.len > 0) {
            /* Attempt to read body as XML, it's fine if this fails. */
            struct aws_byte_cursor xml_doc = aws_byte_cursor_from_buf(&request->send_data.response_body);
            struct aws_byte_cursor error_code_string = {0};
            const char *xml_path[] = {"Error", "Code", NULL};
            if (aws_xml_get_body_at_path(request->allocator, xml_doc, xml_path, &error_code_string) == AWS_OP_SUCCESS) {
                /* Found an <Error><Code> string! Map it to CRT error code. */
                error_code_from_xml = aws_s3_crt_error_code_from_server_error_code_string(error_code_string);
            }
        }
    }

    if (error_code_from_status == AWS_ERROR_SUCCESS) {
        /* Status-code was OK, so assume everything's good, unless we found an <Error><Code> in the XML */
        switch (error_code_from_xml) {
            case AWS_ERROR_SUCCESS:
                return AWS_ERROR_SUCCESS;
            case AWS_ERROR_UNKNOWN:
                return AWS_ERROR_S3_NON_RECOVERABLE_ASYNC_ERROR;
            default:
                return error_code_from_xml;
        }
    } else {
        /* Return error based on status-code, unless we got something more specific from XML */
        switch (error_code_from_xml) {
            case AWS_ERROR_SUCCESS:
                return error_code_from_status;
            case AWS_ERROR_UNKNOWN:
                return error_code_from_status;
            default:
                return error_code_from_xml;
        }
    }
}
```

graebm (approved):
fix & ship

PR #346, opened 2023-08-31
Fix: Get object with checksum leak when retry happens

TingDaoK:
- We allocate buffers and checksum, but in case of error, we never free them.
- The fix is just free them on either case
- TODO: I think we can probably move the checksum stuff into the send_data so it gets clean up with request/response data and make it cleaner.


.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ env:
   AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
   AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
   AWS_REGION: us-east-1
+  CTEST_PARALLEL_LEVEL: 4
```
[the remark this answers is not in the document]
TingDaoK: https://github.com/awslabs/aws-crt-builder/pull/248

tests/mock_s3_server/mock_s3_server.py
```diff
@@ -29,15 +34,42 @@ class S3Opts(Enum):
     ListParts = 6
 
 
-base_dir = os.path.dirname(os.path.realpath(__file__))
+@dataclass
+class Response:
+    path: str
+    disconnect_after_headers = False
+    generate_body_size: Optional[int] = None
+    response_json_path: str = None
```
graebm, line 42: trivial:
```suggestion
    json_path: str = None
```

tests/mock_s3_server/mock_s3_server.py
```diff
@@ -275,11 +300,19 @@ def handle_get_object_modified(start_range, end_range, request):
...
+def handle_get_object(wrapper, request, parsed_path, head_request=False):
+    wrapper.info("handle_get_object")
+    if parsed_path.path == "/get_object_checksum_retry" and not head_request:
+        TrioHTTPWrapper.retry_request_received_continuous = TrioHTTPWrapper.retry_request_received_continuous + 1
 
-def handle_get_object(request, parsed_path):
+        wrapper.info("retry_request_received_continuous is " +
+                     str(TrioHTTPWrapper.retry_request_received_continuous))
+    else:
+        TrioHTTPWrapper.retry_request_received_continuous = 0
```
graebm, line 315: I was trying to suggest a way to use the Response object to communicate between the 2 functions, rather than relying on 1 function setting a singleton that another function checks. I realize you'll need a singleton to count retries, but hoped we use keep its use local to 1 function

Like, you access the singleton here, and use that to set response.fail_after_sending_headers

then in the send_response_from_json() function you do `if response.fail_after_sending_headers:` instead of `if TrioHTTPWrapper.retry_request_received_continuous == 1`

tests/mock_s3_server/mock_s3_server.py
```diff
@@ -29,15 +34,42 @@ class S3Opts(Enum):
...
+                    wrapper.info("Throttling")
+                # Flip the flag
+                TrioHTTPWrapper.should_throttle = not TrioHTTPWrapper.should_throttle
+            self.response_json_path = response_file
 
 
 class TrioHTTPWrapper:
     _next_id = count()
+    retry_request_received_continuous = 0
+    should_throttle = SHOULD_THROTTLE
```
graebm, line 69: you could probably move the throttle variable into the Response class, it's another one where we modify some global state so that something special can happen later on the next response

graebm (commented):
Sorry to get hung up on dumb scripts 😓

tests/mock_s3_server/mock_s3_server.py
TingDaoK: Maybe I didn't get what you mean, but we need a state to control the next response, so, how can we move it to the response? We can add it to the response, but we still need a state on the server to track between requests.

TingDaoK: Oh, I moved it to the response as a static variable.

graebm: we'll talk monday
i am failing to type what i mean

TingDaoK: talked offline and revamped the script

tests/mock_s3_server/mock_s3_server.py
```diff
@@ -264,7 +305,7 @@ def handle_get_object_modified(start_range, end_range, request):
     data_length = end_range - start_range
 
     if start_range == 0:
-        return "/get_object_modified_first_part", data_length, True
+        return ResponseConfig("/get_object_modified_first_part", data_length)
```
graebm, line 308: Since you're just passing positional args, this is going to end up setting `disconnect_after_headers=data_length`
```suggestion
        return ResponseConfig("/get_object_modified_first_part", generate_body_size=data_length)
```

tests/mock_s3_server/mock_s3_server.py
```diff
@@ -351,10 +399,18 @@ async def handle_mock_s3_request(wrapper, request):
         if type(event) is h11.EndOfMessage:
             break
         assert type(event) is h11.Data
+    if response_config is None:
```
graebm, line 402: utterly trivial: space between blocks
```suggestion

    if response_config is None:
```

graebm (approved):
fix & ship

PR #347, opened 2023-09-07
Fix: Hang on file upload after period of inactivity

TingDaoK:
*Issue #, if available:*

https://github.com/awslabs/aws-crt-java/issues/665

*Description of changes:*
- without content-length, we will schedule more requests to prepare than needed. And those extra request will be cleaned up, however, the client level count of `num_requests_being_prepared` will still keep record for those.
- After the dns resolve purge all records due to inactivity, we trigger the part we are not going full speed, https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L1404-L1413, and lead to a hang.


source/s3_client.c
```diff
@@ -1193,27 +1194,9 @@ static void s_s3_client_process_work_default(struct aws_s3_client *client) {
...
-            client->synced_data.num_failed_prepare_requests,
-            &client->threaded_data.num_requests_being_prepared);
-
-        client->synced_data.num_failed_prepare_requests = 0;
-
-        AWS_ASSERT(sub_result == AWS_OP_SUCCESS);
-        (void)sub_result;
-    }
+    aws_atomic_fetch_sub(&client->stats.num_requests_being_prepared, num_requests_queued);
+    aws_atomic_fetch_sub(&client->stats.num_requests_being_prepared, client->synced_data.num_failed_prepare_requests);
```
graebm, line 1198: this is a big change.
We've never depended on the `stats` for behavior before, right?

I believe they've just been used for performance analysis so far. I'm not sure we've been super careful about ensuring their correctness. That's why they're atomics and not properly controlled by a lock, we didn't super care if we printed all the stats and the numbers didn't add up.

DmitriyMusatkin: semi related to this: should we consider just simplifying the current logic? 

we currently increment being_prepared count when we prep requests, then in prep_done callback we push prepped request to queue and update failed numbers and then in the scheduling loop we try to rectify what the number of being prepared should be. all of this happens in critical sections. 

should the logic to dec being_prepared just live in prep_done callback? i.e. always inc when kicking off prep and always dec when prep is done

TingDaoK: yeah, chatted with Mike, the logic will be simpler if just keep the no-op requests in the queue, and clean them up when the client literate through the queue.

The callback can be invoked from another thread, if we decrease counter from the callback, we will need a lock around it, and increase the complexity of the code.

DmitriyMusatkin: callback already has a lock on client. but sure, im ok with removing noops from queue later as well

TingDaoK: yeah, callback has a lock, but we don't have lock to increase the number. https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L1387

tests/s3_tester.c
```diff
@@ -31,6 +31,17 @@
 #    pragma warning(disable : 4232) /* function pointer to dll symbol */
 #endif
 
+/* A function level attribute to disable ThreadSanitizer instrumentation. */
+#if defined(__clang__)
```
DmitriyMusatkin, line 35: https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/macros.h#L116

source/s3_client.c
```diff
@@ -1515,6 +1514,12 @@ void aws_s3_client_update_connections_threaded(struct aws_s3_client *client) {
 
         struct aws_s3_request *request = aws_s3_client_dequeue_request_threaded(client);
         const uint32_t max_active_connections = aws_s3_client_get_max_active_connections(client, request->meta_request);
+        if (request->is_noop) {
+            /* If request is no-op, finishes and cleans up the request */
+            s_s3_client_meta_request_finished_request(client, request->meta_request, request, AWS_ERROR_SUCCESS);
```
graebm, line 1519: are there any lifetime issues, where maybe the meta-request completed (successfully or unsuccessfully) while this no-op request moved from the meta-req-thread to the client thread?

source/s3_client.c
```diff
@@ -1515,6 +1514,12 @@ void aws_s3_client_update_connections_threaded(struct aws_s3_client *client) {
...
+            request = aws_s3_request_release(request);
+            continue;
+        }
 
         /* Unless the request is marked "always send", if this meta request has a finish result, then finish the request
          * now and release it. */
```
graebm, line 1525: trivial, but the code after this is an `if/else if/else if/else` sequence. May as well insert this block into the sequence. vs a standalone `if (no-op) continue` followed by a `if/else if/else if/else` sequence
```suggestion
        } else if (!request->always_send && aws_s3_meta_request_has_finish_result(request->meta_request)) {
            /* Unless the request is marked "always send", if this meta request has a finish result,
             * then finish the request now and release it. */
```

tests/s3_tester.c
```diff
@@ -1335,6 +1335,14 @@ int aws_s3_tester_client_new(
     return AWS_OP_SUCCESS;
 }
 
+/* Disable tsan as we hack into the client threaded data */
```
graebm, line 1338: 😬

DmitriyMusatkin (approved):

source/s3_client.c
TingDaoK: nope, the no-op request keeps the meta-request around until we call `s_s3_client_meta_request_finished_request`. There is an meta-request-level track of how many request was made vs completed.

tests/s3_tester.c
TingDaoK: tried, the client thread is still running, and changing those threaded data from multiple places. 

Maybe better to just suppress tsan

PR #353, opened 2023-09-21
Parallel preparation

TingDaoK:
- Introduce a new API that can handle reading in parallel from it.
- Added an implementation to create parallel input stream from file
- Changed the S3 client to create parallel input stream if the upload source is a file
- Use all the io eventloop group to prepare request if parallel input stream is available

Increased the upload performance if the client is handling less meta request then the number of threads we created for io event. 

For a single upload of a 30GiB file, we increased the speed to read the whole read from 14 secs to around 6 secs by this PR.

Quick thinks:
- Why are we getting 6 secs to read from disk(cached)? Pure reading from cached file with 8 thread can do faster
   - Our client doesn't read the file as fast as possible, **reading from file is probably NOT the bottle neck if we read in parallel and we have the file cached.** 
- Why use 8 threads to read?
   - From my testing, 8 threads can read the file in 236 Gbps, faster than the network bandwidth we have, so, it's a reasonable number to avoid file read as the bottle neck.
   - I tested to use 32 threads to read, I don't see any improvement from our client. The tracing shows the IO threads still worked around 6 secs. Again, I think with parallel read and cached file, the reading is no longer the bottle neck.
   - From my testing with just reading, around 32 threads will be able to reach the best performance. However, even with 1 thread, the reading throughput is higher than what we currently have. 
    ```
    [ec2-user@ip-172-31-14-35 build]$ ./s3-benchrunner-c 1
    Total execution time: 4.76 seconds
    total read_Gib 240
    Read through_put: 50.38 Gbps
    [ec2-user@ip-172-31-14-35 build]$ ./s3-benchrunner-c 8
    Total execution time: 1.01 seconds
    total read_Gib 240
    Read through_put: 236.49 Gbps
    [ec2-user@ip-172-31-14-35 build]$ ./s3-benchrunner-c 16
    Total execution time: 0.56 seconds
    total read_Gib 240
    Read through_put: 424.87 Gbps
    [ec2-user@ip-172-31-14-35 build]$ ./s3-benchrunner-c 32
    Total execution time: 0.51 seconds
    total read_Gib 240
    Read through_put: 471.96 Gbps
    ```
- Will parallel read affect multiple meta requests?
   - I did a test with 24 concurrent upload with the same 30 GiB file. The throughput doesn't change much. From the main branch, I got ~`54Gbps`, from this branch, I got ~`54Gbps` as well. Note: the max-throughput benchmark, I got ~`55Gbps`  

TODO:

- ~Instead of open file descriptor multiple times, can we use `mmap` instead? Will it help to improve the throughput?~ -> https://github.com/awslabs/aws-c-s3/pull/354
- ~research on mmap across platforms, and use case (huge file with limit memory). Limit the memory usage?~ -> https://github.com/awslabs/aws-c-s3/pull/354
- Unknown content-length with parallel read
- Skipping parts should still read the parts to make sure we keep the durability check.


waahm7 (commented):
Haven't reviewed everything yet.

include/aws/s3/private/s3_parallel_read_stream.h
```diff
@@ -0,0 +1,121 @@
...
+     * TODO: error handling:
+     * - The stream reach end before all parts
+     * - Failed to create a elg task to read
+     * - Asyc related stuff as the original async stream (If we cannot fill the dest, we wait?)
+     * - etc.
+     */
+    struct aws_future_bool *(*read)(
+        struct aws_parallel_input_stream *stream,
+        size_t start_position,
+        size_t end_position,
```
graebm, line 48: uint64_t

graebm: or offset/length is common in C apis

we don't need to worry about confusion with `object_range_end` which is confusingly an inclusive number

and we don't need to worry about start <= end

`uint64_t offset; size_t length`?

include/aws/s3/private/s3_parallel_read_stream.h
```diff
@@ -0,0 +1,121 @@
...
+
+AWS_PUSH_SANE_WARNING_LEVEL
+
+struct aws_byte_buf;
+struct aws_future_bool;
+struct aws_input_stream;
+
+struct aws_event_loop_group;
+
+struct aws_parallel_input_stream {
```
graebm, line 21: trivial: struct is named "parallel_input_stream" but file is named "parallel_read_stream"

name the file: s3_parallel_stream.h/c or s3_parallel_input_stream.h/c

include/aws/s3/private/s3_parallel_read_stream.h
```diff
@@ -0,0 +1,121 @@
...
+     * Read into the buffer in parallel.
+     * The implementation needs to support this to be invoked concurrently from multiple threads
+     *
+     * TODO: error handling:
+     * - The stream reach end before all parts
+     * - Failed to create a elg task to read
+     * - Asyc related stuff as the original async stream (If we cannot fill the dest, we wait?)
+     * - etc.
+     */
+    struct aws_future_bool *(*read)(
```
graebm, line 45: why is this async? it does force use to do an extra allocation
why not just synchronous like aws_input_stream?

source/s3_parallel_read_stream.c
```diff
@@ -0,0 +1,240 @@
...
+    struct aws_parallel_input_stream *stream,
+    size_t start_position,
+    size_t end_position,
+    struct aws_byte_buf *dest) {
+    /* Ensure the buffer has space available */
+    if (dest->len == dest->capacity) {
+        struct aws_future_bool *future = aws_future_bool_new(stream->alloc);
+        aws_future_bool_set_error(future, AWS_ERROR_SHORT_BUFFER);
+        return future;
+    }
```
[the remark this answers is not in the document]
graebm: I vote you do all error-checking here

source/s3_parallel_read_stream.c
```diff
@@ -0,0 +1,240 @@
...
+
+struct aws_parallel_input_stream_from_file_impl {
+    struct aws_parallel_input_stream base;
+
+    struct aws_string *file_path;
+    struct aws_event_loop_group *reading_elg;
+    size_t num_workers;
+
+    struct aws_event_loop **assigned_event_loops;
+    struct aws_input_stream **assigned_file_streams;
```
graebm, line 69: this is really complicated.
the meta-request is preparing parts on multiple threads
but then the parallel_stream is also managing its own thread-pool

can we look into just opening the file on-demand and reading from it synchronously?

include/aws/s3/private/s3_parallel_read_stream.h
TingDaoK: async is more generic. I would assume people want to do parallel read from all sorts of things, and some may not be a synchronous call. 

Although, yeah, if we do open the file on needed, we don't need it to be async, but I think make it async is more future prove. 

Also, we can make it async if we need to, but I don't think this will add complexity now, as we already have code path to handle async stream, and well, an extra allocation I guess not really matter

PR #354, opened 2023-10-02
Use mmap to for parallel read streams

TingDaoK:
- A quick internal only impl of mmap wrapper.

Why mmap?
- We can avoid the possible FD limits with parallel read.
- We are not loading the file from disk to memory until needed
- It's "supposed" to be faster

Is mmap really better?
- ~I try to simply memcpy the mapped memory to another buffer, the OS seems to just skip loading the real data, I didn't find a way to easily ensure we load the data.~ It was a bug in my test program, now, I saw it's just the same as multi-thread fread for both cached or cold files.
- From the S3 tests, I am seeing slightly performance improvement with mmap and just fread.
```
mmap 
Run:1 Secs:12.621 Gb/s:20.4 Mb/s:20418.3 GiB/s:2.4 MiB/s:2434.1
Run:2 Secs:8.083 Gb/s:31.9 Mb/s:31880.9 GiB/s:3.7 MiB/s:3800.5
Run:3 Secs:8.686 Gb/s:29.7 Mb/s:29669.5 GiB/s:3.5 MiB/s:3536.9
Run:4 Secs:9.205 Gb/s:28.0 Mb/s:27995.9 GiB/s:3.3 MiB/s:3337.4
Run:5 Secs:8.259 Gb/s:31.2 Mb/s:31201.7 GiB/s:3.6 MiB/s:3719.5
Run:6 Secs:9.032 Gb/s:28.5 Mb/s:28532.3 GiB/s:3.3 MiB/s:3401.3
Run:7 Secs:9.140 Gb/s:28.2 Mb/s:28195.1 GiB/s:3.3 MiB/s:3361.1
Run:8 Secs:7.305 Gb/s:35.3 Mb/s:35278.1 GiB/s:4.1 MiB/s:4205.5
Run:9 Secs:9.986 Gb/s:25.8 Mb/s:25805.1 GiB/s:3.0 MiB/s:3076.2
Run:10 Secs:9.243 Gb/s:27.9 Mb/s:27881.3 GiB/s:3.2 MiB/s:3323.7

fread
Run:1 Secs:12.924 Gb/s:19.9 Mb/s:19940.2 GiB/s:2.3 MiB/s:2377.1
Run:2 Secs:9.986 Gb/s:25.8 Mb/s:25806.7 GiB/s:3.0 MiB/s:3076.4
Run:3 Secs:9.382 Gb/s:27.5 Mb/s:27466.3 GiB/s:3.2 MiB/s:3274.2
Run:4 Secs:10.024 Gb/s:25.7 Mb/s:25707.0 GiB/s:3.0 MiB/s:3064.5
Run:5 Secs:8.134 Gb/s:31.7 Mb/s:31683.2 GiB/s:3.7 MiB/s:3776.9
Run:6 Secs:9.643 Gb/s:26.7 Mb/s:26722.7 GiB/s:3.1 MiB/s:3185.6
Run:7 Secs:9.860 Gb/s:26.1 Mb/s:26136.3 GiB/s:3.0 MiB/s:3115.7
Run:8 Secs:9.008 Gb/s:28.6 Mb/s:28606.2 GiB/s:3.3 MiB/s:3410.1
Run:9 Secs:9.545 Gb/s:27.0 Mb/s:26998.6 GiB/s:3.1 MiB/s:3218.5
Run:10 Secs:8.926 Gb/s:28.9 Mb/s:28869.0 GiB/s:3.4 MiB/s:3441.5
```
- From tracing of the s3 tests, I saw the improvement from 7.5 secs to 6.8 secs, where our IO threads are doing intensive works. But, I would assume it's not really related to the difference between mmap and fread, as they are very likely not the bottle neck of the process.

How will mmap affect our memory usage?
- Needs to test and tracing. `mmap` supports to load the data on needed and we control how many data to read into memory. It "SHOULD" have no affect on our memory usage. However, I am not sure, I'll run some tracing and see the memory usage to make sure.
- Used `top` to track the memory usage, found that OS will load the data into memory and keep it until we unmap the memory. Fixed the memory issue by only map the needed part. The latest change has no effects on memory usage now.

### DIFF
--- a/include/aws/s3/private/aws_mmap.h
+++ b/include/aws/s3/private/aws_mmap.h
@@ -17,10 +17,7 @@ struct aws_allocator;
 
 AWS_EXTERN_C_BEGIN
 
-struct aws_mmap_context {
-    void *impl;
-    const char *content;
-};
+struct aws_mmap_context;
 
 /**
  * TODO: supports different mode and extra.
@@ -32,6 +29,25 @@ AWS_S3_API struct aws_mmap_context *aws_mmap_context_new(struct aws_allocator *a
  * TODO: real refcount.
  */
 AWS_S3_API struct aws_mmap_context *aws_mmap_context_release(struct aws_mmap_context *context);
+
+/**
+ * Get mapped content. OS will load the content into memory when the content is accessed.
+ * You need to invoke `aws_mmap_context_unmap_content` once read from the memory, otherwise, leak will happen
+ *
+ * @param context       The mmap context
+ * @return              The mapped address
+ **/
+AWS_S3_API void *aws_mmap_context_map_content(struct aws_mmap_context *context);
+
+/**
+ * Remove any mappings for those entire pages containing any part of the address space of the process starting at addr
+ * and continuing for len bytes
+ *
+ * @param mapped_addr   The start address of the mapped content
+ * @param len           The length to free
+ * @return AWS_OP_SUCCESS on succes, AWS_OP_ERR on error.
+ **/
+AWS_S3_API int aws_mmap_context_unmap_content(void *mapped_addr, size_t len);
 
 AWS_EXTERN_C_END
 #endif /* AWS_MMAP_H */
--- a/include/aws/s3/private/aws_mmap.h
+++ b/include/aws/s3/private/aws_mmap.h
@@ -7,6 +7,10 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
+/**
+ * TODO: move this to part of <aws/common/file.h> instead
+ */
+
 #include <aws/s3/s3.h>
 
 struct aws_allocator;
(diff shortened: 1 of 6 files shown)


include/aws/s3/private/aws_mmap.h
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Get mapped content. OS will load the content into memory when the content is accessed.
+ * You need to invoke `aws_mmap_context_unmap_content` on the `out_start_addr` once read from the memory is done,
+ * otherwise, leak will happen
+ *
+ * @param context           The mmap context
+ * @param length            The length of memory to map
+ * @param offset            The offset starting at the file to map.
+ * @param out_start_addr    The start address for the mapped memory, it will be a multiple
+ *                          of the page size as returned by sysconf(_SC_PAGE_SIZE)
```
graebm, line 42: I guess the dream of making the codebase less complex didn't pan out. Nor is it getting any better performance. So all this really brings is that we don't have to worry about running out of file descriptors.

Have the pure-Java transfer managers every run into issues with file-descriptor limits? If it's never really been an issue for them, then maybe we should go that route, for simplicity's sake. We can always pull this mmap stuff out of our back pockets if customers report they're hitting those limits

PR #355, opened 2023-10-05
Parallel prepare - open file on needed

TingDaoK:
Parallel preparation, the parallel stream will open the file when reading from it and seek to the right offset to read, and close the file after reading.

We checked it has no performance impact with mmap impl, and it's simpler. Simpler is better

With this approach, the user of the parallel input stream will have to manage the threads and concurrence, but it's only used by our s3 client, which already handles the concurrence and thread pool, we are good.

Original: https://github.com/awslabs/aws-c-s3/pull/353

- Follow up PR for keep reading the parts from resume: https://github.com/awslabs/aws-c-s3/pull/356


source/s3_parallel_input_stream.c
```diff
@@ -0,0 +1,165 @@
...
+        AWS_CONTAINER_OF(stream, struct aws_parallel_input_stream_from_file_impl, base);
+
+    aws_string_destroy(impl->file_path);
+
+    aws_mem_release(stream->alloc, impl);
+
+    return;
+}
+
+static int s_get_last_modified_time(const char *file_name, uint64_t *out_time) {
```
graebm, line 80: utterly trivial: if you just passed it as aws_string here, you wouldn't have to convert every time you called this function
```suggestion
static int s_get_last_modified_time(struct aws_string *file_name, uint64_t *out_time) {
```

source/s3_parallel_input_stream.c
```diff
@@ -0,0 +1,165 @@
...
+        aws_ref_count_acquire(&stream->ref_count);
+    }
+    return stream;
+}
+
+struct aws_parallel_input_stream *aws_parallel_input_stream_release(struct aws_parallel_input_stream *stream) {
+    if (stream != NULL) {
+        aws_ref_count_release(&stream->ref_count);
+    }
+    return stream;
```
graebm, line 44:
```suggestion
    return NULL;
```

source/s3_parallel_input_stream.c
```diff
@@ -0,0 +1,165 @@
...
+    struct aws_parallel_input_stream *stream,
+    uint64_t offset,
+    struct aws_byte_buf *dest) {
+    /* Ensure the buffer has space available */
+    if (dest->len == dest->capacity) {
+        struct aws_future_bool *future = aws_future_bool_new(stream->alloc);
+        aws_future_bool_set_error(future, AWS_ERROR_SHORT_BUFFER);
+        return future;
+    }
+    /* TODO: restore the buffer on failure. */
```
graebm, line 57: It's much harder to pull this off with an async API. I'd just delete this TODO
```suggestion
```

source/s3_parallel_input_stream.c
```diff
@@ -0,0 +1,165 @@
...
+
+static void s_para_from_file_destroy(struct aws_parallel_input_stream *stream) {
+    struct aws_parallel_input_stream_from_file_impl *impl =
+        AWS_CONTAINER_OF(stream, struct aws_parallel_input_stream_from_file_impl, base);
+
+    aws_string_destroy(impl->file_path);
+
+    aws_mem_release(stream->alloc, impl);
+
+    return;
```
graebm, line 77: trivial: but why tho
```suggestion
```

source/s3_parallel_input_stream.c
```diff
@@ -0,0 +1,165 @@
...
+struct aws_parallel_input_stream_from_file_impl {
+    struct aws_parallel_input_stream base;
+
+    struct aws_string *file_path;
+    uint64_t last_modified_time;
+};
+
+static void s_para_from_file_destroy(struct aws_parallel_input_stream *stream) {
+    struct aws_parallel_input_stream_from_file_impl *impl =
+        AWS_CONTAINER_OF(stream, struct aws_parallel_input_stream_from_file_impl, base);
```
graebm, line 71: trivial:
```suggestion
    struct aws_parallel_input_stream_from_file_impl *impl = stream->impl;
```

source/s3_parallel_input_stream.c
```diff
@@ -0,0 +1,165 @@
...
+}
+
+struct aws_future_bool *s_para_from_file_read(
+    struct aws_parallel_input_stream *stream,
+    uint64_t offset,
+    struct aws_byte_buf *dest) {
+
+    struct aws_future_bool *future = aws_future_bool_new(stream->alloc);
+    struct aws_parallel_input_stream_from_file_impl *impl =
+        AWS_CONTAINER_OF(stream, struct aws_parallel_input_stream_from_file_impl, base);
```
graebm, line 98: trivial
```suggestion
    struct aws_parallel_input_stream_from_file_impl *impl = stream->impl;
```

source/s3_parallel_input_stream.c
```diff
@@ -0,0 +1,165 @@
...
+    /* TODO: restore the buffer on failure. */
+    struct aws_future_bool *future = stream->vtable->read(stream, offset, dest);
+    return future;
+}
+
+struct aws_parallel_input_stream_from_file_impl {
+    struct aws_parallel_input_stream base;
+
+    struct aws_string *file_path;
+    uint64_t last_modified_time;
```
graebm, line 66: highly debatable: ensuring that the file remains unchanged is .... honorable. But like ... that's user error man. And I get the feeling supporting this cross-platform could be annoying

Unless Java says they do this, because of some real reason, I'd just keep it simple and delete this
```suggestion
```

source/s3_parallel_input_stream.c
```diff
@@ -0,0 +1,165 @@
...
+    /* Check if file modified after we create the input stream */
+    if (last_modified_time != impl->last_modified_time) {
+        aws_raise_error(AWS_ERROR_S3_FILE_MODIFIED);
+        goto done;
+    }
+    file_stream = aws_input_stream_new_from_file(stream->alloc, aws_string_c_str(impl->file_path));
+    if (aws_input_stream_seek(file_stream, offset, AWS_SSB_BEGIN)) {
+        goto done;
+    }
+    /* Keep reading until fill the buffer */
```
graebm, line 119: This code is good. I just wondered if we should be checking for EOF before doing the read(), but googled it and [apparently not](https://stackoverflow.com/questions/17263830/fseek-on-a-position-beyond-eof-does-not-trigger-eof-using-feof-how-come)!

Maybe worth adding a note for anyone else who wonders the same?
```suggestion
    /* Keep reading until fill the buffer.
     * Note that we must read() after seek() to determine if we're EOF, the seek alone won't trigger it. */
```

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -153,8 +156,9 @@ struct aws_s3_meta_request {
 
     struct aws_s3_endpoint *endpoint;
 
-    /* Event loop to schedule IO work related on, ie, reading from streams, streaming parts back to the caller, etc...
-     * After the meta request is finished, this will be reset along with the client reference.*/
+    /* Event loop to schedule IO work related on and requires to be serial, ie, reading from non-parallel streams,
+     * streaming parts back to the caller, etc... After the meta request is finished, this will be reset along with the
+     * client reference.*/
```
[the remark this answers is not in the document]
TingDaoK: I was trying to edit because for preparation in parallel now happens across different eventloops, instead of the only one assigned with the request.

source/s3_parallel_input_stream.c
```diff
@@ -0,0 +1,165 @@
...
+    aws_string_destroy(impl->file_path);
+
+    aws_mem_release(stream->alloc, impl);
+
+    return;
+}
+
+static int s_get_last_modified_time(const char *file_name, uint64_t *out_time) {
+    struct stat attrib;
+    if (stat(file_name, &attrib)) {
```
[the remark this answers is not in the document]
TingDaoK: well it does work with our CI at least.

source/s3_parallel_input_stream.c
TingDaoK: They do check the file last modified time. https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java#L397

I added it because I saw them doing it. I can check with them if there are real reasons behind.

TingDaoK: well, checked the PR added it. https://github.com/aws/aws-sdk-java-v2/pull/2796

They don't have a real ask for this, was trying to be nice to give more info about checksum mismatch

graebm: ok, fair

source/s3_parallel_input_stream.c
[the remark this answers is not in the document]
graebm: This is super fishy.
All googling leads me here: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=msvc-170

where they put underscores on it like "_stat"

tests/s3_tester.c
```diff
@@ -1548,35 +1547,7 @@ int aws_s3_tester_send_meta_request_with_options(
             /* if uploading via filepath, write input_stream out as tmp file on disk, and then upload that */
             if (options->put_options.file_on_disk) {
```
graebm, line 1548: We don't have any tests for aws_parallel_input_stream_read() failing...

but I guess that's hard to pull off, since the public aws_s3_meta_request_options don't actually take a aws_parallel_input_stream... so we can't make a fake one that fails the Nth time you call read() (like we have for [input-stream](https://github.com/awslabs/aws-c-io/blob/main/include/aws/testing/stream_tester.h) and [async-stream](https://github.com/awslabs/aws-c-io/blob/main/include/aws/testing/async_stream_tester.h)

huh

we could have aws_s3_meta_request_options take an `struct aws_parallel_input_stream *send_parallel_stream`, but marked like `/* FOR TESTING ONLY */` ??

TingDaoK: GOOD CALLOUT on the the failure case, it caught the bug as we are preparing parts out of data, and we set the part info in the list based on index. So, that if the later parts finish before the early parts and something went wrong, the part info list will have garbage in it.

graebm (approved):

PR #356, opened 2023-10-05
instead of skipping, let's prepare the parts and mark it noop

TingDaoK:
- Refactoring the resume logic to instead of skipping parts during preparing other parts, we prepare those parts already uploaded, and then make it noop. After we read the part, we check the checksum and length.

### TODO:
Maybe, we can be more smarter? If the part uploaded doesn't have a checksum, we can skip reading the part? Or let user config that if they want us to read the parts uploaded.


source/s3_auto_ranged_put.c
```diff
@@ -518,24 +508,17 @@ static bool s_s3_auto_ranged_put_update(
...
-                    if (part != NULL) {
-                        /* part already downloaded, skip it here and prepare will take care of adjusting the buffer */
-                        ++auto_ranged_put->threaded_update_data.next_part_number;
+                size_t part_index = auto_ranged_put->threaded_update_data.next_part_number - 1;
 
-                    } else {
-                        // incomplete part found. break out and create request for it.
-                        break;
-                    }
+                struct aws_s3_mpu_part_info *part = NULL;
```
DmitriyMusatkin, line 517: minor: maybe add field to part info to indicate that we got info for that part from list parts, i.e. it was previously uploaded. relying on part being present in the list seems a bit fragile of a logic

source/s3_auto_ranged_put.c
```diff
@@ -802,184 +788,6 @@ static int s_verify_part_matches_checksum(
     return return_status;
 }
 
-/* Data for the async skip-parts-from-stream job */
```
DmitriyMusatkin, line 805: +1 on removing all this code

DmitriyMusatkin (approved):

source/s3_auto_ranged_put.c
```diff
@@ -1206,9 +990,15 @@ static void s_s3_prepare_upload_part_on_read_done(void *user_data) {
         --auto_ranged_put->synced_data.num_parts_pending_read;
 
         auto_ranged_put->synced_data.is_body_stream_at_end = is_body_stream_at_end;
+        if (request->uploaded_before_resume) {
```
graebm, line 993: 1) remove the `{ }` block around the critical section...
2) ... so we can move the part_info declaration close to where we set it
```suggestion
    struct aws_s3_mpu_part_info *previously_uploaded_info = NULL;
    if (request->was_previously_uploaded) {
        aws_array_list_get_at(
                &auto_ranged_put->synced_data.part_list, &previously_uploaded_info, request->part_number - 1);
```
those blocks are just to be decorative and make the critical section stand out more. But now it's making the code harder to read

source/s3_auto_ranged_put.c
```diff
@@ -1220,6 +1010,27 @@ static void s_s3_prepare_upload_part_on_read_done(void *user_data) {
         aws_s3_meta_request_unlock_synced_data(meta_request);
     }
     /* END CRITICAL SECTION */
+    if (uploaded_part_info) {
+        /* Check for the durability */
```
graebm, line 1014: Adding explainer about why it's safe to touch this after dropping the lock (copy pasted from when we did a similar thing in the skipping logic)
```suggestion
    /* END CRITICAL SECTION */
    
    if (previously_uploaded_part) {
        /* Part was previously uploaded, check that it matches what we just read.
         * (Yes it's weird that we keep a pointer to the part_info even after
         * releasing the lock that protects part_list. But it's the resizable
         * part_list that needs lock protection. A previously uploaded part_info is const,
         * and it's on the heap, so it's safe to keep the pointer around)
```

source/s3_auto_ranged_put.c
```diff
@@ -1220,6 +1010,27 @@ static void s_s3_prepare_upload_part_on_read_done(void *user_data) {
...
+        if (request->request_body.capacity != uploaded_part_info->size) {
```
graebm, line 1015: it's weird to compare capacity, when you could just use length
```suggestion
        if (request->request_body.len != uploaded_part_info->size) {
```

source/s3_auto_ranged_put.c
```diff
@@ -1412,27 +1171,22 @@ static int s_s3_review_multipart_upload(struct aws_s3_request *request) {
...
+    struct aws_allocator *allocator = request->allocator;
 
-    if (error_code != AWS_ERROR_SUCCESS) {
-        aws_future_http_message_set_error(complete_mpu_prep->on_complete, error_code);
-        goto on_done;
-    }
+    struct aws_future_http_message *message_future = aws_future_http_message_new(allocator);
 
     AWS_FATAL_ASSERT(auto_ranged_put->upload_id);
+    int error_code = AWS_OP_SUCCESS;
```
graebm, line 1183: don't need this anymore, since we're not pulling an error_code out of a future anymore
```suggestion
```

source/s3_auto_ranged_put.c
```diff
@@ -1412,27 +1171,22 @@ static int s_s3_review_multipart_upload(struct aws_s3_request *request) {
...
 
     if (request->num_times_prepared == 0) {
         /* Invoke upload_review_callback, and fail meta-request if user raises an error */
         if (s_s3_review_multipart_upload(request) != AWS_OP_SUCCESS) {
             error_code = aws_last_error();
-            aws_future_http_message_set_error(complete_mpu_prep->on_complete, error_code);
+            aws_future_http_message_set_error(message_future, error_code);
```
graebm, line 1189: we can get rid of error_code tmp
```
            aws_future_http_message_set_error(message_future, aws_last_error());
```

source/s3_auto_ranged_put.c
```diff
@@ -82,8 +80,6 @@ static bool s_s3_auto_ranged_put_update(
     uint32_t flags,
```
graebm, line 80: A few lines up, in `struct aws_s3_prepare_complete_multipart_upload_job`, delete this member:
```
/* async step: skip remaining parts from input stream that were previously uploaded */
struct aws_future_void *asyncstep_skip_remaining_parts;
```

graebm (approved):

PR #359, opened 2023-10-23
Pool buffers for upload

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Current logic of recreating buffers for every part results in a lot of memory acquire/free cycles. since each meta request will have the same sized buffers for each part, just keep them around and reuse

### DIFF
--- a/include/aws/s3/private/s3_auto_ranged_put.h
+++ b/include/aws/s3/private/s3_auto_ranged_put.h
@@ -45,6 +45,8 @@ struct aws_s3_auto_ranged_put {
          * Throughout codebase 0 based part numbers are usually referred to as part index.
          */
         uint32_t next_part_number;
+
+        struct aws_array_list buffer_pool;
     } threaded_update_data;
 
     /* Members to only be used when the mutex in the base type is locked. */
--- a/source/s3_auto_ranged_put.c
+++ b/source/s3_auto_ranged_put.c
@@ -102,6 +102,34 @@ static int s_s3_auto_ranged_put_pause(
     struct aws_s3_meta_request *meta_request,
     struct aws_s3_meta_request_resume_token **resume_token);
 
+/**
+ * Helper to compute request body size.
+ * Basically returns either part size or if content is not equally divisible into parts, the size of the remaining last
+ * part.
+ */
+static size_t s_compute_request_body_size(
+    const struct aws_s3_meta_request *meta_request,
+    uint32_t part_number,
+    uint64_t *offset_out) {
+    AWS_PRECONDITION(meta_request);
+
+    const struct aws_s3_auto_ranged_put *auto_ranged_put = meta_request->impl;
+
+    size_t request_body_size = meta_request->part_size;
+    /* Last part--adjust size to match remaining content length. */
+    if (auto_ranged_put->has_content_length && part_number == auto_ranged_put->total_num_parts_from_content_length) {
+        size_t content_remainder = (size_t)(auto_ranged_put->content_length % (uint64_t)meta_request->part_size);
+
+        if (content_remainder > 0) {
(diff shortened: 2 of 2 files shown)


source/s3_auto_ranged_put.c
```diff
@@ -102,6 +102,34 @@ static int s_s3_auto_ranged_put_pause(
     struct aws_s3_meta_request *meta_request,
     struct aws_s3_meta_request_resume_token **resume_token);
 
+/**
+ * Helper to compute request body size.
+ * Basically returns either part size or if content is not equally divisible into parts, the size of the remaining last
+ * part.
+ */
+static size_t s_compute_request_body_size(
```
TingDaoK, line 110: Trivial: Why are we moving this function around? I don't see any change made?

include/aws/s3/private/s3_auto_ranged_put.h
```diff
@@ -45,6 +45,8 @@ struct aws_s3_auto_ranged_put {
          * Throughout codebase 0 based part numbers are usually referred to as part index.
          */
         uint32_t next_part_number;
+
+        struct aws_array_list buffer_pool;
```
TingDaoK, line 49: Seems like we should use linked_list instead as we keep push/pop.

source/s3_auto_ranged_put.c
DmitriyMusatkin: moved it back. logic was simplified and i no longer need it at the start

include/aws/s3/private/s3_auto_ranged_put.h
graebm: array_list is fine if this thing is only a few dozen in size. simpler to use because we don't need to allocate a unique type with a node in it

source/s3_auto_ranged_put.c
```diff
@@ -547,6 +585,14 @@ static bool s_s3_auto_ranged_put_update(
 
                 request->part_number = auto_ranged_put->threaded_update_data.next_part_number;
 
+                AWS_ZERO_STRUCT(request->request_body);
```
graebm, line 588: trivial
```suggestion
                /* If there's a buffer in the pool, grab it */
                AWS_ZERO_STRUCT(request->request_body);
```

source/s3_auto_ranged_put.c
```diff
@@ -924,7 +942,9 @@ struct aws_future_http_message *s_s3_prepare_upload_part(struct aws_s3_request *
         /* Read the body */
         uint64_t offset = 0;
         size_t request_body_size = s_compute_request_body_size(meta_request, request->part_number, &offset);
-        aws_byte_buf_init(&request->request_body, meta_request->allocator, request_body_size);
+        if (request->request_body.capacity == 0) {
+            aws_byte_buf_init(&request->request_body, meta_request->allocator, request_body_size);
```
graebm, line 946:
```suggestion
        /* If we didn't get a buffer from the pool, then it needs to be initialized */
        if (request->request_body.capacity == 0) {
            AWS_ASSERT(request->request_body.capacity >= request_body_size);
            aws_byte_buf_init(&request->request_body, meta_request->allocator, request_body_size);
```

source/s3_auto_ranged_put.c
```diff
@@ -1545,6 +1565,14 @@ static void s_s3_auto_ranged_put_request_finished(
                     }
                 }
 
+                if (meta_request->part_size == request->request_body.capacity) {
+                    /* Steal the buffer from request for reuse. */
```
graebm, line 1569:
```suggestion
                    /* Give buffer to pool, so future parts can re-use it */
```

graebm (approved):

PR #364, opened 2023-10-26
Populate error payload for HeadObject request failure

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-crt-java/issues/698

*Description of changes:*
S3 does not send an error explanation in the body if the HEAD_OBJECT request fails. Populate the `error_response_body` with aws_http_status_text to have something reasonable instead of an empty error_response_body


tests/s3_data_plane_tests.c
```diff
@@ -4225,6 +4225,38 @@ static int s_test_s3_get_object_fail_headers_callback(struct aws_allocator *allo
...
+            {
+                .object_path = aws_byte_cursor_from_c_str("/does_not_exist"),
+            },
+    };
+
+    ASSERT_SUCCESS(aws_s3_tester_send_meta_request_with_options(NULL, &options, &meta_request_test_results));
+    ASSERT_INT_EQUALS(meta_request_test_results.finished_error_code, AWS_ERROR_S3_INVALID_RESPONSE_STATUS);
+    struct aws_byte_cursor expected_error = aws_byte_cursor_from_c_str("Not Found");
+
+    ASSERT_BIN_ARRAYS_EQUALS(
```
DmitriyMusatkin, line 4250: suggestion: get cursor from buffer and then use ASSERT_CURSOR_VALUE_CSTRING_EQUALS

DmitriyMusatkin (approved):

source/s3_meta_request.c
```diff
@@ -1764,21 +1764,33 @@ bool aws_s3_meta_request_body_has_no_more_data(const struct aws_s3_meta_request
...
 
-        if (request->send_data.response_body.capacity > 0) {
+        if (failed_request->send_data.response_body.capacity > 0) {
             result->error_response_body = aws_mem_calloc(meta_request->allocator, 1, sizeof(struct aws_byte_buf));
 
             aws_byte_buf_init_copy(
-                result->error_response_body, meta_request->allocator, &request->send_data.response_body);
+                result->error_response_body, meta_request->allocator, &failed_request->send_data.response_body);
+        } else if (
+            failed_request->request_tag == AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_HEAD_OBJECT &&
```
TingDaoK, line 1783: do not use `AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_HEAD_OBJECT` directly.

For different type of meta request, the `request_tag` may result in the same enum value, but have different meaning. EG: only for auto_ranged_get, 0 means HEAD_OBJECT, for auto_ranged_put, 0 means List_part. 

Use `get_request_type` vfunction on the request to get the `enum aws_s3_request_type ` that shared across different meta request type instead.

TingDaoK: Or double check your meta request type to make sure it's a auto_ranged_get

PR #368, opened 2023-11-09
(WIP) Mem limiter

DmitriyMusatkin:
*Issue #, if available:*
https://github.com/aws/aws-sdk-java-v2/issues/4034

*Description of changes:*


source/s3_buffer_pool.c
```diff
@@ -0,0 +1,225 @@
...
+
+    buffer_pool->base_allocator = allocator;
+    buffer_pool->block_size = block_size;
+    buffer_pool->current_mem_usage = 0;
+    buffer_pool->max_mem_usage = max_mem_usage;
+    if (aws_mutex_init(&buffer_pool->mutex)) {
+        goto on_error;
+    }
+
+    aws_array_list_init_dynamic(&buffer_pool->blocks, allocator, s_min_num_blocks, sizeof(struct s3_buffer_pool_block));
```
JonathanHenson, line 73: this is nice since IIRC this will put all blocks on the same allocation.... Now just need to make sure this stays on the same numa node at some point.

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    aws_array_list_clean_up(&buffer_pool->blocks);
+
+    aws_mutex_clean_up(&buffer_pool->mutex);
+    struct aws_allocator *base = buffer_pool->base_allocator;
+    aws_mem_release(base, buffer_pool);
+}
+
+void aws_s3_buffer_pool_trim(struct aws_s3_buffer_pool *buffer_pool) {
```
JonathanHenson, line 102: I'm surprised this doesn't need a mutex?

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,225 @@
...
+    aws_mutex_unlock(&buffer_pool->mutex);
+
+    return (struct aws_s3_pooled_buffer){.ptr = alloc_ptr, .size = size};
+}
+
+struct aws_s3_pooled_buffer aws_s3_buffer_pool_acquire_buffer(struct aws_s3_buffer_pool *buffer_pool, size_t size) {
+    AWS_PRECONDITION(buffer_pool);
+
+    if (size < buffer_pool->block_size) {
+        return s_primary_acquire(buffer_pool, size);
```
JonathanHenson, line 169: the locking in this function seems error prone. Maybe lock in this function rather than s_primary_acquire?

source/s3_client.c
```diff
@@ -1193,8 +1313,18 @@ static void s_s3_client_process_work_default(struct aws_s3_client *client) {
...
+    s_s3_client_schedule_buffer_pool_trim_synced(client);
+
     aws_linked_list_swap_contents(&meta_request_work_list, &client->synced_data.pending_meta_request_work);
 
+    while (aws_priority_queue_size(&client->synced_data.requests_waiting_for_mem) > 0) {
+        struct aws_s3_request *request = NULL;
+        aws_priority_queue_pop(&client->synced_data.requests_waiting_for_mem, (void **)&request);
+        AWS_FATAL_ASSERT(request != NULL);
+
+        aws_priority_queue_push(&client->synced_data.requests_waiting_for_mem, &request);
```
JonathanHenson, line 1325: how is this not an infinite loop?

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,225 @@
...
+}
+
+void aws_s3_buffer_pool_trim(struct aws_s3_buffer_pool *buffer_pool) {
+    size_t swap_count = 0;
+
+    for (size_t i = 0; i < aws_array_list_length(&buffer_pool->blocks); ++i) {
+        struct s3_buffer_pool_block *block;
+        aws_array_list_get_at_ptr(&buffer_pool->blocks, (void **)&block, i);
+
+        if (block->block_ptr == 0) {
```
TingDaoK, line 109: should this be `block->alloc_count == 0` instead?

include/aws/s3/private/s3_buffer_pool.h
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+AWS_EXTERN_C_BEGIN
+
+struct aws_s3_buffer_pool;
+
+struct aws_s3_pooled_buffer {
+    size_t size;
+    uint8_t *ptr;
+};
```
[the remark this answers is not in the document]
DmitriyMusatkin: My reasoning for a new structure is that it makes it obvious that we should clean it up on its own - it felt really weird passing regular buf or cursor back to buffer pool.

i put struct into header since its private and its easier to deal with it by value, but users should not care about whats in it. and later on we can add fancier things to it for making pooling easier, like which block it came from

source/s3_buffer_pool.c
DmitriyMusatkin: yes, thats a bug in trimming. fixed

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,225 @@
...
+    aws_array_list_init_dynamic(&buffer_pool->blocks, allocator, s_min_num_blocks, sizeof(struct s3_buffer_pool_block));
+
+    return buffer_pool;
+
+on_error:
+    aws_mem_release(allocator, buffer_pool);
+    return NULL;
+}
+
+void aws_s3_buffer_pool_destroy(struct aws_s3_buffer_pool *buffer_pool) {
```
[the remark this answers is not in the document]
DmitriyMusatkin: its private only right now and lifetime is tied to client, so for simplicity i just went with new/destroy. but sure in future if we want to expose it, we can do acquire/release

source/s3_buffer_pool.c
DmitriyMusatkin: it does. that was a miss. it used to be a helper than always ran within a mutex. will fix.

source/s3_buffer_pool.c
DmitriyMusatkin: I pushed the allocations to prepare stage, so it will run from io prep threads. 
so in practice it can allocate from different numa nodes for bigger blocks, but then same block can be reused by different nodes. 
In future we can tie blocks to specific numa nodes, something as trivial as keeping a list of blocks per numa node will work

source/s3_client.c
```diff
@@ -134,6 +161,18 @@ void aws_s3_set_dns_ttl(size_t ttl) {
...
+static int s_s3_request_waiting_for_mem_priority_queue_pred(const void *a, const void *b) {
+    const struct aws_s3_request *const *request_a = a;
+    AWS_PRECONDITION(request_a);
+    AWS_PRECONDITION(*request_a);
+
+    const struct aws_s3_request *const *request_b = b;
+    AWS_PRECONDITION(request_b);
+    AWS_PRECONDITION(*request_b);
+
+    return (*request_a)->num_times_tried_buffer_acquire > (*request_b)->num_times_tried_buffer_acquire;
```
graebm, line 173: can we prioritize certain requests like CancelMultipartUpload? Isn't there already some high-priority-request concept? AWS_S3_REQUEST_FLAG_ALWAYS_SEND?

include/aws/s3/private/s3_buffer_pool.h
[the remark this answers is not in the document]
graebm: trivial: aws_byte_cursor seems fine to me

include/aws/s3/private/s3_buffer_pool.h
```diff
@@ -0,0 +1,96 @@
...
+/*
+ * Acquire buffer of specified size.
+ * Returned buffer is empty (0 ptr and size) if buffer cannot be allocated.
+ * Warning: Always release buffer using release api. Its fine to wrap pooled
+ * buffer in another struct as long as it does not try to free underlying pointer
+ * (ex. aws_byte_buf_from_pooled_buffer)
+ */
+AWS_S3_API struct aws_s3_pooled_buffer aws_s3_buffer_pool_acquire_buffer(
+    struct aws_s3_buffer_pool *buffer_pool,
+    size_t size);
```
graebm, line 68:
```suggestion
/*
 * Try to acquire buffer of specified size.
 * Returned buffer is empty (0 ptr and size) if buffer cannot be allocated.
 * Warning: Always release buffer using release api. Its fine to wrap pooled
 * buffer in another struct as long as it does not try to free underlying pointer
 * (ex. aws_byte_buf_from_pooled_buffer)
 */
AWS_S3_API struct aws_s3_pooled_buffer aws_s3_buffer_pool_try_acquire_buffer(
    struct aws_s3_buffer_pool *buffer_pool,
    size_t size);
```

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -128,6 +128,7 @@ struct aws_s3_meta_request_vtable {
  */
 struct aws_s3_meta_request {
     struct aws_allocator *allocator;
+    struct aws_s3_buffer_pool *buffer_pool;
```
graebm, line 131: is this just an ergonomic thing? I'd prefer we be clear about ownership and just do `request->client->buffer_pool`?
```suggestion
```

graebm (commented):
I didn't get very far, so some comments may be uninformed. I'll continue tomorrow...

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,228 @@
...
+    for (size_t i = 0; i < aws_array_list_length(&buffer_pool->blocks); ++i) {
+        struct s3_buffer_pool_block *block;
+        aws_array_list_get_at_ptr(&buffer_pool->blocks, (void **)&block, i);
+
+        if (block->alloc_count == 0) {
+            block->offset_ptr = block->block_ptr + size;
+            alloc_ptr = block->block_ptr;
+            block->alloc_count += 1;
+            break;
+        }
```
graebm, line 138: trivial/debatable: you could either eliminate this if-block if you reset the offset_ptr from the release() function.

or you could just have this if-block reset the offset_ptr, and let the next if-block handle everything else

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,228 @@
...
+    }
+    aws_mutex_unlock(&buffer_pool->mutex);
+
+    return (struct aws_s3_pooled_buffer){.ptr = alloc_ptr, .size = alloc_ptr == NULL ? 0 : size};
+}
+
+void aws_s3_buffer_pool_release_buffer(
+    struct aws_s3_buffer_pool *buffer_pool,
+    struct aws_s3_pooled_buffer pooled_buffer) {
+    if (buffer_pool == NULL || pooled_buffer.ptr == NULL || pooled_buffer.size == 0) {
```
graebm, line 192: I don't see us doing a size == 0 check in acquire, so if somehow someone passed size 0 we'd end up with mismatched refcounts

either eliminate this size == 0 check, or add one in acquire()

source/s3_auto_ranged_get.c
```diff
@@ -180,6 +180,7 @@ static bool s_s3_auto_ranged_get_update(
                             AWS_S3_REQUEST_FLAG_RECORD_RESPONSE_HEADERS | AWS_S3_REQUEST_FLAG_PART_SIZE_RESPONSE_BODY);
 
                         request->discovers_object_size = true;
+                        request->part_size = meta_request->part_size;
```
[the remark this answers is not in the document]
DmitriyMusatkin: Fixed. i this this was oversight originally.

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,251 @@
...
+        (void *)alloc_ptr, size, buffer_pool->current_alloc_usage);
+        buffer_pool->current_alloc_usage += size;
+    }
+
+    aws_mutex_unlock(&buffer_pool->mutex);
+
+    return (struct aws_s3_pooled_buffer){.ptr = alloc_ptr, .size = alloc_ptr == NULL ? 0 : size};
+}
+
+void aws_s3_buffer_pool_release_buffer(
```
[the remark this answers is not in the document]
DmitriyMusatkin: yeah, trimming itself is expensive to do on every release, but one thing im considering is when number of active requests in flight in scheduling falls to relatively low number, run a trim and get rid of blocks that have not been used for a while

include/aws/s3/private/s3_meta_request_impl.h
graebm: debatable/trivial: still think `request->client->buffer_pool` is better than redundant pointer (extra variable makes ownership unclear, and it's one more thing that might get out of sync, or forget to set)

but also whatever

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,319 @@
...
+            buffer_pool->primary_reserved += size;
+        } else {
+            buffer_pool->secondary_reserved += size;
+        }
+    }
+
+    aws_mutex_unlock(&buffer_pool->mutex);
+
+    if (ticket == NULL) {
+        aws_raise_error(AWS_ERROR_S3_INSUFFICIENT_MEMORY);
```
graebm, line 196: AWS_LOGF_ERROR about the request exceeding the memory limit, so they know what to change to fix it

Should the error be something like `AWS_ERROR_S3_EXCEEDS_MEMORY_LIMIT`? I fear ~INSUFFICIENT_MEMORY~ sounds too much like OOM?

source/s3_auto_ranged_get.c
```diff
@@ -196,6 +196,15 @@ static bool s_s3_auto_ranged_get_update(
                     request->part_range_end = meta_request->part_size - 1; /* range-end is inclusive */
                     request->discovers_object_size = true;
 
+                    struct aws_s3_buffer_pool_ticket *ticket 
+                        = aws_s3_buffer_pool_reserve(meta_request->client->buffer_pool, meta_request->part_size);
+
+                    if (ticket == NULL) {
+                        goto has_work_remaining;
```
graebm, line 203: If we have simultaneous meta-requests with different-sized parts, is there an issue where the pool will start rejecting large parts, but still let small parts through, so the large-parts get stuck forever with small-parts cutting in front in line

we might need some system to prevent this?

like, once ANY reservation fails in an update() loop, then all subsequent reservations should fail?

We could fast-follow with a fix for this, since it's pretty difficult at the moment to have different meta-requests using different sized parts (but I know python would like it easily customizable at a per-meta-request level)

source/s3_auto_ranged_get.c
```diff
@@ -266,6 +282,8 @@ static bool s_s3_auto_ranged_get_update(
                     request->part_number,
                     &request->part_range_start,
                     &request->part_range_end);
+                
+                request->ticket = ticket;
```
graebm, line 286: trivial: it's a little spooky that this assignment happens 15 lines after the `ticket` is created. the `ticket` would leak if someone added an early-out branch in the interim.

why not just create the ticket directly into the member variable? like:
```
request->ticket = aws_s3_buffer_pool_reserve(...);
if (request->ticket == NULL)
```

source/s3_auto_ranged_put.c
```diff
@@ -556,29 +556,41 @@ static bool s_s3_auto_ranged_put_update(
...
+                    AWS_LOGF_DEBUG(
+                        AWS_LS_S3_META_REQUEST,
+                        "id=%p: Returning request %p for part %d",
+                        (void *)meta_request,
+                        (void *)request,
+                        request->part_number);
+                } else {
+                    AWS_LOGF_DEBUG(
+                        AWS_LS_S3_META_REQUEST,
+                        "id=%p: Failed to allocate due to lack of memory",
```
graebm, line 591: mention "exceeding memory limit". "lack of memory" is going to make people think malloc failed

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,319 @@
...
+void aws_s3_buffer_pool_release_ticket(
+    struct aws_s3_buffer_pool *buffer_pool,
+    struct aws_s3_buffer_pool_ticket *ticket) {
+
+    if (buffer_pool == NULL || ticket == NULL || ticket->ptr == NULL) {
+        if (ticket != NULL) {
+            aws_mem_release(buffer_pool->base_allocator, ticket);
+        }
+        return;
+    }
```
graebm, line 271: trivial:
```suggestion
    AWS_PRECONDITION(buffer_pool);
    if (ticket == NULL) {
        return;
    }
    if (ticket->ptr == NULL) {
        aws_mem_release(buffer_pool->base_allocator, ticket);
        return;
    }
```

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,319 @@
...
+    size_t block_size,
+    size_t mem_limit) {
+
+    if (mem_limit < GB_TO_BYTES(1)) {
+        AWS_LOGF_ERROR(AWS_LS_S3_CLIENT, "Failed to initialize buffer pool. Min supported value for Memory Limit is 1GB.");
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        return NULL;
+    }
+
+    if (block_size != 0 || (block_size < 1024 * 1024 || block_size % 1024 != 0)) {
```
graebm, line 94: why allow block_size to be 0?

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,319 @@
...
+
+    size_t overall_taken = buffer_pool->primary_used + buffer_pool->primary_reserved + 
+                            buffer_pool->secondary_used + buffer_pool->secondary_reserved;
+
+    /*
+     * 
+     */
+    if (size > buffer_pool->primary_size_cutoff && 
+        (size + overall_taken) > buffer_pool->mem_limit &&
+        (buffer_pool->primary_allocated > (buffer_pool->primary_used + buffer_pool->primary_reserved + buffer_pool->block_size))) {
```
graebm, line 177: trivial: yes, please add comment summarizing the logic, like "if this would exceed the limit, and a trim may help, then trim and see if we get any more memory"

source/s3_buffer_pool.c
```diff
@@ -101,151 +142,178 @@ void aws_s3_buffer_pool_destroy(struct aws_s3_buffer_pool *buffer_pool) {
...
+    aws_mutex_unlock(&buffer_pool->mutex);
+
+    if (ticket == NULL) {
+        aws_raise_error(AWS_ERROR_S3_INSUFFICIENT_MEMORY);
     }
+    return ticket;
 }
 
-static struct aws_s3_pooled_buffer s_primary_acquire(struct aws_s3_buffer_pool *buffer_pool, size_t size) {
+static uint8_t *s_primary_acquire(struct aws_s3_buffer_pool *buffer_pool, size_t size) {
```
graebm, line 201: trivial:
```suggestion
static uint8_t *s_primary_acquire_synced(struct aws_s3_buffer_pool *buffer_pool, size_t size) {
```

source/s3_buffer_pool.c
```diff
@@ -101,151 +142,178 @@ void aws_s3_buffer_pool_destroy(struct aws_s3_buffer_pool *buffer_pool) {
...
-void aws_s3_buffer_pool_release_buffer(
+void aws_s3_buffer_pool_release_ticket(
     struct aws_s3_buffer_pool *buffer_pool,
-    struct aws_s3_pooled_buffer pooled_buffer) {
-    if (buffer_pool == NULL || pooled_buffer.ptr == NULL || pooled_buffer.size == 0) {
+    struct aws_s3_buffer_pool_ticket *ticket) {
+
+    if (buffer_pool == NULL || ticket == NULL || ticket->ptr == NULL) {
+        if (ticket != NULL) {
+            aws_mem_release(buffer_pool->base_allocator, ticket);
```
graebm, line 268: when releasing a ticket that has no ptr, don't we still need to adjust the `buffer_pool->primary/secondary_reserved`?

source/s3_buffer_pool.c
```diff
@@ -101,151 +142,178 @@ void aws_s3_buffer_pool_destroy(struct aws_s3_buffer_pool *buffer_pool) {
...
 
         if (block->alloc_count == 0) {
-            size_t swap_loc = aws_array_list_length(&buffer_pool->blocks) - 1 - swap_count;
-            if (swap_loc == i) {
-                break;
-            } else {
-                aws_array_list_swap(&buffer_pool->blocks, i, swap_loc);
-            }
+            aws_mem_release(buffer_pool->base_allocator, block->block_ptr);
+            aws_array_list_erase(&buffer_pool->blocks, i);
```
graebm, line 152: don't make any changes unless we actually find that this is a bottleneck some day... just sharing a thought:
if this were a linked-list, erase would be O(1) instead of O(N)
BUT then we'd need to allocate each block struct, instead of letting it live by-value in an array-list 🤷‍♀️

source/s3_auto_ranged_get.c
```diff
@@ -194,6 +195,7 @@ static bool s_s3_auto_ranged_get_update(
 
                     request->part_range_start = 0;
                     request->part_range_end = meta_request->part_size - 1; /* range-end is inclusive */
+                    request->part_size = meta_request->part_size;
```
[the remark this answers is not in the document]
DmitriyMusatkin: removed. i mostly used it for parts that do not match meta req part size (last one in upload), but it was not really necessary.

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,228 @@
...
+    for (size_t i = 0; i < aws_array_list_length(&buffer_pool->blocks); ++i) {
+        struct s3_buffer_pool_block *block;
+        aws_array_list_get_at_ptr(&buffer_pool->blocks, (void **)&block, i);
+
+        if (block->alloc_count == 0) {
+            size_t swap_loc = aws_array_list_length(&buffer_pool->blocks) - 1 - swap_count;
+            if (swap_loc == i) {
+                break;
+            } else {
+                aws_array_list_swap(&buffer_pool->blocks, i, swap_loc);
```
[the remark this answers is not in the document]
DmitriyMusatkin: agree. this was premature optimization. removed it and went with a simpler code.

source/s3_buffer_pool.c
DmitriyMusatkin: Error seems to be too excessive, especially since its arguably not an error. Trace?

source/s3_auto_ranged_get.c
DmitriyMusatkin: Discussed offline. Added concept of reservation hold thats placed on buffer whenever reservation fails and cleared on the following round of scheduling.

source/s3_auto_ranged_get.c
DmitriyMusatkin: moved it a bit up. 
allocating into request is a bit of a chicken and egg problem. if reservation fails we dont want to create a request at all

source/s3_buffer_pool.c
DmitriyMusatkin: Agree. i was somewhat concerned about O(N) and went with a swapping approach to reduce complexity of overall trim. That was probably premature optimization and from some initial testing it does not seem to affect perf too much.

source/s3_auto_ranged_put.c
```diff
@@ -942,7 +954,12 @@ struct aws_future_http_message *s_s3_prepare_upload_part(struct aws_s3_request *
         /* Read the body */
         uint64_t offset = 0;
         size_t request_body_size = s_compute_request_body_size(meta_request, request->part_number, &offset);
-        aws_byte_buf_init(&request->request_body, meta_request->allocator, request_body_size);
+        if (request->request_body.capacity == 0) {
+            AWS_FATAL_ASSERT(request->ticket);
+            request->request_body =
+                aws_s3_buffer_pool_acquire_buffer(request->meta_request->client->buffer_pool, request->ticket);
+            request->request_body.capacity = request_body_size;
```
graebm, line 961: is there any chance that `request_body_size` is larger than `meta_request->part_size`? because `meta_request->part_size` is what we used when reserving the ticket

source/s3_client.c
```diff
@@ -243,6 +254,17 @@ struct aws_s3_client *aws_s3_client_new(
         return NULL;
     }
 
+    if (client_config->max_part_size != 0 && client_config->memory_limit_in_bytes != 0 &&
+        client_config->max_part_size >
+            (client_config->memory_limit_in_bytes / s_default_max_part_size_to_mem_lim_multiplier)) {
```
graebm, line 259: trivial: use existing helper function?
```suggestion
            s_default_max_part_size_based_on_mem_limit(client_config->memory_limit_in_bytes)) {
```

source/s3_client.c
```diff
@@ -293,17 +346,18 @@ struct aws_s3_client *aws_s3_client_new(
...
+    *((size_t *)&client->part_size) = part_size;
 
     if (client_config->max_part_size != 0) {
         *((uint64_t *)&client->max_part_size) = client_config->max_part_size;
     } else {
         *((uint64_t *)&client->max_part_size) = s_default_max_part_size;
     }
+
+    if (client_config->max_part_size > s_default_max_part_size_based_on_mem_limit(mem_limit)) {
+        *((uint64_t *)&client->max_part_size) = s_default_max_part_size_based_on_mem_limit(mem_limit);
```
graebm, line 358: this could be a breaking change.

It's unlikely that anyone was setting config->max_part_size, so they were allowed to upload 4GiB (32bit system) or 5GiB (64bit system) parts by default.

But now, the largest they can upload is 1/4 the default memory-limit, so 256MiB or 512MiB on 32bit, or 2GiB 1GiB or 512MiB on 64bit

source/s3_client.c
```diff
@@ -1130,6 +1230,54 @@ static void s_s3_client_schedule_process_work_synced_default(struct aws_s3_clien
...
+    (void)task;
+    (void)task_status;
+
+    /* Client keeps a reference to the event loop group; a 'canceled' status should not happen.*/
+    AWS_ASSERT(task_status == AWS_TASK_STATUS_RUN_READY);
+
+    struct aws_s3_client *client = arg;
+    AWS_PRECONDITION(client);
+
+    uint32_t num_reqs_in_flight = (uint32_t)aws_atomic_load_int(&client->stats.num_requests_in_flight);
```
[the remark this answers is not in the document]
graebm: still iffy about using `stats`. why not check `aws_linked_list_empty(&client->threaded_data.meta_requests)`?

source/s3_client.c
```diff
@@ -1130,6 +1189,55 @@ static void s_s3_client_schedule_process_work_synced_default(struct aws_s3_clien
     client->synced_data.process_work_task_scheduled = true;
 }
 
+/* Task function for trying to find a request that can be processed. */
+static void s_s3_client_trim_buffer_pool_task(struct aws_task *task, void *arg, enum aws_task_status task_status) {
+    AWS_PRECONDITION(task);
+    (void)task;
+    (void)task_status;
```
graebm, line 1196: debatable: early-out if task_status != RUN?

so we don't have some weird thing happen when this task is canceled during client shutdown?

source/s3_client.c
```diff
@@ -1130,6 +1189,55 @@ static void s_s3_client_schedule_process_work_synced_default(struct aws_s3_clien
...
+
+    struct aws_s3_client *client = arg;
+    AWS_PRECONDITION(client);
+
+    aws_s3_client_lock_synced_data(client);
+    client->synced_data.trim_buffer_pool_task_scheduled = false;
+    aws_s3_client_unlock_synced_data(client);
```
graebm, line 1203: we could move `trim_buffer_pool_task_scheduled` from `synced_data` to `threaded_data` and not bother with locks here, right? we only ever schedule this and run it from the client-work-thread?

source/s3_client.c
DmitriyMusatkin: its max part size. this will switch max part size from 4GB to 512MB and try to enforce 2GB in flight. I doubt if anything close to 4GB part size ever worked in the past unless we were able to enforce that only a couple parts were running at the same time, given that we panic on failed allocations.

source/s3_client.c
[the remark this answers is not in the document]
DmitriyMusatkin: threaded.meta_requests can have a subset of requests, since there could be some in the synced.pending queue. stats one seemed easier to use and we already rely on it in scheduling and logic for it is fairly simple.

source/s3_buffer_pool.c
DmitriyMusatkin: disables primary indirectly. i mostly used it for testing. but can take it out

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,379 @@
...
+            ++chunks_used; /* round up */
+        }
+
+        bool found = false;
+        for (size_t i = 0; i < aws_array_list_length(&buffer_pool->blocks); ++i) {
+            struct s3_buffer_pool_block *block;
+            aws_array_list_get_at_ptr(&buffer_pool->blocks, (void **)&block, i);
+
+            if (block->block_ptr <= ticket->ptr && block->block_ptr + block->block_size > ticket->ptr) {
+                size_t alloc_i = (ticket->ptr - block->block_ptr) / buffer_pool->chunk_size;
```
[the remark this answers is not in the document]
DmitriyMusatkin: put chunks_used to ticket. with alloc_i it gets tricky with trim, there are possible workarounds for that, but just recalculating it for now

source/s3_auto_ranged_put.c
DmitriyMusatkin: no. request_body_size is always smaller than meta_request->part_size. 
request_body_size is either meta_request->part_size or in case of the last part it will be remainder of data. 

ideally we should be able to use bigger buffers than needed in upload, but we currently rely on capacity for some checks in code. i fixed couple places to be more generic, but at this point i dont have complete confidence that it will work correctly with full buffer everywhere. hence why its doing the trick of setting smaller capacity.

will put a comment about it

source/s3_buffer_pool.c
```diff
@@ -0,0 +1,410 @@
...
+    AWS_PRECONDITION(buffer_pool);
+
+    if (buffer_pool->has_reservation_hold) {
+        return NULL;
+    }
+
+    if (size == 0) {
+        AWS_LOGF_ERROR(AWS_LS_S3_CLIENT, "Could not reserve from buffer pool. 0 is not a valid size.");
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        return NULL;
```
graebm, line 213: Real code never checks the error-code from `aws_s3_buffer_pool_reserve()`. Real code assumes it failed due to the memory limit being temporarily exceeded.
- so if size was 0, the meta-request would get stuck perpetually trying to acquire a ticket
- similarly, if `size > mem_limit` the meta-request would be stuck perpetually trying to acquire a ticket. I realize you adjusted the client's `max_part_size` to make this impossible, but it might be worth an ASSERT or FATAL_ASSERT here in case client code changes in the future and something slips through


Anyway, we should probably do one of these things:
1) real code calling aws_s3_buffer_pool_reserve() continues to assume that it can only fail when the mem limit is temporarily exceeded. And the reserve() function here FATAL_ASSERTs that the size is legal (not 0, not exceeding mem_limit)
2) OR real code calling aws_s3_buffer_pool_reserve() checks to differentiate the expected "waiting for memory" from other errors that terminate the meta-request

source/s3_client.c
```diff
@@ -80,6 +81,12 @@ static size_t s_dns_host_address_ttl_seconds = 5 * 60;
  * 30 seconds mirrors the value currently used by the Java SDK. */
 static const uint32_t s_default_throughput_failure_interval_seconds = 30;
 
+/* Default multiplier between max part size and memory limit */
+static const size_t s_default_max_part_size_to_mem_lim_multiplier = 4;
```
graebm, line 85: style: there's a relationship between this magic number, and the `buffer_pool->primary_size_cutoff = chunk_size * 4;` magic number from s3_buffer_pool.c

Add comment about this relationship

Or is it worth having 1 magic number in s3_buffer_pool.c, and common function both can call, like `aws_s3_buffer_pool_get_primary_size_cutoff(size_t chunk_size)`? Or first we create the buffer_pool, then s3_client.c calls a getter to clamp the client->max_part_size if necessary, like:
```
client->max_part_size = aws_size_min(client->max_part_size, aws_s3_buffer_pool_get_primary_size_cutoff(client->buffer_pool));
```

source/s3_client.c
```diff
@@ -257,12 +279,43 @@ struct aws_s3_client *aws_s3_client_new(
...
+    }
+
+    size_t part_size;
+    if (client_config->part_size != 0) {
+        part_size = (size_t)client_config->part_size;
+    } else {
+        part_size = s_default_part_size;
+    }
+
+    client->buffer_pool = aws_s3_buffer_pool_new(allocator, part_size, mem_limit);
```
graebm, line 311: need to check that this succeeded. It will raise an error if mem_limit isn't just right

graebm (approved):
fix & ship

source/s3_buffer_pool.c
DmitriyMusatkin: yeah, good point. size == 0 or > mem_lim is really an indication that something wrong went on before, so for simplicity sake, i think fatal assert makes more sense.

PR #377, opened 2023-11-22
Report S3 operation name of specific request that failed.

graebm:
**Issue:**
It was difficult to map errors from aws-c-s3 into errors in other AWS SDKs without knowing exactly which HTTP request failed.

For example, if `AWS_S3_META_REQUEST_TYPE_PUT_OBJECT` fails, they need to know whether it was a "PutObject", "CreateMultipartUpload", "UploadPart", or "CompleteMultipartUpload" HTTP request that failed.

**Description of changes:**
- Report the S3 operation name of the specific HTTP request that failed (e.g. "CreateMultipartUpload").
- Likewise, report that operation name in metrics.
- For `AWS_S3_META_REQUEST_TYPE_DEFAULT`, where aws-c-s3 isn't sure what operation is being performed, allow the user to pass in the operation name.
- Internally, pass request_type and operation_name around directly. Instead of querying them via vtable gymnastics, which I found confusing and error-prone.


source/s3_copy_object.c
```diff
@@ -37,14 +37,13 @@ static void s_s3_copy_object_request_finished(
...
 
-static int s_s3_copy_object_request_type(const struct aws_s3_request *request) {
+static enum aws_s3_request_type s_s3_copy_object_request_type(const struct aws_s3_request *request) {
     switch (request->request_tag) {
         case AWS_S3_COPY_OBJECT_REQUEST_TAG_GET_OBJECT_SIZE:
             /* It's a HEAD request of GetObject call */
             return AWS_S3_REQUEST_TYPE_HEAD_OBJECT;
         case AWS_S3_COPY_OBJECT_REQUEST_TAG_BYPASS:
-            /* A single copy object request, same as default */
-            return AWS_S3_REQUEST_TYPE_DEFAULT;
```
graebm, line 47: this is correct, right?

source/s3_copy_object.c
```diff
@@ -37,14 +37,13 @@ static void s_s3_copy_object_request_finished(
...
+            return AWS_S3_REQUEST_TYPE_COPY_OBJECT;
```
graebm, line 46: Is this change correct? BYPASS == "CopyObject"?

include/aws/s3/s3_client.h
```diff
@@ -974,6 +1009,17 @@ int aws_s3_request_metrics_get_thread_id(const struct aws_s3_request_metrics *me
...
+ * Be warned this string's lifetime is tied to the metrics object.
+ */
+AWS_S3_API
+int aws_s3_request_metrics_get_operation_name(
+    const struct aws_s3_request_metrics *metrics,
+    const struct aws_string **out_operation_name);
+
 /* Get the request type from request metrics. */
 AWS_S3_API
 void aws_s3_request_metrics_get_request_type(
```
jamesbornholt, line 1025: IIRC we added this function and all the `aws_s3_request_type` stuff specifically for Mountpoint for use in metrics. If Mountpoint is still the only customer for it and you feel like you want to break it, it would be really easy for us to switch to this new interface instead, since we really only needed it to get operation names anyway.

graebm: I'll keep the enums. I'm finding uses for both enum and string

source/s3_auto_ranged_put.c
```diff
@@ -480,7 +462,9 @@ static bool s_s3_auto_ranged_put_update(
                 request = aws_s3_request_new(
                     meta_request,
                     AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_LIST_PARTS,
```
TingDaoK, line 464: Not exactly for this PR, but do we still need the separated tags for each meta request type? Feel like it's a repeated info here.

source/s3_request.c
```diff
@@ -29,6 +33,8 @@ struct aws_s3_request *aws_s3_request_new(
     request->meta_request = aws_s3_meta_request_acquire(meta_request);
 
     request->request_tag = request_tag;
+    request->request_type = request_type;
+    request->operation_name = operation_name;
```
TingDaoK, line 37: debatable: I don't really like the idea of request doesn't own the string, it feels easy to mess up the lifetime.

source/s3_default_meta_request.c
```diff
@@ -143,7 +158,14 @@ static bool s_s3_meta_request_default_update(
                     goto has_work_remaining;
                 }
 
-                request = aws_s3_request_new(meta_request, 0, 1, AWS_S3_REQUEST_FLAG_RECORD_RESPONSE_HEADERS);
+                request = aws_s3_request_new(
+                    meta_request,
+                    0 /*request_tag*/,
+                    AWS_S3_REQUEST_TYPE_DEFAULT,
+                    meta_request_default->operation_name ? aws_string_c_str(meta_request_default->operation_name)
+                                                         : NULL,
```
TingDaoK, line 166: debatable: will it be better to have a default string instead of null if nothing was set? 
The reason I think is:
- Without setting a operation name in this case is not an ERROR, using a NULL more like an error to me.
- The TYPE enum we have a default value for the type, but the operation name does not.

source/s3_auto_ranged_put.c
```diff
@@ -510,7 +496,9 @@ static bool s_s3_auto_ranged_put_update(
                 request = aws_s3_request_new(
                     meta_request,
                     AWS_S3_AUTO_RANGED_PUT_REQUEST_TAG_CREATE_MULTIPART_UPLOAD,
-                    0,
+                    AWS_S3_REQUEST_TYPE_CREATE_MULTIPART_UPLOAD,
+                    "CreateMultipartUpload",
```
TingDaoK, line 500: debatable: I would prefer to have a global static string for the operation names we set (I know you are not a fan of those).
- In case the name changed from the model, it's easier to us to update
- multiple places sharing the same string.

source/s3_copy_object.c
TingDaoK: yep

source/s3_auto_ranged_put.c
graebm: There's still 1 case where the meta-request has 2 tags that use the same-request type:
https://github.com/awslabs/aws-c-s3/blob/f961971f3851782d40126f0597f007c73ba4ae0c/source/s3_auto_ranged_get.c#L49-L51

🤷‍♀️

source/s3_auto_ranged_put.c
graebm: I've got another idea that will put all the strings in the same place... next push

include/aws/s3/s3_client.h
```diff
@@ -481,6 +483,15 @@ struct aws_s3_meta_request_options {
...
     enum aws_s3_meta_request_type type;
 
+    /**
+     * Optional.
+     * The S3 operation name (e.g. "CreateBucket").
+     * You should only set this when type is AWS_S3_META_REQUEST_TYPE_DEFAULT;
+     * it is automatically populated for other meta-request types.
+     * This name is used to fill out details in metrics and error reports.
+     */
+    struct aws_byte_cursor operation_name;
```
waahm7, line 493: I think we're overcomplicating things by using both an enum and a string for the same data. It seems unnecessary to have a request-level variable set only in one specific case just to return it later. If a user sets 'default', returning NULL or 'Default' should be sufficient. User can include any additional information in the user_data of the request.

waahm7: In most cases, it is just repeated information, like type = getObject, operationName = getObject.

DmitriyMusatkin (approved):

source/s3_request.c
TingDaoK: I just thought out something like: We can have the operation name only for default type, and use the request type for all other operations. So, that we can safe some allocation, lol

graebm: whatever, I just made it an aws_string
simple
obvious

source/s3_auto_ranged_put.c
graebm: I eliminated pretty much all the repeatedly typed strings
Still not using global static variables 😛

source/s3_default_meta_request.c
graebm: yeah, it's super debatable whether to have "" or NULL, and whether that differs based on if you're storing `const char *` or `aws_string *`.

Anyway, these are `aws_string *` now, and I'm still using NULL for defaults. because allocating an aws_string, just to store "" is too weird for me

source/s3_auto_ranged_put.c
graebm: We talked, and we'd all like to remove the concept of "request tag".
But that's a sizable change, so we're going to tackle it in the future.
I added a TODO

include/aws/s3/s3_client.h
graebm: I eliminated a lot of the repeated typing.
But I still think it's worth having both enum and string, so we can let users pass through the names of the N different S3 operations we don't have explicit enums for.

include/aws/s3/s3_client.h
```diff
@@ -481,6 +497,15 @@ struct aws_s3_meta_request_options {
     /* The type of meta request we will be trying to accelerate. */
     enum aws_s3_meta_request_type type;
 
+    /**
+     * Optional.
+     * The S3 operation name (e.g. "CreateBucket").
+     * You should only set this when type is AWS_S3_META_REQUEST_TYPE_DEFAULT;
```
waahm7, line 503: `Only set this when` implies that it's an error to set this in other cases. However, for some workflows, it might be simpler to always set this, and we may or may not use it.
```suggestion
     * This will only be used when type is AWS_S3_META_REQUEST_TYPE_DEFAULT;
```

waahm7 (approved):

TingDaoK (approved):

PR #389, opened 2023-12-04
Skip headRequest for RangeGet Part 1

waahm7:
*Description of changes:*
This approach skips the head request when checksum verification is enabled and size_hint indicates that it is a one-part file by replacing it with a getPart request. We cancel the request if the content length of the first part is greater than part_size. This method helps us avoid extra requests for objects with content_length <= part_size.


source/s3_auto_ranged_get.c
```diff
@@ -600,14 +624,15 @@ static void s_s3_auto_ranged_get_request_finished(
 
     bool found_object_size = false;
     bool request_failed = error_code != AWS_ERROR_SUCCESS;
+    int request_error_code = error_code;
```
graebm, line 627: seems error-prone to have 2 different error_code variables being used in this function. Someone's going to use the wrong one someday.

maybe, instead of 2 different error_codes, introduce
`bool first_part_size_mismatch = (error_code == AWS_ERROR_S3_GET_PART_SIZE_MISMATCH);`
and just keep this function using the 1 `error_code`

OR rename the function param ~int error_code)~ -> `int request_error_code)`. And then later in the function introduce the new `int error_code = AWS_ERROR_SUCCESS;`, so the 2 different variables aren't coexisting for so long, and aren't switching roles halfway through the function

source/s3_auto_ranged_get.c
```diff
@@ -691,6 +715,15 @@ static void s_s3_auto_ranged_get_request_finished(
                 auto_ranged_get->synced_data.head_object_completed = true;
                 AWS_LOGF_DEBUG(AWS_LS_S3_META_REQUEST, "id=%p Head object completed.", (void *)meta_request);
                 break;
+            case AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_PART_NUMBER:
+                auto_ranged_get->synced_data.get_first_part_completed = true;
+                AWS_LOGF_DEBUG(AWS_LS_S3_META_REQUEST, "id=%p Get Part Number completed.", (void *)meta_request);
+                if (error_code == AWS_ERROR_S3_GET_PART_SIZE_MISMATCH && found_object_size) {
+                    /* Try to fetch the first part again as a ranged get */
```
graebm, line 722: what happens if AWS_ERROR_S3_GET_PART_SIZE_MISMATCH but we NOT found_object_size?

source/s3_auto_ranged_get.c
```diff
@@ -691,6 +715,15 @@ static void s_s3_auto_ranged_get_request_finished(
...
+                    break;
+                }
+                ++auto_ranged_get->synced_data.num_parts_requested;
```
graebm, line 725: I made another comment elsewhere about "why don't we increment num_parts_requested when we create the GET_FIRST_PART?"

Now I see. But I think this would be less confusing if we DID increment num_parts_requested there in the normal way, and then over here if we realize that we need to re-request the part we do the weird thing like:
```
/* Try to fetch the first part again as a ranged get */
auto_ranged_get->synced_data.num_parts_requested = 0;
break;

File: source/s3_meta_request.c [hunk 4]
```diff
@@ -1504,18 +1542,29 @@ void aws_s3_meta_request_send_request_finish_default(
...
-                (void *)meta_request,
-                error_code,
-                aws_error_str(error_code),
-                (void *)request,
-                response_status);
+            if (error_code == AWS_ERROR_S3_GET_PART_SIZE_MISMATCH) {
+                /* Log at info level instead of error as it's expected and not a fatal error */
+                AWS_LOGF_INFO(
+                    AWS_LS_S3_META_REQUEST,
+                    "id=%p Meta request cannot recover from error %d (%s). (request=%p, response status=%d)",
```
graebm, line 1552: do the error msg need to be updated? I'm confused how the comment says "not a fatal error" but the msg says "cannot recover from error"

source/s3_auto_ranged_get.c
waahm7: The error should be replaced by why we were not able to find the object size. The only possible case this can happen for now is when Content-Length was present but there was no Content-Range header. So we will report Content-Range not found error to the user.

include/aws/s3/private/s3_auto_ranged_get.h
```diff
@@ -12,6 +12,7 @@ enum aws_s3_auto_ranged_get_request_type {
     AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_HEAD_OBJECT,
     AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_PART,
     AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_INITIAL_MESSAGE,
+    AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_PART_NUMBER,
```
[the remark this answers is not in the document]
waahm7: Thanks, I have updated TYPE_PART to TYPE_GET_RANGED.

source/s3_meta_request.c
```diff
@@ -1260,6 +1265,39 @@ static int s_s3_meta_request_incoming_headers(
...
+        request->request_tag == AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_PART_NUMBER) {
+        uint64_t content_length;
+        if (!aws_s3_parse_content_length_response_header(
+                request->allocator, request->send_data.response_headers, &content_length)) {
+            if (content_length > meta_request->part_size ||
+                (content_length < meta_request->part_size &&
+                 aws_http_headers_has(
+                     request->send_data.response_headers,
+                     g_mp_parts_count_header_name) /* Object has multiple parts if this header is present */)) {
+                return aws_raise_error(AWS_ERROR_S3_GET_PART_SIZE_MISMATCH);
```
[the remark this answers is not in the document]
waahm7: Thanks, after some benchmarking and offline discussion, we decided to add a size_hint and we only cancel the request if the user provides a bad hint.

source/s3_auto_ranged_get.c
waahm7: Thanks, fixed.

source/s3_auto_ranged_get.c
```diff
@@ -677,21 +787,34 @@ static void s_s3_auto_ranged_get_request_finished(
...
         switch (request->request_tag) {
             case AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_HEAD_OBJECT:
                 auto_ranged_get->synced_data.head_object_completed = true;
                 AWS_LOGF_DEBUG(AWS_LS_S3_META_REQUEST, "id=%p Head object completed.", (void *)meta_request);
                 break;
-            case AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_PART:
+            case AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_PART_NUMBER:
+                AWS_LOGF_DEBUG(AWS_LS_S3_META_REQUEST, "id=%p Get Part Number completed.", (void *)meta_request);
+                if (first_part_size_mismatch && found_object_size) {
+                    /* Try to fetch the first part again as a ranged get */
```
graebm, line 812: trivial:
```suggestion
                    /* We canceled GET_PART_NUMBER_1 because it was bigger than expected.
                     * Try to fetch the first part again as a ranged get */
```

include/aws/s3/s3_client.h
```diff
@@ -689,6 +689,15 @@ struct aws_s3_meta_request_options {
      * from the buffer and compare them them to previously uploaded part checksums.
      */
     struct aws_s3_meta_request_resume_token *resume_token;
+
+    /*
+     * Optional.
+     * Object size hint in bytes.
+     * This hint helps optimize downloads for small (less than or equal to part_size) or large files.
```
graebm, line 696: trivial: me trying to help people understand what "optimize" means.
```suggestion
     * The optimal strategy for downloading a file depends on its size.
     * Set this hint to help the S3 client choose the best strategy for this particular file.
```

graebm (commented):
Looks great! Just suggesting comments and naming

graebm (approved):
Fix & ship

include/aws/s3/private/s3_auto_ranged_get.h
```diff
@@ -10,8 +10,9 @@
 
 enum aws_s3_auto_ranged_get_request_type {
     AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_HEAD_OBJECT,
-    AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_PART,
+    AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_RANGE,
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

PR #393, opened 2023-12-09
support meta request level override of part size and mpu threshold 

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/s3_client.c
```diff
@@ -1233,17 +1252,29 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
...
+                            content_length, part_size_config, client_max_part_size, &adjusted_part_size, &num_parts)) {
                         return NULL;
                     }
                 }
+                if (adjusted_part_size != part_size_config) {
+                    AWS_LOGF_DEBUG(
+                        AWS_LS_S3_META_REQUEST,
+                        "The multipart upload part size has been adjusted to %" PRIu64 "",
+                        (uint64_t)adjusted_part_size);
+                }
```
graebm, line 1268: trivial: I would move this logic up before we set multipart_upload_threshold. Then, if part-size gets adjusted, the multipart_upload_threshold will also be adjusted.

Also, not sure it's worth the potential confusion of having 2 `part_size_config` and `adjusted_part_size` variables? Maybe just have 1 `part_size` variable that we update as the function progresses?

graebm (approved):

PR #395, opened 2023-12-15
Cancel/Pause stream optimization 

TingDaoK:
*Issue #, if available:*

- We previously cancel/pause meta request by not making any new request, but wait for anything inflight to complete, which doesn't need to happen.
- We want to cancel streams quickly without waiting for http level

*Description of changes:*

- Keep a list of requests that have ongoing http level activities.
- When cancel or pause invoked, go through the list and cancel the underlying http streams, which will actually cancel the request without waiting for HTTP req/resp to complete.
- But, for MPU cancel, we still need to wait for abort MPU to complete.


include/aws/s3/private/s3_request.h
```diff
@@ -113,6 +113,15 @@ struct aws_s3_request {
     /* Linked list node used for queuing. */
     struct aws_linked_list_node node;
 
+    /* Linked list node used for tracking the request is active from HTTP level. */
+    struct aws_linked_list_node on_going_http_node;
```
waahm7, line 117: _on_going_http_node is confusing. Maybe make it more descriptive. 
Something like `ongoing_http_request_list_node`

source/s3_auto_ranged_put.c
```diff
@@ -1711,6 +1711,17 @@ static int s_s3_auto_ranged_put_pause(
      */
     aws_s3_meta_request_set_fail_synced(meta_request, NULL, AWS_ERROR_S3_PAUSED);
 
+    while (!aws_linked_list_empty(&meta_request->synced_data.ongoing_http_request_list)) {
```
waahm7, line 1714: Can we now look into supporting cancel for an unknown content_length?

DmitriyMusatkin, line 1714: nit: this can probably be pulled into a helper in meta req, so we dont have to repeat code twice

DmitriyMusatkin (approved):

PR #396, opened 2023-12-18
Refactor GetObject Empty file Logic - Part 2

waahm7:
*Description of changes:*
- Remove `AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_INITIAL_MESSAGE` request type and use `AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_PART_NUMBER_1` if `AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_RANGE` fails due to an empty object. We will try downloading the first part again with `partNumber`; it will either download an empty file or download the updated file.
- Fix `aws_s3_calculate_auto_ranged_get_num_parts` for an empty file.
- Allow users to pass 0 as a `size_hint`.
- Fix onHeaders callback for an empty file.


source/s3_auto_ranged_get.c
```diff
@@ -91,8 +91,10 @@ struct aws_s3_meta_request *aws_s3_meta_request_auto_ranged_get_new(
     auto_ranged_get->initial_message_has_range_header = aws_http_headers_has(headers, g_range_header_name);
     auto_ranged_get->initial_message_has_if_match_header = aws_http_headers_has(headers, g_if_match_header_name);
     auto_ranged_get->synced_data.first_part_size = auto_ranged_get->base.part_size;
-    auto_ranged_get->object_size_hint = options->object_size_hint;
-
+    if (options->object_size_hint) {
```
graebm, line 94: utterly trivial:
```suggestion
    if (options->object_size_hint != NULL) {
```

source/s3_auto_ranged_get.c
```diff
@@ -117,6 +119,16 @@ static enum aws_s3_auto_ranged_get_request_type s_s3_get_request_type_for_discov
     struct aws_s3_auto_ranged_get *auto_ranged_get = meta_request->impl;
     AWS_ASSERT(auto_ranged_get);
 
+    /*
+     * If the object is empty, download the file using a GetObject with PartNumber request. This will succeed if
+     * the file is empty, or it will download the file.
+     */
```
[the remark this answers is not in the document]
graebm: this explains what we're doing, but there are subtle reasons we're doing it this way that should be explained somewhere 

explain why we're bothering to send another request, when we already sent one and there's no body anyway

and explain why we're downloading with part-number, instead of something simpler like a basic GetObject

If this isn't the right place for that explanation, tell the reader here where they can find that explanation

source/s3_auto_ranged_get.c
```diff
@@ -117,6 +119,16 @@ static enum aws_s3_auto_ranged_get_request_type s_s3_get_request_type_for_discov
...
+    if (auto_ranged_get->synced_data.object_range_empty != 0) {
```
graebm, line 126: I'm really confused how we end up back in `s_s3_get_request_type_for_discovering_object_size()` if we've already discovered that `object_range_empty`

When `object_range_empty` becomes true, `object_range_empty` becomes true too
And after `object_range_empty` become true, we don't call `s_s3_get_request_type_for_discovering_object_size()` .. right?

source/s3_auto_ranged_get.c
```diff
@@ -535,7 +518,8 @@ static int s_discover_object_range_and_content_length(
     uint64_t *out_total_content_length,
     uint64_t *out_object_range_start,
     uint64_t *out_object_range_end,
-    uint64_t *out_first_part_size) {
+    uint64_t *out_first_part_size,
+    bool *empty_file_error) {
```
DmitriyMusatkin, line 522: nit: should this have out_ prefix?

source/s3_auto_ranged_get.c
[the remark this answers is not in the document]
waahm7: `HeadObject` doesn't fail, and we do get the Etag, and even the checksum if it's available. After `HeadObject`, if the file is empty, we just complete the request. I have added a test for this case [here] (https://github.com/awslabs/aws-c-s3/pull/389/files#diff-0f16837166c257c5841490cb42e6840835b93fbbcdc798e896ec0728bbe07bd5R3604).

source/s3_auto_ranged_get.c
```diff
@@ -821,6 +806,17 @@ static void s_s3_auto_ranged_get_request_finished(
                 }
                 /* fall through */
             case AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_RANGE:
+                if (empty_file_error) {
```
DmitriyMusatkin, line 809: we do have logic somewhere to prevent file getting different parts of file if it being actively updated on s3 side. what currently happens if we make a req, and get file is empty error, but then when we get the first part, its been updated on s3 side and is not empty anymore?

waahm7: If we make a request and receive an empty file error, we don't store the ETag at https://github.com/awslabs/aws-c-s3/blob/main/source/s3_auto_ranged_get.c#L630C13-L630C34 since the request failed. So, we try to download the file again with a `partNumber`.
- If the file is still empty, we provide good headers to the user, and the overall request is successful.
- If the file has been updated on the S3 side, we proceed with the normal flow of downloading the first part, storing its ETag, and then parallel downloading the rest of the parts while ensuring that the file has not changed again.

source/s3_auto_ranged_get.c
waahm7: I have improved the documentation to add more details.

source/s3_auto_ranged_get.c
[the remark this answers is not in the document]
waahm7: I have added more details to the documentation about the flow. 

`HeadObject` doesn't fail, and we do get the ETag, and even the checksum, if available. After `HeadObject`, if the file is empty, we simply complete the request. I have added a test for this case and fixed a bug. Now we have the following `empty_file` cases; let me know if I have missed any:
- We execute `AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_RANGE`; it fails, and we download the file with `AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_PART_NUMBER_1`.
- We execute `AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_PART_NUMBER_1`, and it succeeds with a checksum.
- We perform `HeadObject`, and it succeeds with a checksum.
- We execute `AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_RANGE`, and the user has provided a range header; it fails with a 416 RANGE_NOT_SATISFIABLE error.

source/s3_auto_ranged_get.c
waahm7: Thanks, fixed.

DmitriyMusatkin (approved):

graebm (approved):

Issue #370, opened 2023-11-14
Handle range header client-side


eeroel:
### Describe the feature

Would it be possible to avoid the HeadObject requests when doing a GET range request? I noticed this comment but I wonder if it's something that's feasible, or in the plans? https://github.com/awslabs/aws-c-s3/blob/83008e577804643bc632ae4e603f36ab96219b9b/source/s3_auto_ranged_get.c#L166

### Use Case

When reading data in Parquet format (e.g. data lake applications), the file footer needs to be read first, so an implementation that reads from S3 needs to start with a HeadObject request and thus already knows the object size. The data itself may then be read in several small range requests, so making redundant HeadObject requests for each of those adds up latency. I understand that this library is optimized for throughput, but it would be great if there was a way to have those performance benefits without introducing latency in cases where the amount of data read is small.

### Proposed Solution

I'm not familiar with the internals of the auto-range request implementation, but maybe the first request could be made to the last range (at the end of the object) so that an Unsatisfiable error will be returned if the range is out of bounds?

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

jmklix:
This is something that we would like to add support for, but this is not currently a high priority.

waahm7:
@eeroel Thank you for creating the issue. I have implemented client-side range-header handling, provided the range header includes a start-range. If the range header includes a start range, we no longer perform a HeadRequest. Does this solve your issue?

eeroel:
> @eeroel Thank you for creating the issue. I have implemented client-side range-header handling, provided the range header includes a start-range. If the range header includes a start range, we no longer perform a HeadRequest. Does this solve your issue?

Nice, yes it does!

waahm7:
@eeroel Thanks, this is resolved in https://github.com/awslabs/aws-c-s3/releases/tag/v0.4.8.

PR #397, opened 2023-12-20
Refactor GetObject request with Range Header to Skip HeadRequest - Part 3

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-c-s3/issues/370

*Description of changes:*
- This PR refactors our `GetObject` request with `Range` header logic to be parsed on the client side, as long as the Range includes a start-range.


include/aws/s3/private/s3_auto_ranged_get.h
```diff
@@ -53,6 +53,12 @@ struct aws_s3_auto_ranged_get {
     } synced_data;
 
     uint32_t initial_message_has_range_header : 1;
+
+    bool initial_message_has_start_range;
```
waahm7, line 57: Using bool because we can not have a pointer to bit field https://stackoverflow.com/questions/13547352/c-cannot-take-address-of-bit-field.

source/s3_util.c
```diff
@@ -488,6 +488,74 @@ int aws_s3_parse_content_length_response_header(
...
+            *out_initial_message_has_end_range = true;
+        }
+
+        /* verify there is not anything extra */
+        if (aws_byte_cursor_next_split(&range_header_value, '-', &substr)) {
+            goto done;
+        }
+
+        /* verify that start-range <= end-range */
+        if (out_initial_message_has_end_range && out_initial_start_range > out_initial_end_range) {
```
waahm7, line 544: When we perform a `HeadRequest` with `startRange > endRange` such as `bytes=20-10`, S3 does not add a `Content-Range` header in the `HeadRequest` response, and `Content-Length` reflects the total length of the object. The CLI downloads the entire object. The current CRT behavior is that it fails with `AWS_ERROR_S3_MISSING_CONTENT_RANGE_HEADER` at https://github.com/awslabs/aws-c-s3/blob/main/source/s3_auto_ranged_get.c#L510C24-L510C66. According to the HTTP [spec](https://www.rfc-editor.org/rfc/rfc9110.html#section-14.1.1-6), “An int-range is invalid if the last-pos value is present and less than the first-pos.” . This changes it to `AWS_ERROR_S3_INVALID_RANGE_HEADER` error at request time.

source/s3_util.c
```diff
@@ -488,6 +488,74 @@ int aws_s3_parse_content_length_response_header(
...
+            *out_initial_message_has_end_range = true;
+        }
+
+        /* Verify that there is nothing extra */
+        if (aws_byte_cursor_next_split(&range_header_value, '-', &substr)) {
+            goto done;
+        }
+
+        /* verify that start-range <= end-range */
+        if (*out_initial_message_has_end_range && *out_initial_start_range > *out_initial_end_range) {
```
waahm7, line 545: When we perform a `HeadRequest` with `startRange > endRange` such as `bytes=20-10`, S3 does not add a `Content-Range` header in the `HeadRequest` response, and `Content-Length` reflects the total length of the object. The CLI downloads the entire object. The current CRT behavior is that it fails with `AWS_ERROR_S3_MISSING_CONTENT_RANGE_HEADER` at https://github.com/awslabs/aws-c-s3/blob/main/source/s3_auto_ranged_get.c#L510C24-L510C66. According to the HTTP [spec](https://www.rfc-editor.org/rfc/rfc9110.html#section-14.1.1-6), “An int-range is invalid if the last-pos value is present and less than the first-pos.” . This changes it to `AWS_ERROR_S3_INVALID_RANGE_HEADER` error at request time.

source/s3_auto_ranged_get.c
```diff
@@ -651,11 +699,16 @@ static int s_discover_object_range_and_content_length(
...
-
-            /* When discovering the object size via first-part, the object range is the entire object. */
-            object_range_start = 0;
-            object_range_end = total_content_length - 1; /* range-end is inclusive */
-
+            if (auto_ranged_get->initial_message_has_range_header) {
+                object_range_end = object_size - 1;
+                if (auto_ranged_get->initial_message_has_end_range) {
+                    object_range_end = aws_min_u64(object_range_end, auto_ranged_get->initial_object_range_end);
+                }
```
graebm, line 708: utterly trivial: reads better as if/else
```suggestion
                if (auto_ranged_get->initial_message_has_end_range) {
                    object_range_end = aws_min_u64(object_range_end, auto_ranged_get->initial_object_range_end);
                } else {
                    object_range_end = object_size - 1;
                }
```

source/s3_auto_ranged_get.c
```diff
@@ -651,11 +699,16 @@ static int s_discover_object_range_and_content_length(
...
+            } else {
+                /* When discovering the object size via first-part, the object range is the entire object. */
```
graebm, line 710: utterly trivial: this is probably an old comment, but "via first-part" is a bit confusing, now that we have GET_OBJECT_WITH_PART_NUMBER_1, but this comment is referring to first-part-with-range

source/s3_auto_ranged_get.c
```diff
@@ -739,17 +791,32 @@ static void s_s3_auto_ranged_get_request_finished(
...
                 request->request_tag == AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_PART_NUMBER_1) {
-                /* For now, we can assume that discovery of size via the first part of the object does not apply to
-                 * breaking up a ranged request. If it ever does, then we will need to repopulate this header. */
-                AWS_ASSERT(!auto_ranged_get->initial_message_has_range_header);
-                aws_http_headers_erase(response_headers, g_content_range_header_name);
+
+                if (!auto_ranged_get->initial_message_has_range_header) {
+                    /* content range isn't applicable. */
+                    aws_http_headers_erase(response_headers, g_content_range_header_name);
+                } else {
```
graebm: utterly trivial: in my opinion, code reads nicer as `if(X)/else`, vs `if (!X)/else`

source/s3_auto_ranged_get.c
```diff
@@ -739,17 +791,32 @@ static void s_s3_auto_ranged_get_request_finished(
...
+                        object_range_end,
+                        object_size);
+                    aws_http_headers_set(
+                        response_headers,
+                        g_content_range_header_name,
+                        aws_byte_cursor_from_c_str(content_range_buffer));
+                }
             }
 
+            uint64_t content_length = object_size ? object_range_end - object_range_start + 1 : 0;
```
graebm, line 819: what happens if `object_size` is 0 and `initial_message_has_range_header`? Does that always result in `empty_file_error`?

I just want to be sure the we're fabricating the correct Content-Range header a few lines up

source/s3_util.c
```diff
@@ -488,6 +488,86 @@ int aws_s3_parse_content_length_response_header(
...
+    uint64_t *out_initial_end_range) {
+
+    AWS_PRECONDITION(request_headers);
+    AWS_PRECONDITION(out_initial_message_has_start_range);
+    AWS_PRECONDITION(out_initial_message_has_end_range);
+    AWS_PRECONDITION(out_initial_start_range);
+    AWS_PRECONDITION(out_initial_end_range);
+
+    bool initial_message_has_start_range = false, initial_message_has_end_range = false;
+    uint64_t initial_start_range = 0, initial_end_range = 0;
```
graebm, line 505: I don't think it's written in our coding standards, but never initialize more than one variable per line. It's very error-prone (people forget to initialize one of the earlier variables)
```suggestion
    bool initial_message_has_start_range = false;
    bool initial_message_has_end_range = false;
    uint64_t initial_start_range = 0;
    uint64_t initial_end_range = 0;
```

source/s3_util.c
```diff
@@ -488,6 +488,86 @@ int aws_s3_parse_content_length_response_header(
...
+
+    if (aws_http_headers_get(request_headers, g_range_header_name, &range_header_value)) {
+        return aws_raise_error(AWS_ERROR_S3_INVALID_RANGE_HEADER);
+    }
+
+    struct aws_byte_cursor range_header_start = aws_byte_cursor_from_c_str("bytes=");
+
+    int result = AWS_OP_ERR;
+    /* verify bytes= */
+    if (aws_byte_cursor_starts_with(&range_header_value, &range_header_start)) {
```
graebm, line 517: utterly trivial: early out, instead of indenting the rest of the function

```suggestion
    /* verify bytes= */
    if (!aws_byte_cursor_starts_with(&range_header_value, &range_header_start)) {
        goto done;
    }
```

source/s3_util.c
```diff
@@ -488,6 +488,86 @@ int aws_s3_parse_content_length_response_header(
...
+        *out_initial_start_range = initial_start_range;
+        *out_initial_end_range = initial_end_range;
+        result = AWS_OP_SUCCESS;
+    }
+
+done:
+    if (result != AWS_OP_SUCCESS) {
+        aws_raise_error(AWS_ERROR_S3_INVALID_RANGE_HEADER);
+    }
+    return result;
```
graebm, line 568: trivial: this function doesn't need to do any cleanup ... so not sure what benefit you're getting out of this goto label.  I would do one of these:

1) just do `return aws_raise_error(AWS_ERROR_S3_INVALID_RANGE_HEADER);` everywhere, instead of `goto done`

OR 

2) simplify by just having this be an `error` label:
```
    if (something_unexpected) {
        goto error; /* more obvious than goto done */
    }

    ...
    *out_initial_message_has_start_range = initial_message_has_start_range;
    *out_initial_message_has_end_range = initial_message_has_end_range;
    *out_initial_start_range = initial_start_range;
    *out_initial_end_range = initial_end_range;
    return AWS_OP_SUCCESS;

error:
    return aws_raise_error(AWS_ERROR_S3_INVALID_RANGE_HEADER);
}

```

tests/s3_util_tests.c
```diff
@@ -144,6 +144,121 @@ static int s_test_s3_strip_quotes(struct aws_allocator *allocator, void *ctx) {
...
+    };
+
+    const struct aws_byte_cursor invalid_range_header_values[] = {
+        AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("bytes=-"),
+        AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("bytes=10-5"),
+        AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("byts=0-5"),
+        AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("5-10"),
+    };
+    bool initial_message_has_start_range = false, initial_message_has_end_range = false;
+    uint64_t initial_object_range_start = 0, initial_object_range_end = 0;
```
graebm, line 213: trivial: initialize 1 per line

source/s3_util.c
```diff
@@ -488,6 +488,86 @@ int aws_s3_parse_content_length_response_header(
     return result;
 }
 
+int aws_s3_parse_request_range_header(
+    struct aws_http_headers *request_headers,
+    bool *out_initial_message_has_start_range,
+    bool *out_initial_message_has_end_range,
+    uint64_t *out_initial_start_range,
+    uint64_t *out_initial_end_range) {
```
graebm, line 496: utterly trivial: why bother naming these "initial_message"? it makes the tests weird. This is a standalone function that doesn't rely on state, but naming it "initial" makes the variables seem like we're informing this function about some previous state
```suggestion
    bool *out_has_start_range,
    bool *out_has_end_range,
    uint64_t *out_start_range,
    uint64_t *out_end_range) {
```

tests/s3_util_tests.c
```diff
@@ -144,6 +144,121 @@ static int s_test_s3_strip_quotes(struct aws_allocator *allocator, void *ctx) {
...
+
+    (void)ctx;
+
+    struct range_header_example {
+        struct aws_byte_cursor header_value;
+
+        bool initial_message_has_start_range;
+        bool initial_message_has_end_range;
+        uint64_t initial_object_range_start;
+        uint64_t initial_object_range_end;
```
graebm, line 158: why are the bools "initial_message" but the ints "initial_object"? Is there some subtle difference I'm not picking up on?

(personally, I would omit the "initial_X" prefix from all of them)

graebm (commented):
many trivial comments

source/s3_auto_ranged_get.c
```diff
@@ -236,11 +254,37 @@ static bool s_s3_auto_ranged_get_update(
...
                     case AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_RANGE:
+                        AWS_LOGF_INFO(
+                            AWS_LS_S3_META_REQUEST,
+                            "id=%p: Doing a getObject with range to discover the size of the object",
+                            (void *)meta_request);
+
+                        uint64_t part_range_start = 0;
+                        uint64_t first_part_size = meta_request->part_size;
+                        // TODO: Should we align the first part? I need to run some benchmarks before re-adding first
+                        // part alignment.
```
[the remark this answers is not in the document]
waahm7: We will never get the checksum of the first partial part request, even if we align it to part_size, so we can't validate the checksum. I observed no difference in performance between aligning the first request or not aligning it. Therefore, I have removed the logic for aligning the first part.

source/s3_auto_ranged_get.c
waahm7: It will result in 416 Range not satisfiable error, and we will never trigger the onHeaders as we will fail to find the object_size at https://github.com/awslabs/aws-c-s3/blob/main/source/s3_auto_ranged_get.c#L612.

source/s3_auto_ranged_get.c
waahm7: Thanks, updated.

source/s3_auto_ranged_get.c
waahm7: Thanks, updated.

source/s3_util.c
waahm7: Thanks, updated.

source/s3_util.c
waahm7: Thanks, I have refactored this function to simply return, as it doesn't perform any cleanup.

source/s3_util.c
waahm7: Thanks, I have updated it to first option.

source/s3_util.c
waahm7: Thanks, fixed.

source/s3_auto_ranged_get.c
waahm7: The updated `object_range_end` is used in `aws_min_u64`. I have changed it to an if/else structure and updated that line to directly use `object_size - 1`.

source/s3_meta_request.c
```diff
@@ -1355,7 +1357,7 @@ static int s_s3_meta_request_incoming_body(
     }
 
     if (request->send_data.response_body.capacity == 0) {
-        if (request->has_part_size_response_body) {
+        if (request->has_part_size_response_body && successful_response) {
```
waahm7, line 1360: Ignore the buffer ticket, as the error response might not fit into the originally reserved buffer.

include/aws/s3/private/s3_auto_ranged_get.h
waahm7: Thanks, I have updated it to option 1 for now. I would have preferred option 2 but don't want to include that refactor in this PR.

graebm (approved):

PR #399, opened 2023-12-21
fix two bugs from metrics and cancel

TingDaoK:
*Issue #, if available:*

- The downstream CI caught the bugs.
- One is when we have pending streaming requests in the list, we finishing the meta request by just release the refcount on those requests. But we wait the eventloop to delivery the metrics while they delivery the body streaming for those requests. So, those requests will leak the metrics memory
- The second one is that when the cancel for the http stream was invoked after the stream has already completed, or completing. The error code on complete will not be the error code we asked for canceling. So, we cannot rely on the error code to check for the node has been removed or not. 

*Description of changes:*

- Finish up the metrics from the meta request finish function as well. Double-checked that nothing else will pop from the priority list to deliver the body
- Rely on the `synced_data.http_stream` to track the linked-list node should be removed or not


source/s3_meta_request.c
```diff
@@ -1935,6 +1940,8 @@ void aws_s3_meta_request_finish_default(struct aws_s3_meta_request *meta_request
         struct aws_linked_list_node *request_node = aws_linked_list_pop_front(&release_request_list);
         struct aws_s3_request *release_request = AWS_CONTAINER_OF(request_node, struct aws_s3_request, node);
         AWS_FATAL_ASSERT(release_request != NULL);
+        /* The pending body streaming requests cleaned up here. Finish the metrics for those requests. */
```
graebm, line 1943: trivial: just adding some text that would have helped me understand the complex lifetime
```suggestion
        /* This pending-body-streaming request was never moved to the event-delivery queue,
         * so its metrics were never finished. Finish them now. */
```

source/s3_meta_request.c
```diff
@@ -1385,14 +1385,16 @@ static void s_s3_meta_request_stream_complete(struct aws_http_stream *stream, in
...
-    if (error_code != AWS_ERROR_S3_CANCELED && error_code != AWS_ERROR_S3_PAUSED) {
-        /* BEGIN CRITICAL SECTION */
+    /* BEGIN CRITICAL SECTION */
+    {
         aws_s3_meta_request_lock_synced_data(meta_request);
-        AWS_ASSERT(request->synced_data.http_stream != NULL);
-        aws_linked_list_remove(&request->ongoing_http_requests_list_node);
+        if (request->synced_data.http_stream) {
+            /* The request has been cancelled, and the node has been removed. */
+            aws_linked_list_remove(&request->ongoing_http_requests_list_node);
```
graebm, line 1393: I didn't review the [PR](https://github.com/awslabs/aws-c-s3/pull/395) where this was added, but I think there are some more race conditions...

graebm: https://github.com/awslabs/aws-c-s3/blob/c91f4c3c3bbee8dcd4a49cff0713825aef33e7cb/source/s3_meta_request.c#L1074-L1094

We activate the stream, then acquire the lock, then modify `http_stream` and `ongoing_http_requests_list`.

Once you call activate(), the stream may complete before we acquire the lock. So the stream would try to take itself out of the list. Then we'd put the stream in the list. This probably leads to a leak or dangling reference.

graebm: Another race condition:
- The meta-request is cancelled, so we iterate the `ongoing_http_requests_list` and cancel streams
- Then on another thread we just finished activating a stream and put it into the list, it runs to completion instead of getting canceled.

So from this function we probably want to check whether we should still be making the request, and NOT make the request, or not activate it (is that allowed?), or cancel it instead of adding to list

graebm: also: does `http_stream` solely exist so we can cancel it? If so, let's make that a bit more clear by doing things like:
- set it NULL after removing from `ongoing_http_requests`
- rename it ~http_stream~ -> `ongoing_http_stream` to make the connection to `ongoing_http_requests` more clear

source/s3_meta_request.c
```diff
@@ -1071,28 +1071,41 @@ void aws_s3_meta_request_send_request(struct aws_s3_meta_request *meta_request,
...
-
-        goto error_finish;
-    }
-
     {
         /* BEGIN CRITICAL SECTION */
         aws_s3_meta_request_lock_synced_data(meta_request);
+        if (aws_s3_meta_request_has_finish_result_synced(meta_request) && !request->always_send) {
+            /* The meta request is finishing up, drop the HTTP stream to the ground. */
+            aws_raise_error(meta_request->synced_data.finish_result.error_code);
```
graebm, line 1079: trivial: should we raise some kind of CANCELED error here instead? It's not going to end up as the final meta-request error code, because it already has one. But it might make more sense in the logs to see parts getting torn down with CANCELED errors

source/s3_meta_request.c
```diff
@@ -1071,28 +1071,41 @@ void aws_s3_meta_request_send_request(struct aws_s3_meta_request *meta_request,
...
+            aws_s3_meta_request_unlock_synced_data(meta_request);
+            goto error_finish;
+        }
+
+        /* Activate the stream within the lock as once the activate invoked, the HTTP level callback can happen right
+         * after.  */
+        if (aws_http_stream_activate(stream) != AWS_OP_SUCCESS) {
```
graebm, line 1086: le sigh: I hate that the activate() function can fail
unrelated to this PR (other than it shows how it's annoying)

source/s3_meta_request.c
```diff
@@ -1385,14 +1398,16 @@ static void s_s3_meta_request_stream_complete(struct aws_http_stream *stream, in
...
     }
-    if (error_code != AWS_ERROR_S3_CANCELED && error_code != AWS_ERROR_S3_PAUSED) {
-        /* BEGIN CRITICAL SECTION */
+    /* BEGIN CRITICAL SECTION */
+    {
         aws_s3_meta_request_lock_synced_data(meta_request);
-        AWS_ASSERT(request->synced_data.http_stream != NULL);
-        aws_linked_list_remove(&request->ongoing_http_requests_list_node);
+        if (request->synced_data.ongoing_http_stream) {
+            /* The request has been cancelled, and the node has been removed. */
```
graebm, line 1405: this comment is confusing. the request is complete but it wasn't necessarily canceled.
and we're about to remove the node, but the comment make it sound like it was already removed

delete the comment entirely, the code makes enough sense now, or replace with something like
```suggestion
            /* Remove from list of ongoing requests */
```

source/s3_meta_request.c
```diff
@@ -1677,12 +1692,34 @@ void aws_s3_meta_request_cancel_ongoing_http_requests_synced(struct aws_s3_meta_
             AWS_CONTAINER_OF(request_node, struct aws_s3_request, ongoing_http_requests_list_node);
         if (!request->always_send) {
             /* Cancel the ongoing http stream, unless it's always send. */
-            aws_http_stream_cancel(request->synced_data.http_stream, error_code);
+            aws_http_stream_cancel(request->synced_data.ongoing_http_stream, error_code);
         }
-        request->synced_data.http_stream = NULL;
+        request->synced_data.ongoing_http_stream = NULL;
```
graebm, line 1697: debatable: seems accident-prone that we put `always_send` requests in this list, but then we need to be careful about not canceling.

Maybe rename ~ongoing_http_stream~ and ~ongoing_http_requests_list~ to `cancelable_http_stream` and `cancelable_http_requests_list`, and `always_send` requests shouldn't touch those variables at all

source/s3_meta_request.c
TingDaoK: well, at least we can still invoke it within the lock, positive sigh

source/s3_meta_request.c
TingDaoK: yeah, updated the code, but I forgot to update the comment around it ...

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -226,8 +226,8 @@ struct aws_s3_meta_request {
         /* True if the finish result has been set. */
         uint32_t finish_result_set : 1;
 
-        /* To track the aws_s3_request that are active from HTTP level */
-        struct aws_linked_list ongoing_http_requests_list;
+        /* To track the aws_s3_request that are cancelled from HTTP level */
```
graebm, line 229: trivial: comment sounds like it tracks requests that have already been cancelled
```suggestion
        /* To track aws_s3_requests with cancellable HTTP streams */
```

include/aws/s3/private/s3_request.h
```diff
@@ -114,12 +114,12 @@ struct aws_s3_request {
...
 
     /* Linked list node used for tracking the request is active from HTTP level. */
-    struct aws_linked_list_node ongoing_http_requests_list_node;
+    struct aws_linked_list_node cancelable_http_streams_list_node;
 
     /* The meta request lock must be held to access the data */
     struct {
         /* The underlying http stream, only valid when the request is active from HTTP level */
-        struct aws_http_stream *http_stream;
+        struct aws_http_stream *ongoing_http_stream;
```
graebm, line 122:
```suggestion
        struct aws_http_stream *cancellable_http_stream;
```

source/s3_meta_request.c
```diff
@@ -1071,28 +1071,51 @@ void aws_s3_meta_request_send_request(struct aws_s3_meta_request *meta_request,
...
         aws_linked_list_push_back(
-            &meta_request->synced_data.ongoing_http_requests_list, &request->ongoing_http_requests_list_node);
-        request->synced_data.http_stream = stream;
+            &meta_request->synced_data.cancelable_http_streams_list, &request->cancelable_http_streams_list_node);
+        request->synced_data.ongoing_http_stream = stream;
+
         aws_s3_meta_request_unlock_synced_data(meta_request);
         /* END CRITICAL SECTION */
+    } else {
+        /* If the request always send, it is not cancellable, we simply active the stream. */
```
graebm, line 1102:
```suggestion
        /* If the request always send, it is not cancellable, we simply activate the stream. */
```

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -226,8 +226,8 @@ struct aws_s3_meta_request {
...
+        struct aws_linked_list cancelable_http_streams_list;
```
graebm, line 230: I deeply apologize, but I gave a bad name suggestion. Apparently it's more common to spell this with two L's

Here is a [Cancellable](https://developer.apple.com/documentation/combine/cancellable) trait in Swift. And Apple is pretty damned choosy about names

include/aws/s3/private/s3_request.h
```diff
@@ -114,12 +114,12 @@ struct aws_s3_request {
     struct aws_linked_list_node node;
 
     /* Linked list node used for tracking the request is active from HTTP level. */
-    struct aws_linked_list_node ongoing_http_requests_list_node;
+    struct aws_linked_list_node cancelable_http_streams_list_node;
```
graebm, line 117:
```suggestion
    struct aws_linked_list_node cancellable_http_streams_list_node;
```

source/s3_meta_request.c
```diff
@@ -1385,14 +1408,15 @@ static void s_s3_meta_request_stream_complete(struct aws_http_stream *stream, in
...
-    if (error_code != AWS_ERROR_S3_CANCELED && error_code != AWS_ERROR_S3_PAUSED) {
-        /* BEGIN CRITICAL SECTION */
+    /* BEGIN CRITICAL SECTION */
+    {
         aws_s3_meta_request_lock_synced_data(meta_request);
-        AWS_ASSERT(request->synced_data.http_stream != NULL);
-        aws_linked_list_remove(&request->ongoing_http_requests_list_node);
+        if (request->synced_data.ongoing_http_stream) {
+            aws_linked_list_remove(&request->cancelable_http_streams_list_node);
+        }
```
graebm, line 1416: null out the stream pointer
```suggestion
        if (request->synced_data.cancellable_http_stream) {
            aws_linked_list_remove(&request->cancellable_http_streams_list_node);
            request->synced_data.cancellable_http_stream = NULL;
        }
```

source/s3_meta_request.c
```diff
@@ -1670,17 +1694,38 @@ bool aws_s3_meta_request_are_events_out_for_delivery_synced(struct aws_s3_meta_r
...
-            aws_linked_list_pop_front(&meta_request->synced_data.ongoing_http_requests_list);
+            aws_linked_list_pop_front(&meta_request->synced_data.cancelable_http_streams_list);
         struct aws_s3_request *request =
-            AWS_CONTAINER_OF(request_node, struct aws_s3_request, ongoing_http_requests_list_node);
-        if (!request->always_send) {
-            /* Cancel the ongoing http stream, unless it's always send. */
-            aws_http_stream_cancel(request->synced_data.http_stream, error_code);
+            AWS_CONTAINER_OF(request_node, struct aws_s3_request, cancelable_http_streams_list_node);
+        AWS_ASSERT(!request->always_send);
+        /* Cancel the ongoing http stream, unless it's always send. */
```
graebm, line 1703: comment about "unless it's always send" not necessary anymore
```suggestion

```

graebm (approved):

PR #401, opened 2023-12-22
Fix get-object-part

TingDaoK:
*Issue #, if available:*
- The "partNumber" is the query parameter instead of header

*Description of changes:*
- Properly check for partNumber, use default type if partNumber exist.


source/s3_client.c
```diff
@@ -1168,20 +1168,28 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
...
+            struct aws_byte_cursor path_and_query;
+
+            if (aws_http_message_get_request_path(options->message, &path_and_query) == AWS_OP_SUCCESS) {
+                /* If the initial request already has partNumber, the request is not
+                 * splittable(?). Treat it as a Default request.
+                 * TODO: Still need tests to verify that the request of a part is
+                 * splittable or not */
+                struct aws_byte_cursor found_string;
+                struct aws_byte_cursor part_number_query_str = aws_byte_cursor_from_c_str("partNumber");
+                if (aws_byte_cursor_find_exact(&path_and_query, &part_number_query_str, &found_string) ==
```
graebm, line 1180: find isn't going to cut it. This would match: "rampartNumbering.s3.us-west-2.amazonaws.com"

I'm sure we have query param helpers somewhere in our codebase

source/s3_client.c
```diff
@@ -1168,20 +1168,28 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
...
+                        client,
+                        AWS_S3_REQUEST_TYPE_GET_OBJECT,
+                        content_length,
+                        false /*should_compute_content_md5*/,
+                        options);
+                }
             }
-
+            /* Ignore any error while we try to find the partNumber query parameter. */
+            aws_reset_error();
```
graebm, line 1192: just curious, does this matter? we don't make any guarantees that the error on the stack is non-zero.

We should not have any code that checks whether aws_last_error() to determine whether or not an error has occurred. 

If you were doing this because it led to some later code reporting a stale error, that later code is buggy and should be fixed.

source/s3_client.c
TingDaoK: I tried to find one, but there is only a uri parsing with query param, which requires the whole uri, I can add something to the uri.h to support it

source/s3_client.c
TingDaoK: Not matter, will remove it

source/s3_client.c
TingDaoK: https://github.com/awslabs/aws-c-common/pull/1080

source/s3_client.c
```diff
@@ -1168,20 +1168,34 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
...
+                        if (aws_byte_cursor_eq(&param.key, &part_number_query_str)) {
+                            return aws_s3_meta_request_default_new(
+                                client->allocator,
+                                client,
+                                AWS_S3_REQUEST_TYPE_GET_OBJECT,
+                                content_length,
+                                false /*should_compute_content_md5*/,
+                                options);
+                        }
+                    }
```
waahm7, line 1196: Debatable: I think it would be better to invest in a better helper like `aws_http_headers_has` or `aws_http_headers_get`. This logic will need to be repeated wherever we need to extract a single parameter from query params.

Or, we could still use `find` and try to locate `&partNumber=` or `?partNumber=`.

TingDaoK: I have no strong feelings about it.

My take is we are used as a client level http lib, and this is the first time we ever really want to parse the query params, maybe let's hold it back until the next time we want to reuse it?

source/s3_client.c
```diff
@@ -1168,20 +1168,34 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
...
-                    content_length,
-                    false /*should_compute_content_md5*/,
-                    options);
+            struct aws_byte_cursor path_and_query;
+
+            if (aws_http_message_get_request_path(options->message, &path_and_query) == AWS_OP_SUCCESS) {
+                /* If the initial request already has partNumber, the request is not
+                 * splittable(?). Treat it as a Default request.
+                 * TODO: Still need tests to verify that the request of a part is
+                 * splittable or not */
```
graebm, line 1177: is this TODO still accurate? you added a test. can we remove this?

graebm (approved):

waahm7 (approved):

source/s3_client.c
TingDaoK: we still need that TODO about splitting a part.

DmitriyMusatkin (approved):

PR #407, opened 2024-02-28
Stop limiting num-connections based on num-known-IPs (Improve S3-Express performance)

graebm:
**Issue:**
Disappointing S3-Express performance.

**Description of changes:**
Stop limiting num-connections based on num-known-IPs.

**Diagnosing the issue:**
We found that num-connections was never getting very high, because [num-connections scales based on the num-known-IPs](https://github.com/awslabs/aws-c-s3/blob/593c2ab24608d3e78708d51657be22f6ab99cb50/source/s3_client.c#L179). S3-Express endpoints have very few IPs, so their num-connections weren't scaling very high.

The algorithm was adding 10 connections per known-IP. On a 100Gb/s machine, this maxed out at 250 connections once 25 IPs were known. But S3-Express endpoints only have 4 unique IPs, so they never got higher than 40 connections.

This algorithm was written back when S3 returned 1 IP per DNS query. The intention was to throttle connections until more IPs were known, in order to spread load among S3's server fleet. However, as of Aug 2023 [S3 provides multiple IPs per DNS query](https://aws.amazon.com/about-aws/whats-new/2023/08/amazon-s3-multivalue-answer-response-dns-queries/). So now, we can scale up to max connections after the first DNS query and still be spreading load.

We also believed that spreading load was a key to good performance. But I found that spreading the load didn't have much impact on performance (at least now, in 2024, on the 100Gb/s machine I was using). Tests where I hard-coded a single IP and hit it with max-connections didn't differ much from tests where the load was spread among 8 IPs or 100 IPs.

I want to get this change out quickly and help S3-Express, so I picked magic numbers where the num-connections math ends up with the same result as the old algorithm. Normal S3 performance is mildly improved (max-connections is reached immediately, instead of scaling up over 30sec as it finds more IPs). S3 Express performance is MUCH improved.

**Future Work:**
Improve this algorithm further:
- expect higher throughput on connections to S3 Express
- expect lower throughput on connections transferring small objects
- dynamic scaling without a bunch of magic numbers ??? (sounds cool, but I don't have any ideas how this would work yet)

### DIFF
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -64,7 +64,7 @@ static const uint32_t s_max_requests_multiplier = 4;
  * TODO: Improve this algorithm (expect higher throughput for S3 Express,
  * expect lower throughput for small objects, etc)
  */
-static const double s_throughput_per_connection_gbps = 100.0 / 160;
+static const double s_throughput_per_connection_gbps = 100.0 / 250;
 
 /* After throughput math, clamp the min/max number of connections */
 const uint32_t g_min_num_connections = 10;     /* Magic value based on: 10 was old behavior */
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -51,21 +51,24 @@ struct aws_s3_meta_request_work {
 
 static const enum aws_log_level s_log_level_client_stats = AWS_LL_INFO;
 
+/* max-requests-in-flight = ideal-num-connections * s_max_requests_multiplier */
 static const uint32_t s_max_requests_multiplier = 4;
 
-/* TODO Provide analysis on origins of this value. */
-static const double s_throughput_per_vip_gbps = 4.0;
+/* This is used to determine the ideal number of HTTP connections. Algorithm is roughly:
+ * ideal-num-connections = throughput-target-gpbs / s_throughput_per_connection_gbps
+ *
+ * Magic value based on:
+ * We found 160 connections gave best performance for 30 GiB download
+ * (3840 8MiB parts) on 100Gbps c5n.18xlarge.
+ *
+ * TODO: Improve this algorithm (expect higher throughput for S3 Express,
+ * expect lower throughput for small objects, etc)
+ */
+static const double s_throughput_per_connection_gbps = 100.0 / 160;
 
-/* Preferred amount of active connections per meta request type. */
-const uint32_t g_num_conns_per_vip_meta_request_look_up[AWS_S3_META_REQUEST_TYPE_MAX] = {
-    10, /* AWS_S3_META_REQUEST_TYPE_DEFAULT */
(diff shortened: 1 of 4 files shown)


source/s3_client.c
```diff
@@ -151,32 +152,9 @@ uint32_t aws_s3_client_get_max_active_connections(
     struct aws_s3_client *client,
     struct aws_s3_meta_request *meta_request) {
     AWS_PRECONDITION(client);
+    (void)meta_request;
```
TingDaoK, line 155: why not just remove `meta_request` from the args, it's a private helper.

Or do we even need this helper? Can we just have it as a member of client? It won't change after the client initialized now, right?

source/s3_client.c
```diff
@@ -1687,7 +1669,7 @@ static bool s_s3_client_should_update_meta_request(
     size_t num_known_vips = client->vtable->get_host_address_count(
         client->client_bootstrap->host_resolver, endpoint->host_name, AWS_GET_HOST_ADDRESS_COUNT_RECORD_TYPE_A);
     if (num_known_vips == 0 && (client->threaded_data.num_requests_being_prepared +
-                                client->threaded_data.request_queue_size) >= g_max_num_connections_per_vip) {
+                                client->threaded_data.request_queue_size) >= g_min_num_connections) {
```
TingDaoK, line 1672: do we still want this?

TingDaoK (commented):
> DNS Load Balancing: DNS resolver continuously harvests Amazon S3 IP addresses. When load is spread across the S3 fleet, overall throughput is better than if all connections were hammering the same IP simultaneously.

Should we update the readme [here](https://github.com/awslabs/aws-c-s3?tab=readme-ov-file#key-features)?

source/s3_client.c
graebm: I'm pretty sure we'll be using meta_request again in the near future. If we're going to use different magic numbers for S3-Express vs S3-Vanilla, there will be some `meta-request->storage_class` variable that we'll need to check

source/s3_client.c
graebm: I don't know yes/no for certain. I wanted to limit the extent of these changes so I'm keeping the old behavior where we limit the number of requests in the prep stage when we don't know *any* IP addresses yet.

I also wondered if this was added in response to a specific issue, or just something that was added superstitiously. The check was added in this PR: [Multiple Bucket Support](https://github.com/awslabs/aws-c-s3/pull/136). There's a specific commit ([Preventing too many requests from being prepared](https://github.com/awslabs/aws-c-s3/pull/136/commits/19fd44f3be0e073b648d28ea0609031e25c04cf7)) where Ryan adds that adds this check. So it seems pretty deliberate.

So maybe this could be removed? Or the numbers tweaked? We'll address that another day

TingDaoK (approved):

PR #412, opened 2024-03-18
Implements Connection Pool Destroy Delay

waahm7:
*Description of changes:*
**Problem:**
If the user makes requests serially, we destroy the connection pool between each request, which impacts performance.

**Solution:**
Add a 3-second delay before destroying the connection pool, unless the Client is also being destroyed. If the user makes another request within those 3 seconds, the connection pool is not destroyed.

Benchmark:
Downloading 1000 64Kb files in serial on c5n.18xlarge
```
Main Branch:
Run:1 Secs:33.994031 Gb/s:0.015423
Run:2 Secs:34.439879 Gb/s:0.015223
Run:3 Secs:35.191540 Gb/s:0.014898
Run:4 Secs:34.023091 Gb/s:0.015410
Run:5 Secs:34.815079 Gb/s:0.015059
Run:6 Secs:35.207595 Gb/s:0.014891
Run:7 Secs:36.181037 Gb/s:0.014491
Run:8 Secs:35.388686 Gb/s:0.014815
Run:9 Secs:35.315042 Gb/s:0.014846
Run:10 Secs:37.175047 Gb/s:0.014103
Overall Throughput (Gb/s) Median:0.014895 Mean:0.014916 Min:0.014103 Max:0.015423 Variance:0.000000 StdDev:0.000384
Overall Duration (Secs) Median:35.199567 Mean:35.173103 Min:33.994031 Max:37.175047 Variance:0.847040 StdDev:0.920348
Peak RSS:20.632812 MiB

This PR:
Run:1 Secs:21.822355 Gb/s:0.024025
Run:2 Secs:18.661068 Gb/s:0.028095
Run:3 Secs:16.470886 Gb/s:0.031831
Run:4 Secs:16.819767 Gb/s:0.031171
Run:5 Secs:15.395681 Gb/s:0.034054
Run:6 Secs:16.630556 Gb/s:0.031526
Run:7 Secs:16.218798 Gb/s:0.032326
Run:8 Secs:16.514378 Gb/s:0.031747
Run:9 Secs:17.100299 Gb/s:0.030660
Run:10 Secs:16.876424 Gb/s:0.031066
Overall Throughput (Gb/s) Median:0.031348 Mean:0.030650 Min:0.024025 Max:0.034054 Variance:0.000007 StdDev:0.002620
Overall Duration (Secs) Median:16.725162 Mean:17.251021 Min:15.395681 Max:21.822355 Variance:2.927835 StdDev:1.711092
Peak RSS:20.628906 MiB

```


source/s3_client.c
```diff
@@ -1528,9 +1528,49 @@ static void s_s3_client_process_work_default(struct aws_s3_client *client) {
...
+
+    /* BEGIN CRITICAL SECTION */
+    aws_s3_client_lock_synced_data(client);
+
+    if (client->synced_data.process_endpoint_lifecycle_changes) {
+        client->synced_data.process_endpoint_lifecycle_changes = false;
+        for (struct aws_hash_iter iter = aws_hash_iter_begin(&client->synced_data.endpoints);
+             !aws_hash_iter_done(&iter);
+             aws_hash_iter_next(&iter)) {
+            struct aws_s3_endpoint *endpoint = (struct aws_s3_endpoint *)iter.element.value;
```
waahm7, line 1543: An alternative approach that I considered was maintaining two linked lists of endpoints: `pending_cleanup_task` and `cleanup_task_scheduled`, which would allow us to only loop through endpoints that have the required state. However, considering that by default an AWS account has a limit of 100 buckets, I thought it was not worth the complexity.

source/s3_endpoint.c
```diff
@@ -278,11 +317,13 @@ static void s_s3_endpoint_release(struct aws_s3_endpoint *endpoint) {
...
-         * Since it's a crime to hold a lock while invoking a callback,
-         * we make sure that we've released the client's lock before proceeding... */
-        s_s3_endpoint_ref_count_zero(endpoint);
+        if (client_active) {
+            /* schedule the cleanup task */
+            aws_s3_client_schedule_process_work(endpoint->client);
+        } else {
+            /* do a sync cleanup since client is not active anymore to avoid any cleanup delay */
+            s_s3_endpoint_ref_count_zero(endpoint);
+        }
```
waahm7, line 326: Question/Confusion: Is it worth avoiding `aws_event_loop_schedule_task_now` here? If we simplify the code here to always trigger `aws_s3_client_schedule_process_work`, we will clean up the endpoint on the client process work thread, but does it matter?

source/s3_endpoint.c
```diff
@@ -267,9 +299,16 @@ static void s_s3_endpoint_release(struct aws_s3_endpoint *endpoint) {
...
+    bool should_destroy = endpoint->client_synced_data.ref_count == 1;
+    bool client_active = endpoint->client->synced_data.active == 1;
     if (should_destroy) {
-        aws_hash_table_remove(&endpoint->client->synced_data.endpoints, endpoint->host_name, NULL, NULL);
+        if (client_active) {
+            endpoint->client_synced_data.state = AWS_S3_ENDPOINT_STATE_PENDING_CLEANUP_TASK;
+            endpoint->client->synced_data.process_endpoint_lifecycle_changes = true;
+        } else {
+            endpoint->client_synced_data.state = AWS_S3_ENDPOINT_STATE_DESTROYING;
+            aws_hash_table_remove(&endpoint->client->synced_data.endpoints, endpoint->host_name, NULL, NULL);
```
waahm7, line 310: Question/Confusion: Is it worth avoiding aws_event_loop_schedule_task_now here? If we simplify the code here to always trigger aws_s3_client_schedule_process_work, we will clean up the endpoint on the client process work thread, but does it matter?

source/s3_client.c
```diff
@@ -1528,9 +1528,51 @@ static void s_s3_client_process_work_default(struct aws_s3_client *client) {
...
```
DmitriyMusatkin, line 1543: this seems a bit scary: this can create 10s of things that need to happen on event loop. we also try to avoid locking this client too much from this thread as it can block other threads from doing their work. 

do we actually need every endpoint to have a separate termination event? can we simplify the logic and have a simple event that repeats every n seconds and cleans up whatever endpoints do not have work? (i dont think we care if endoint gets cleaned up in 3s exactly or worst case in 6s). we do something similar for dns from what i remember

waahm7: Thanks, Created a new https://github.com/awslabs/aws-c-s3/pull/419 with the new logic.

PR #413, opened 2024-03-18
Revert "Work around asan issue with GH runner image >= 20240310.1.0 (…

TingDaoK:
…#409)"

This reverts commit 6ab26a431f17bf5c1fabc6f7ef2045575ee75b87.

*Issue #, if available:*

*Description of changes:*

Revert the temporary workaround. 
GH action plan to release the fix on Mar 22, 2024, according to [here](https://github.com/actions/runner-images/issues/9491#issuecomment-1999433974). After that, the workaround is not needed, and should be removed. 

Preparing the PR for the change. We can merge it once GH action releases the fix.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -20,12 +20,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout Sources
-        uses: actions/checkout@v4
-      - name: Fix kernel mmap rnd bits
-        # High entropy setting in GH runner images >= 20240310.1.0
-        # causes ASAN blowing up here and there:
-        # https://github.com/actions/runner-images/issues/9491
-        run: sudo sysctl vm.mmap_rnd_bits=28
+        uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/codecov.yml
```diff
@@ -20,12 +20,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout Sources
-        uses: actions/checkout@v4
-      - name: Fix kernel mmap rnd bits
-        # High entropy setting in GH runner images >= 20240310.1.0
-        # causes ASAN blowing up here and there:
-        # https://github.com/actions/runner-images/issues/9491
-        run: sudo sysctl vm.mmap_rnd_bits=28
+        uses: actions/checkout@v3
```
waahm7, line 23: We can probably keep v4?

waahm7 (approved):

.github/workflows/codecov.yml
TingDaoK: I just wanna revert the whole commit, it's fine to not updating the version for checkout action, we still use v2 some where.

PR #414, opened 2024-03-19
UserAgent Updates

waahm7:
*Description of changes:*
Adds platform/ec2InstanceType if available to the user agent header. Here is new header value:
```
 UserAgentHeader:CRTS3NativeClient/0.1.x platform/c5n.18xlarge
 ```
 or
 ```
  UserAgentHeader:CRTS3NativeClient/0.1.x platform/unknown
 ```

### CODE REVIEW

File: source/s3_util.c [hunk 1]
```diff
@@ -352,9 +355,13 @@ void aws_s3_add_user_agent_header(struct aws_allocator *allocator, struct aws_ht
 
     const struct aws_byte_cursor space_delimiter = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL(" ");
     const struct aws_byte_cursor forward_slash = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("/");
-
-    const size_t user_agent_product_version_length =
-        g_user_agent_header_product_name.len + forward_slash.len + g_s3_client_version.len;
+    struct aws_byte_cursor platform_cursor = aws_s3_get_current_platform_info()->instance_type;
```
waahm7, line 358: Discussion point: This call can result in an imds call for one time. Should we define a new API which returns instance type only if we know it already?

waahm7: I have updated it to only used cached results.

source/s3_platform_info.c
```diff
@@ -554,9 +554,12 @@ struct aws_byte_cursor aws_s3_get_ec2_instance_type(struct aws_s3_platform_info_
...
-    struct aws_s3_platform_info_loader *loader) {
+    struct aws_s3_platform_info_loader *loader,
+    bool cached_only) {
     /* getting the instance type will set it on the loader the first time if it can */
-    aws_s3_get_ec2_instance_type(loader);
+    if (!cached_only) {
+        aws_s3_get_ec2_instance_type(loader);
+    }
     /* will never be mutated after the above call. */
     return &loader->lock_data.current_env_platform_info;
```
TingDaoK, line 564: hmmmm, we are touching the lock_data. I haven't dig deep into the implementation, but the comment says "will never be mutated after the above call", but, we can now skip `aws_s3_get_ec2_instance_type`, will it be a possible race condition?

waahm7: Thanks, in theory, yes. In practice, this would not have occurred since the flow is that this API is called first to determine target_throughput, and the S3 client is created later in the Python bindings. I have updated the method to use an API with locks instead, since it can have a race condition in theory.

TingDaoK (approved):

PR #417, opened 2024-03-20
Partial fix: Too many "stalled" uploads can deadlock the S3 Client

graebm:
**Issue:**
Mountpoint (which wraps aws-c-s3 with a filesystem-like API) had a user that opened 100+ files at once. The user wrote data to some of the later files they opened, and waited for those writes to complete. But aws-c-s3 was waiting on data from the first few files. Both sides were waiting on each other. It was a deadlock.

**Description of changes:**
* Add regression test that demonstrates the issue.
* Add `aws_s3_meta_request_options.maximize_async_stream_reads_internal_use_only` setting that increases the number of "stalled" uploads the S3 client can tolerate.
    * Instead being limited by maximum connections (25, given default target throughput), it will be limited by `memory_limit_in_bytes` (240, given 64bit platform with default values for target throughput, memory limit, and part size).
    * This option will be removed when a better solution is developed. Likely it will be some kind of `aws_s3_meta_request_write_async()` call, which lets the user tell the S3 client when data is available. Vs the current situation with async-stream, where the S3 client is telling the user when it's a good time to provide the data.


waahm7 (approved):
> that increases the number of "stalled" uploads the S3 client can tolerate, from 25 -> 240. 

It may be worth mentioning that what values of target_throughput resulted in 25, and what the values of mem limit and part_size resulted in 240 for future readers.

DmitriyMusatkin (approved):

TingDaoK (approved):
just curious, why 240? Should it be 2GB/8MB -> 250?

maybe worth to mentioned the default memory limits is based on the target throughput

graebm:
I updated the PR description with some more description of where those numbers came from

PR #418, opened 2024-03-26
Add async write() function - fixes "stalled" uploads deadlocking S3 Client

graebm:
**Issue:**
The S3 client can deadlock if too many uploads are "stalled" (not providing data when  the S3 client asks for it).

Mountpoint (which wraps aws-c-s3 with a filesystem-like API) had a user that opened 100+ files at once. The user wrote data to some of the later files they opened, and waited for those writes to complete. But aws-c-s3 was waiting on data from the first few files. Both sides were waiting on each other. It was a deadlock.

**Description of changes:**
- Add async `aws_s3_meta_request_write()` function.
    - Mountpoint should use this, instead of the async-streams. Async-streams were still "pull based", where the S3 client decided when data should be provided. The new write() function is push-based. Mountpoint can send data whenever. "Stalled" uploads will not deadlock the S3 client.
- Revert https://github.com/awslabs/aws-c-s3/pull/417
    - This workaround didn't help all that much. The S3 client still deadlocked at the same value of "N stalled uploads" if your workloads also mixed in other types of non-upload meta requests.

**Design:**
The memory-ownership for this API evolved over time. Here were the stages:
1) At first, the user was required to keep memory valid until the async-write completed. (as of commit [5614546](https://github.com/awslabs/aws-c-s3/tree/561454640119ef38513e31103676ecb7271d5831))
    * PRO: No locks held while memcpy'ing data into a part. Good memory usage, only copy the data when it's ready to send.
    * CON: no way for user to drop memory without canceling the meta-request and blocking their thread until the write completed, which risks deadlock if blocking on a thread the meta-request needs to use.
2) Next, we had write() copy all memory immediately (as of commit [5e4dd89](https://github.com/awslabs/aws-c-s3/tree/5e4dd89e2cb6bbdc92e337b9e845e2032edcb602))
    * PRO: user does not need to keep memory alive after write() call
    * CON: a huge write immediately forces a huge allocation. Writing 10GiB means at least 20GiB total memory usage during the write() call. Users would end up needing to chunk up large work themselves, rather than let aws-c-s3 handle it.
3) Finally, go back to approach 1, but guarantee that canceling a meta-request will **immediately** complete any pending writes.
    * PRO: Back to good memory usage
    * CON: Need to hold a lock when memcpy'ing data into a part.

Some quick benchmarking showed similar performance from all 3 approaches (tested 30GiB upload, and simultaneous upload of 100 5GiB objects, calling write in 256KiB chunks), so 🤷‍♀️. Going with approach #3 since it doesn't risk deadlock, and doesn't risk enormous memory usage.

**TODO:**
- Improve buffering of small writes (use buffer pool?)
- Get async-write working with Content-Length (currently only works when Content-Length omitted)
- Get async-write working with DEFAULT meta requests (currently only works with PUT)


source/s3_meta_request.c
```diff
@@ -2155,6 +2169,158 @@ void aws_s3_meta_request_result_setup(
...
+        /* BEGIN CRITICAL SECTION */
+        aws_s3_meta_request_lock_synced_data(meta_request);
+        write_future_to_complete = meta_request->async_write.synced_future;
+        meta_request->async_write.synced_future = NULL;
+        aws_s3_meta_request_unlock_synced_data(meta_request);
+        /* END CRITICAL SECTION */
+
+        if (meta_request->async_write.eof) {
+            *eof = true;
+        }
```
TingDaoK, line 2311: trivial: we should move this before we enter the lock and set the `synced_future` to NULL.
Otherwise, the next call is possible to change the eof (I know we haven't set the future result yet, but the protection around the write depends on the future is set or not)

source/s3_meta_request.c
```diff
@@ -2081,13 +2098,25 @@ struct aws_future_bool *aws_s3_meta_request_read_body(
         return aws_parallel_input_stream_read(meta_request->request_body_parallel_stream, offset, buffer);
     }
 
+    /* Further techniques are synchronous... */
+    struct aws_future_bool *synchronous_read_future = aws_future_bool_new(meta_request->allocator);
+
+    /* If using async-writes, call function which fills the buffer and/or hits EOF  */
+    if (meta_request->request_body_using_async_writes == true) {
+        bool eof = false;
+        if (s_s3_meta_request_read_from_pending_async_writes(meta_request, buffer, &eof) == AWS_OP_SUCCESS) {
```
TingDaoK, line 2107: `s_s3_meta_request_read_from_pending_async_writes` doesn't seem to return anything other than AWS_OP_SUCCESS. Maybe we can make it void instead

source/s3_meta_request.c
graebm: Nice catch! I noticed this too when I made the buffer-always change just now

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -256,6 +253,45 @@ struct aws_s3_meta_request {
         struct aws_array_list event_delivery_array;
     } io_threaded_data;
 
+    /**
+     * Data for async-writes.
+     * Currently, we only allow 1 async-write at a time.
```
DmitriyMusatkin, line 258: per meta req or in total?

source/s3_meta_request.c
graebm: I wasn't sure whether to turn some of the asserts into actual errors ...
but then I hate when I see error-handling for things that shouldn't be possible, it makes me wonder if they really are possible and I should be handling stuff like that everywhere

let's live bold: I'll make them FATAL_ASSERTS

include/aws/s3/s3_client.h
```diff
@@ -829,6 +834,46 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
...
+ * (e.g. received a 404 error response). AWS_ERROR_INVALID_STATE usually
+ * indicates that you're calling write() incorrectly (e.g. not waiting for
+ * previous write to complete).
+ *
+ * You can wait any length of time between calls to write().
+ * Data is buffered until there is enough to upload.
+ *
+ * @param meta_request  Meta request
+ *
+ * @param data          The data to send. This data can be any size.
```
DmitriyMusatkin, line 860: nit: maybe mention that data behind cursor needs to be valid for duration of function call

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -256,6 +253,45 @@ struct aws_s3_meta_request {
...
+     *
+     * When the user calls write(), they may not provide enough data for us to
+     * send an UploadPart. In that case, we immediately mark the write complete,
```
DmitriyMusatkin, line 261: do we do createMPU right away or wait for at least a block full of data?

DmitriyMusatkin: possible optimization for future is to delay until we get full part or eof and then decide on whether to do default or mpu

source/s3_meta_request.c
```diff
@@ -271,17 +280,12 @@ int aws_s3_meta_request_init_base(
...
         meta_request->request_body_async_stream = aws_async_input_stream_acquire(options->send_async_stream);
-        meta_request->initial_request_message = aws_http_message_acquire(options->message);
-        meta_request->maximize_async_stream_reads = options->maximize_async_stream_reads_internal_use_only;
 
-    } else {
-        /* Keep original message around, we'll read from its synchronous body-stream */
-        meta_request->initial_request_message = aws_http_message_acquire(options->message);
+    } else if (options->send_using_async_writes == true) {
+        meta_request->request_body_using_async_writes = true;
+        aws_byte_buf_init(&meta_request->async_write.data_buffer, allocator, 0);
```
DmitriyMusatkin, line 288: is 0 a good starting size for this?

include/aws/s3/s3_client.h
```diff
@@ -829,6 +834,46 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
...
+ *                      Do not call write() again after passing true.
+ *
+ * This function never returns NULL.
+ *
+ * WARNING: This feature is experimental.
+ */
+AWS_S3_API
+struct aws_future_void *aws_s3_meta_request_write(
+    struct aws_s3_meta_request *meta_request,
+    struct aws_byte_cursor data,
```
DmitriyMusatkin, line 874: are there any limits on how bit data can be?

source/s3_meta_request.c
```diff
@@ -2155,6 +2169,160 @@ void aws_s3_meta_request_result_setup(
...
+        /* We've sent all the data */
+        meta_request->async_write.data_buffer.len = 0;
+
+        if (meta_request->async_write.eof) {
+            *eof = true;
+        }
+    } else {
+        /* There's leftover data. Copy it to the front of the buffer.
+         * We'll send it after we get more writes. */
+        memcpy(
```
DmitriyMusatkin, line 2300: concern: potential ub with overlapping src and dest?

include/aws/s3/private/s3_meta_request_impl.h
graebm: per meta request
updated

include/aws/s3/private/s3_meta_request_impl.h
graebm: Currently, we're always doing CreateMultipartUpload. Yeah, agree that's a future optimization we could do for this (or any PUT where we don't know Content-Length)

source/s3_meta_request.c
graebm: Yeah, 0 is a good starting size. We'll allocate when they finally get around to calling write(). In my experience, that's the way to go. It's totally worth debating the optimal *growth* strategy for a buffer once it has non-zero bytes, but that's for the followup PR

include/aws/s3/s3_client.h
graebm: nope. I'm on the fence about limits. I added to the docs about the dangers of large writes. I'm going to pause agonizing over this since we're still debating the always-copy strategy...

source/s3_meta_request.c
graebm: We only get here if we had 1+ parts worth of data, and we're only doing the memcpy if there's `< 1` part's worth of data leftover, should we *shouldn't* have a problem.

But I didn't know about that undefined behavior. I"ll add some asserts

include/aws/s3/s3_client.h
graebm: Aaaand we're not doing always-copy anymore, so this isn't an issue

DmitriyMusatkin (approved):

PR #421, opened 2024-03-29
Allow disabling upload checksums while using upload review

jamesbornholt:
Some clients (Mountpoint) might want to disable actually sending upload
checksums when talking to S3 endpoints that don't support the additional
checksums feature. But they still want to use the upload review callback
to at least check that the parts sent matched the checksums they expect,
to detect corruption while parts were sitting in CRT memory before being sent.
This change makes it possible to set a checksum location of AWS_SCL_NONE
while still computing checksums for each upload part. The upload review
callback will be invoked as normal, including per-part checksums, but the
checksums won't be included in the requests sent to S3.

(My C is pretty rusty these days, so, uh, feedback welcome)


waahm7 (approved):
Thank you for creating the PR. I have made a small change to finalize error handling since now it can finalize in the read_callback. LGTM!

source/s3_checksum_stream.c
```diff
@@ -14,10 +14,29 @@ struct aws_checksum_stream {
     struct aws_input_stream *old_stream;
     struct aws_s3_checksum *checksum;
     struct aws_byte_buf checksum_result;
-    /* base64 encoded checksum of the stream, updated on destruction of stream */
+    /* base64 encoded checksum of the stream, updated at end of stream */
     struct aws_byte_buf *encoded_checksum_output;
+    bool checksum_finalized;
 };
 
+int s_aws_finalize_checksum(struct aws_checksum_stream *impl) {
```
TingDaoK, line 22: trivial: 
```suggestion
static int s_finalize_checksum(struct aws_checksum_stream *impl) {
```

source/s3_checksum_stream.c
```diff
@@ -14,10 +14,29 @@ struct aws_checksum_stream {
...
+    if (impl->checksum_finalized) {
+        return AWS_OP_SUCCESS;
+    }
+
+    int result = aws_checksum_finalize(impl->checksum, &impl->checksum_result, 0);
+    if (result != AWS_OP_SUCCESS) {
+        aws_byte_buf_reset(&impl->checksum_result, true);
+        impl->checksum_finalized = true;
+        return aws_raise_error(AWS_ERROR_S3_CHECKSUM_CALCULATION_FAILED);
```
TingDaoK, line 31: In general, I would expect all the function call, `aws_checksum_finalize`, will raise the proper error if the result is not AWS_OP_SUCCESS, so that the caller don't need to raise another error.

I looked into `aws_checksum_finalize`, it seems to raise lot of different errors and may not be very informative. Yeah, we should override the error code here, but I'd also suggest to log the previous error code for better error tracking.

source/s3_checksum_stream.c
```diff
@@ -14,10 +14,29 @@ struct aws_checksum_stream {
...
+    }
+    AWS_ASSERT(result == AWS_OP_SUCCESS);
```
TingDaoK, line 33: super trivial: no need to assert again right after the if check

TingDaoK (approved):
trivial stuff. 

fix&ship. (if you don't mind, I can fix it for you)

PR #429, opened 2024-05-08
Buffer-pool allows "forced" buffers, which don't count against memory limit

graebm:
**Issue:**
`aws_s3_meta_request_write()` must write to a buffer immediately if the data is less than part-size. Currently, it uses a buffer hooked up the to default allocator ([code here](https://github.com/awslabs/aws-c-s3/blob/cc06c419448b40417caa7b587f61bb4d8b4c08c1/source/s3_meta_request.c#L2260-L2262)). We'd like to get these buffers from the [buffer-pool](https://github.com/awslabs/aws-c-s3/pull/368), to reduce memory fragmentation and resident set size (RSS).

The problem is: the buffer-pool maintains strict memory limits, and won't allow further allocations when that limit is hit. But `aws_s3_meta_request_write()` MUST get a buffer immediately, or else the system can deadlock (see description in PR https://github.com/awslabs/aws-c-s3/pull/418)

**Description of changes:**
Add `aws_s3_buffer_pool_acquire_forced_buffer()`. These buffers can be created even if they exceed the memory limit.

**Future work:**
Modify `aws_s3_meta_request_write()` to use this new function

**Additional thoughts:**
- Anyone using `aws_s3_meta_request_write()` should limit the total number of uploads like: `max-uploads = memory-limit / part-size`. That was the case even before this PR.


DmitriyMusatkin:
also can you add a section on forced buffers to https://github.com/awslabs/aws-c-s3/blob/main/docs/memory_aware_request_execution.md. we probably want to keep it up to date

include/aws/s3/private/s3_buffer_pool.h
```diff
@@ -111,6 +121,17 @@ AWS_S3_API struct aws_byte_buf aws_s3_buffer_pool_acquire_buffer(
     struct aws_s3_buffer_pool *buffer_pool,
     struct aws_s3_buffer_pool_ticket *ticket);
 
+/*
+ * Force immediate acquisition of a buffer from the pool.
+ * This should only be used if waiting to reserve a ticket would risk deadlock.
+ * This cannot fail, not even if the pool has a reservation hold,
+ * not even if the memory limit has been exceeded.
+ */
+AWS_S3_API struct aws_byte_buf aws_s3_buffer_pool_acquire_forced_buffer(
```
DmitriyMusatkin, line 130: not having it be able to fail seems dangerous. we should probably plan for supporting some sort of bound on forced allocations (if not now than in future). we should have an option for client to enforce how much mem can be used by forced allocations

DmitriyMusatkin (approved):
nits. looks fine overall

DmitriyMusatkin (approved):

PR #430, opened 2024-05-09
Improve async write (fewer copies, polling API)

graebm:
**Issue:**
Mountpoint's upload throughput took a 20% hit when they started using the new async write API (PR https://github.com/awslabs/aws-c-s3/pull/418).

We knew the new API could, in the worst case, do an additional copy (see TODOs in [PR description](https://github.com/awslabs/aws-c-s3/pull/418) and [code](https://github.com/awslabs/aws-c-s3/blob/3647b4bfb622f88e938529b292afed12fa60e27f/source/s3_meta_request.c#L2260).). Some quick experimentation showed this was the cause.

**Description of changes:**
- Remove additional copy.
    - Async write gets buffers from the buffer-pool (made possible by PR https://github.com/awslabs/aws-c-s3/pull/429)
- Add new `poll_write()` function, which is simpler to use from Mountpoint.
    - Mountpoint typically does 1MiB or 256KiB writes. So we pretty much always need to copy the data immediately. So let's optimize for that.
    - Rust needed tricky code to cope with the original `write()` API's demand that data stay alive until the write-future completes. And aws-c-s3 needed tricky code to guarantee `cancel()` would synchronously fire any pending write-futures. If we just offer a rust-polling-style API that always copies synchronously, we can simplify a lot of code. So let's do that.


include/aws/s3/s3_client.h
```diff
@@ -834,6 +835,74 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
...
+ * https://docs.rs/futures/latest/futures/io/trait.AsyncWrite.html#tymethod.poll_write
+ * If you prefer completion-style async functions, and your data can outlive
+ * the callstack, use aws_s3_meta_request_write() instead.
+ *
+ * Check the returned `result` struct to see what happened:
+ * 1)   If `result.is_pending == true` then no work was done.
+ *      The waker callback will be invoked when you can call poll_write() again.
+ *      Do not call poll_write() again before the waker is invoked.
+ *
+ * 2)   Else if `result.error_code != 0` then poll_write() did not succeed
```
DmitriyMusatkin, line 863: are there any recoverable errors, or error means everything is on fire and the only way forward is to kill things?

include/aws/s3/s3_client.h
```diff
@@ -834,6 +835,74 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
...
+ *      and you should not call it again. The meta request is guaranteed to finish soon
+ *      (you don't need to worry about canceling the meta request yourself after a failed write).
+ *      A common error code is AWS_ERROR_S3_REQUEST_HAS_COMPLETED, indicating
+ *      the meta request completed for reasons unrelated to the poll_write() call
+ *      (e.g. CreateMultipartUpload received a 403 Forbidden response).
+ *      AWS_ERROR_INVALID_STATE usually indicates that you're calling poll_write()
+ *      incorrectly (e.g. not waiting for waker callback from previous poll_write() call).
+ *
+ * 3)   Else `result.bytes_processed` tells you how much data was processed.
```
DmitriyMusatkin, line 872: calls to write in this case can be done one after the other, without waiting for waker or any sort of backoff?

include/aws/s3/s3_client.h
```diff
@@ -834,6 +835,74 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
...
+ *                      to subsequent poll_write() calls.
+ *
+ * @param waker         Waker callback.
+ *                      If `result.is_pending == true`, then the waker will be called
+ *                      exactly once when it's a good time to call poll_write() again.
+ *                      If `result.is_pending == false`, the waker will never be called.
+ *
+ * @param user_data     Pointer to be passed to the waker callback.
+ *
+ * WARNING: This feature is experimental.
```
DmitriyMusatkin, line 896: side question: should we consider something like experimental annotation that warns people when they use it? openssl has a deprecated flag and users must opt in if they want to avoid warning from those functions

source/s3_auto_ranged_put.c
```diff
@@ -548,8 +548,16 @@ static bool s_s3_auto_ranged_put_update(
 
             if (should_create_next_part_request) {
 
-                struct aws_s3_buffer_pool_ticket *ticket =
-                    aws_s3_buffer_pool_reserve(meta_request->client->buffer_pool, meta_request->part_size);
+                struct aws_s3_buffer_pool_ticket *ticket = NULL;
+                if (meta_request->synced_data.async_write.ready_to_send) {
+                    /* Async-write already has a ticket, take ownership */
+                    AWS_ASSERT(meta_request->synced_data.async_write.buffered_data_ticket);
```
DmitriyMusatkin, line 554: should we make it a fatal? it would fail hard if we dont have a ticket, right?

include/aws/s3/s3_client.h
graebm: "result.is_pending" is the only "recoverable" error, which is why I broke it out into its own variable, instead of it making it a special error code.

Full text is:
```
> Else if `result.error_code != 0` then poll_write() did not succeed
> and you should not call it again. The meta request is guaranteed to finish soon
> (you don't need to worry about canceling the meta request yourself after a failed write).
> A common error code is AWS_ERROR_S3_REQUEST_HAS_COMPLETED, indicating
> the meta request completed for reasons unrelated to the poll_write() call
> (e.g. CreateMultipartUpload received a 403 Forbidden response).
> AWS_ERROR_INVALID_STATE usually indicates that you're calling poll_write()
> incorrectly (e.g. not waiting for waker callback from previous poll_write() call).
 ```

My personal philosophy on APIs where users need to call a sequence of functions correctly to complete the operation is: If anything goes wrong with a function in that sequence, the operation should cancel itself. We shouldn't require the user to inspect the result of every code, to determine whether or not they need to cancel the operation. It's a streaming operation, so random failure was always a possibility.

So users can write simple code like:
```
if meta_request.write() failed:
    # something went wrong, bail out of write loop, meta-request will finish soon
    return
```

Instead of:
```
if meta_request.write() failed:
    # something went wrong, be sure to cancel the meta-request
    # in case this is our fault, then bail out of write loop, meta-request will finish soon
    if error_code != AWS_ERROR_S3_REQUEST_HAS_COMPLETED:
        meta_request.cancel()

    return
```

File: include/aws/s3/s3_client.h (hunk 2 above)
Comment (graebm) [reply include/aws/s3/s3_client.h:872]: yes

File: include/aws/s3/s3_client.h (hunk 3 above)
Comment (graebm) [reply include/aws/s3/s3_client.h:896]: 🤷‍♀️

File: source/s3_auto_ranged_put.c (hunk 4 above)
Comment (graebm) [reply source/s3_auto_ranged_put.c:554]: sure. it's a cheap check

File: source/s3_meta_request.c [hunk 5]
```diff
@@ -2280,76 +2293,129 @@ struct aws_future_void *aws_s3_meta_request_write(
         aws_s3_client_schedule_process_work(meta_request->client);
     }
 
-    return write_future;
+    /* Assert only 1 result field is set (byte_processed could be 0 if data.len is zero) */
+    AWS_ASSERT(
+        (result.is_pending && !result.error_code && !result.bytes_processed) ^
```
DmitriyMusatkin, line 2298: really minor and no need to fix: i know this works, but mixing logical and bitwise operators feels so wrong here

DmitriyMusatkin (approved):

source/s3_meta_request.c
graebm: tweaked to hopefully be less confusing

PR #439, opened 2024-06-17
BREAKING CHANGE: operation_name must be set for DEFAULT meta-requests

graebm:
**Issue:**

For some operations, S3 can return an error response, despite having sent status-code 200 OK (see [knowledge center](https://repost.aws/knowledge-center/s3-resolve-200-internalerror))

To defend against this, aws-c-s3 checked all 200 OK responses to see if they were actually XML errors, [except](https://github.com/awslabs/aws-c-s3/blob/963949202740da924a4282e79aa43d45e5eaaf31/source/s3_meta_request.c#L1496-L1505) responses related to `AWS_S3_META_REQUEST_TYPE_GET_OBJECT`. We knew that operations with "streaming body" responses never did "200 Error", and what if someone did "GetObject" on an XML error message they'd uploaded to S3, LoL 😂 haha 😂 that would never happen 😂.

We heard from a customer that was using `AWS_S3_META_REQUEST_TYPE_DEFAULT` (instead of `AWS_S3_META_REQUEST_TYPE_GET_OBJECT`) to send a "GetObject" request. The object was logs, and the logs began with an XML error message. Since aws-c-s3 didn't realize this was a "GetObject" request, it scanned the body of the response and mistook it for a "200 error".

**Description of changes:**

Users **MUST** set `operation_name` for `AWS_S3_META_REQUEST_TYPE_DEFAULT`. Previously, this was an optional field, but now it is required. This lets us ensure that "GetObject" operations are never checked for "200 Error", even if they were sent via `AWS_S3_META_REQUEST_TYPE_DEFAULT`.


source/s3.c
```diff
@@ -75,6 +76,56 @@ static struct aws_log_subject_info_list s_s3_log_subject_list = {
...
+    DEFINE_REQUEST_TYPE_OPERATION_NAME_PAIR(AWS_S3_REQUEST_TYPE_ABORT_MULTIPART_UPLOAD, "AbortMultipartUpload"),
+    DEFINE_REQUEST_TYPE_OPERATION_NAME_PAIR(AWS_S3_REQUEST_TYPE_COMPLETE_MULTIPART_UPLOAD, "CompleteMultipartUpload"),
+    DEFINE_REQUEST_TYPE_OPERATION_NAME_PAIR(AWS_S3_REQUEST_TYPE_UPLOAD_PART_COPY, "UploadPartCopy"),
+    DEFINE_REQUEST_TYPE_OPERATION_NAME_PAIR(AWS_S3_REQUEST_TYPE_COPY_OBJECT, "CopyObject"),
+    DEFINE_REQUEST_TYPE_OPERATION_NAME_PAIR(AWS_S3_REQUEST_TYPE_PUT_OBJECT, "PutObject"),
+    DEFINE_REQUEST_TYPE_OPERATION_NAME_PAIR(AWS_S3_REQUEST_TYPE_CREATE_SESSION, "CreateSession"),
+};
+
+/* Hash-table for case-insensitive lookup from operation-name -> request-type.
+ * key is `struct aws_byte_cursor *`, pointing into array above.
```
DmitriyMusatkin, line 102: nit: this comment was really confusing on first read and its mostly due to our hash table implementation details. in a classical hash table sense key is still the operation name, and its impl detail thats its a ptr to a cursor backed by the table above. i though you were somehow relying on cursor ptrs as your key on initial read

DmitriyMusatkin (approved):

source/s3_default_meta_request.c
```diff
@@ -105,25 +106,24 @@ struct aws_s3_meta_request *aws_s3_meta_request_default_new(
...
+
+    if (request_type == AWS_S3_REQUEST_TYPE_UNKNOWN) {
+        /* use name to get type */
         meta_request_default->operation_name = aws_string_new_from_cursor(allocator, &options->operation_name);
+        meta_request_default->request_type = aws_s3_request_type_from_operation_name(options->operation_name);
+    } else {
+        /* use type to get name */
+        meta_request_default->request_type = request_type;
+        meta_request_default->operation_name =
+            aws_string_new_from_c_str(allocator, aws_s3_request_type_operation_name(request_type));
```
TingDaoK, line 118: worth to think about make some generic operation name as a static string, so that we don't make too many copy/paste of the same string over and over again? May not be a bottle neck, but seems like a waste for cases like a lot small files.

include/aws/s3/private/s3_default_meta_request.h
```diff
@@ -18,7 +18,7 @@ struct aws_s3_meta_request_default {
     /* Actual type for the single request (may be AWS_S3_REQUEST_TYPE_UNKNOWN) */
     enum aws_s3_request_type request_type;
 
-    /* S3 operation name for the single request (NULL if unknown) */
+    /* S3 operation name for the single request */
```
TingDaoK, line 21: should we warn people that if they don't set it right, they may still have the issue downloading the error xml?

Or some documentation about what will happen if we don't know the operation name.

graebm: I added very explicit details to the docs in s3_client.h

source/s3_default_meta_request.c
graebm: did it

TingDaoK (approved):

PR #440, opened 2024-06-18
Fix shutdown_callback or returning NULL contract for meta_request

waahm7:
*Issue #, if available:*
https://github.com/aws/aws-sdk-java-v2/issues/5277

*Description of changes:*
There are multiple places where a meta-request can fail after doing the aws_s3_meta_request_init_base like https://github.com/awslabs/aws-c-s3/blob/963949202740da924a4282e79aa43d45e5eaaf31/source/s3_auto_ranged_get.c#L98, https://github.com/awslabs/aws-c-s3/blob/963949202740da924a4282e79aa43d45e5eaaf31/source/s3_client.c#L994, https://github.com/awslabs/aws-c-s3/blob/963949202740da924a4282e79aa43d45e5eaaf31/source/s3_client.c#L1055. In those cases, we were triggering the shutdown callback as well as returning NULL, which is a crime and can lead to double-free errors. This PR moves the shutdown_callback assignment to be the last thing set on the meta_request so that we don’t trigger the shutdown_callback if we fail for any reason after creating the meta_request and before returning it to customers.


source/s3_client.c
```diff
@@ -1088,8 +1088,13 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
             aws_error_str(aws_last_error()));
 
         meta_request = aws_s3_meta_request_release(meta_request);
-    } else {
+    } else if (meta_request != NULL) {
```
graebm, line 1091: I don't think it's possible for meta_request to be NULL in this `else` block?

If it were possible, we wouldn't end up logging anything about whether the meta-request creation succeeded or failed

```suggestion
    } else {
```

source/s3_client.c
```diff
@@ -1088,8 +1088,13 @@ struct aws_s3_meta_request *aws_s3_client_make_meta_request(
...
         AWS_LOGF_INFO(AWS_LS_S3_CLIENT, "id=%p: Created meta request %p", (void *)client, (void *)meta_request);
+        /**
+         * shutdown_callback must be the last thing that gets set on the meta_request so that we don’t return NULL and
+         * trigger the shutdown_callback.
+         */
+        meta_request->shutdown_callback = options->shutdown_callback;
```
graebm, line 1097: do we similarly need to worry about the finish_callback?
did you already look into that?

graebm (commented):
looks good overall, but had questions

source/s3_client.c
waahm7: finish_callback is fine because we don't trigger that from meta_request destroy function.

graebm (approved):

PR #443, opened 2024-06-21
rewrite some checksum related code

TingDaoK:
*Issue #, if available:*

- Get into this rabbit hole from removing `body_user_callback_after_checksum` to support write into file directly

*Description of changes:*

- No functionality change, beside I modified the request response header directly instead of making a copy.
- Remove the `*_after_checksum` callbacks, just invoke the right function from the right place
- we don't need to make a copy of headers before we send it to the user, we are not using the response headers anymore afterwards
- reuse some code instead of copy/paste


include/aws/s3/private/s3_util.h
```diff
@@ -323,6 +323,15 @@ int aws_s3_crt_error_code_from_recoverable_server_error_code_string(struct aws_b
 AWS_S3_API
 void aws_s3_request_finish_up_metrics_synced(struct aws_s3_request *request, struct aws_s3_meta_request *meta_request);
 
+/* Check the response headers for checksum to verify, return a running checksum based on the algorithm found. If no
+ * checksum found from header, return null. */
+AWS_S3_API
+struct aws_s3_checksum *aws_s3_check_headers_for_checksum(
```
graebm, line 329: seems like deep in the callstack stuff like `aws_sha1_default_new()` can fail (but we're not checking it). So you might need better indication of "error" vs "no need to do work". And we also probably should be checking this stuff for failure.

DmitriyMusatkin (approved):

PR #445, opened 2024-07-10
Make options more const

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
make everything passed in by pointer in options const as we always copy the underlying data.
everything that is ref counted is kept as is, since refcounted functions dont play with const ptrs nicely

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
       - 'main'
 
 env:
-  BUILDER_VERSION: v0.9.59
+  BUILDER_VERSION: v0.9.61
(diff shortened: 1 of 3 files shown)


graebm (approved):
I'd advise checking this branch from aws-crt-cpp, and making sure it all works nicely, before merging

DmitriyMusatkin:
> I'd advise checking this branch from aws-crt-cpp, and making sure it all works nicely, before merging

crt cpp actually does not wrap s3 now. i checked cpp sdk and its fine with it. i will check other bindinds

SergeyRyabinin (approved):

PR #449, opened 2024-08-15
S3 receive filepath

TingDaoK:
*Issue #, if available:*

- Add an API for writing into file directly from CRT
- The write will still happen from the same thread, sequentially. 

*Description of changes:*


source/s3_meta_request.c
```diff
@@ -234,6 +236,14 @@ int aws_s3_meta_request_init_base(
 
     /* If the request's body is being passed in some other way, set that up.
      * (we checked earlier that the request body is not being passed multiple ways) */
+    if (options->receive_filepath.len > 0) {
```
waahm7, line 239: This is between the doc and the code. So now, the doc doesn't correlate with the code.

include/aws/s3/s3_client.h
```diff
@@ -632,6 +632,14 @@ struct aws_s3_meta_request_options {
      * Do not set the message's body-stream if the body is being passed by other means (see note above) */
     struct aws_http_message *message;
 
+    /**
+     * Optional.
+     * If set, the received data will be written into this file.
+     * the `body_callback` will still be invoked if set.
```
waahm7, line 638: Do we have a use case where we might need both? I think it would be simpler to return an error if both are set until we have a use case for this. You can upload an object and then call the getObject in the tests instead of relying on the body callback.

include/aws/s3/s3_client.h
```diff
@@ -632,6 +632,14 @@ struct aws_s3_meta_request_options {
      * Do not set the message's body-stream if the body is being passed by other means (see note above) */
     struct aws_http_message *message;
 
+    /**
+     * Optional.
+     * If set, the received data will be written into this file.
```
waahm7, line 637: Mention that we will overwrite the file if it already exists.

include/aws/s3/s3_client.h
TingDaoK: there is no specific reason to not invoke the body callback. It's up to the user to decide what to do with the response body. 
I think it's better to not assume what will be the use case for user.

waahm7: We can discuss it, I think it's better to keep it restricted if we want to optimize it in the future, such as out-of-order writes or something. This will restrict almost all future optimizations because we will still need to trigger the body callback.

include/aws/s3/s3_client.h
TingDaoK: good call, reminds me that we actually need to support Java SDK's file options. https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java#L179

Added the support

include/aws/s3/s3_client.h
TingDaoK: I think we can still do future opt in the future by add new options. But, since we always default to some body callback in the binding, it maybe a performance concern.

source/s3.c
```diff
@@ -49,6 +49,8 @@ static struct aws_error_info s_errors[] = {
     AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3EXPRESS_CREATE_SESSION_FAILED, "CreateSession call failed when signing with S3 Express."),
     AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_INTERNAL_PART_SIZE_MISMATCH_RETRYING_WITH_RANGE, "part_size mismatch, possibly due to wrong object_size_hint. Retrying with Range instead of partNumber."),
     AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_REQUEST_HAS_COMPLETED, "Request has already completed, action cannot be performed."),
+    AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_RECV_FILE_EXISTS, "The receive file already exist, cannot override as configuration required."),
+    AWS_DEFINE_ERROR_INFO_S3(AWS_ERROR_S3_RECV_FILE_NOT_EXISTS, "The receive file doesn't exist, cannot create as configuration required."),
```
graebm, line 53: extremely debatable / not sure: Should we re-use AWS_ERROR_FILE_OPEN_FAILURE here?
If so, should we use AWS_ERROR_FILE_OPEN_FAILURE for FILE_ALREADY_EXISTS as well? and add no new error codes?

source/s3_meta_request.c
```diff
@@ -232,6 +234,54 @@ int aws_s3_meta_request_init_base(
...
+                    break;
+                } else {
+                    meta_request->recv_file = aws_fopen(aws_string_c_str(meta_request->recv_filepath), "wb");
+                    break;
+                }
+            case AWS_RECV_FILE_CREATE_OR_APPEND:
+                meta_request->recv_file = aws_fopen(aws_string_c_str(meta_request->recv_filepath), "ab");
+                break;
+            case AWS_RECV_FILE_WRITE_TO_POSITION:
+                if (!aws_path_exists(meta_request->recv_filepath)) {
```
graebm, line 261: may not need this check. you're calling `fopen()` with `r+` below, and docs say that gives an error if the file doesn't exist https://en.cppreference.com/w/c/io/fopen

source/s3.c
TingDaoK: hmmm, it will be harder for sdks to parse the error code and convert it to the expected exception from their side, right?

I think they will need to convert the exited error to their exception based on https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java#L181.

source/s3_meta_request.c
TingDaoK: Similar to the previous one, without the extra check, we will raise `AWS_ERROR_FILE_INVALID_PATH`, which cannot tell the issue is really from the configuration.

include/aws/s3/s3_client.h
```diff
@@ -632,6 +654,23 @@ struct aws_s3_meta_request_options {
...
+     * the `body_callback` will NOT be invoked.
+     * This gives a better performance when receiving data to write to a file.
+     */
+    struct aws_byte_cursor receive_filepath;
```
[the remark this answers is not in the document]
graebm: oh good point, forgot about that 👍

source/s3_meta_request.c
```diff
@@ -1781,7 +1843,21 @@ static void s_s3_meta_request_event_delivery_task(struct aws_task *task, void *a
                     if (meta_request->meta_request_level_running_response_sum) {
                         aws_checksum_update(meta_request->meta_request_level_running_response_sum, &response_body);
                     }
-                    if (meta_request->body_callback != NULL &&
+                    if (meta_request->recv_file) {
+                        /* Write the data directly to the file. No need to seek, since the event will always be
+                         * delivered with the right order. */
+                        if (fwrite((void *)response_body.ptr, response_body.len, 1, meta_request->recv_file) < 1) {
+                            int errno_value = ferror(meta_request->recv_file) ? errno : 0; /* Always cache errno  */
+                            error_code = aws_translate_and_raise_io_error_or(errno_value, AWS_ERROR_FILE_WRITE_FAILURE);
```
graebm, line 1851: aws_translate_and_raise_io_error_or() is like aws_raise_error(): it always returns AWS_OP_ERR(-1)
```suggestion
                            aws_translate_and_raise_io_error_or(errno_value, AWS_ERROR_FILE_WRITE_FAILURE);
                            error_code = aws_last_error()
```

source/s3_meta_request.c
```diff
@@ -1781,7 +1843,21 @@ static void s_s3_meta_request_event_delivery_task(struct aws_task *task, void *a
...
+                            AWS_LOGF_ERROR(
+                                AWS_LS_S3_META_REQUEST,
+                                "id=%p Failed writing to file. errno:%d. aws-error:%s",
+                                (void *)meta_request,
+                                errno_value,
+                                aws_error_name(aws_last_error()));
```
graebm, line 1857:
```suggestion
                                aws_error_name(error_code));
```

source/s3_meta_request.c
```diff
@@ -1781,7 +1843,21 @@ static void s_s3_meta_request_event_delivery_task(struct aws_task *task, void *a
                     if (meta_request->meta_request_level_running_response_sum) {
                         aws_checksum_update(meta_request->meta_request_level_running_response_sum, &response_body);
```
graebm, line 1844: I know this isn't part of your PR. But I see 😬 we're not checking aws_checksum_update() for error

graebm (approved):
Fix & ship

include/aws/s3/s3_client.h
```diff
@@ -632,6 +654,34 @@ struct aws_s3_meta_request_options {
...
+     * Optional.
+     * The specified position to start writing at for the recv file when `recv_file_option` is set to
+     * AWS_S3_RECV_FILE_WRITE_TO_POSITION, ignored otherwise.
+     */
+    uint64_t recv_file_position;
+    /**
+     * Set it to be true to delete the receive file on failure, otherwise, the file will be left as-is.
+     * This only works with recv_filepath set.
+     */
+    bool recv_file_delete_on_failure;
```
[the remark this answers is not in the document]
graebm: in [this stackoverflow post](https://stackoverflow.com/questions/4845122/using-a-numbered-file-descriptor-from-java) one comment pointed to a [code sample](https://www.kfu.com/~nsayer/Java/jni-filedesc.html) that shows pulling the `int fd` value using JNI, because JNI ignores private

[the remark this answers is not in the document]
TingDaoK: If we want to do something fancy like O_DIRECT to improve the performance, it will need to be us to control the behavior from our side.

PR #453, opened 2024-09-18
Fix MemoryLimit primary_allocated stat

waahm7:
*Description of changes:*
- We never decrement the primary_used stat, which leads to incorrect reporting of an ever-increasing number.

### DIFF
--- a/source/s3_buffer_pool.c
+++ b/source/s3_buffer_pool.c
@@ -232,6 +232,7 @@ void s_buffer_pool_trim_synced(struct aws_s3_buffer_pool *buffer_pool) {
         aws_array_list_get_at_ptr(&buffer_pool->blocks, (void **)&block, i);
 
         if (block->alloc_bit_mask == 0) {
+            buffer_pool->primary_allocated -= block->block_size;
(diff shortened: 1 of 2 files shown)


source/s3_buffer_pool.c
```diff
@@ -232,6 +232,7 @@ void s_buffer_pool_trim_synced(struct aws_s3_buffer_pool *buffer_pool) {
         aws_array_list_get_at_ptr(&buffer_pool->blocks, (void **)&block, i);
 
         if (block->alloc_bit_mask == 0) {
+            buffer_pool->primary_allocated -= block->block_size;
```
DmitriyMusatkin, line 235: can you check if we have same bug in secondary used?

DmitriyMusatkin (approved):

source/s3_buffer_pool.c
waahm7: I have checked and I think we don't have it https://github.com/awslabs/aws-c-s3/blob/main/source/s3_buffer_pool.c#L487.

PR #454, opened 2024-10-07
Support header checksum

TingDaoK:
*Issue #, if available:*
- Adding support for providing checksum in the header instead of trailer.
- CRT will read the body into memory and calculate the checksum directly from the memory in this case, and add it to the header of the request instead of trailer

*Description of changes:*


source/s3_request_messages.c
```diff
@@ -730,6 +730,91 @@ struct aws_http_message *aws_s3_abort_multipart_upload_message_new(
...
+    ret_code = AWS_OP_SUCCESS;
+done:
+    if (out_checksum == NULL) {
+        /* In case out_checksum is not set. In this case, we need to clean up the local_encoded_checksum.
+         * Otherwise, the caller will own the out_checksum.
+         */
+        aws_byte_buf_clean_up(local_encoded_checksum);
+        aws_mem_release(allocator, local_encoded_checksum);
+    }
+    return ret_code;
```
graebm, line 815: if there's an error, we may need to cleanup out_checksum

graebm (approved):

PR #455, opened 2024-10-09
supports copy object within directory bucket

TingDaoK:
*Issue #, if available:*

- The current copy object doesn't support s3 express, because for CopyObject and UploadPartCopy, the request has to be signed with regular SigV4 for directory buckets.

*Description of changes:*
- Override to force regular signing when sign the request for CopyObject and UploadPartCopy
- Added tests for copy within the same directory bucket. 


include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -350,6 +350,14 @@ void aws_s3_meta_request_init_signing_date_time_default(
     struct aws_s3_meta_request *meta_request,
     struct aws_date_time *date_time);
 
+AWS_S3_API
+void aws_s3_meta_request_sign_request_default_impl(
+    struct aws_s3_meta_request *meta_request,
+    struct aws_s3_request *request,
+    aws_signing_complete_fn *on_signing_complete,
+    void *user_data,
+    bool force_regular_signing);
```
waahm7, line 359: `regular_signing` is confusing here, since we have three different types of signing algorithms and possibly more in the future. Might be simpler to just move this check to `s_meta_request_resolve_signing_config` and set the appropriate algorithm?

source/s3_copy_object.c
```diff
@@ -796,3 +802,20 @@ static void s_s3_copy_object_request_finished(
...
+void s_s3_copy_object_sign_request(
+    struct aws_s3_meta_request *meta_request,
+    struct aws_s3_request *request,
+    aws_signing_complete_fn *on_signing_complete,
+    void *user_data) {
+
+    /**
+     * https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html
+     * https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html
+     * For CopyObject and UploadPartCopy, the request has to be signed with regular SigV4 for directory buckets.
```
waahm7, line 815:
```suggestion
     * For CopyObject and UploadPartCopy, the request has to be signed with SigV4 for directory buckets.
```

include/aws/s3/private/s3_meta_request_impl.h
TingDaoK: I thought of adding the checking to `s3_meta_request.c`, but I think we want to keep the request type information limits to the separate type of meta request files. So, I'd like to keep the checking for different operations within `s3_copy_object.c`. I think the bool is enough for our current use case, if we ever want to support more complicated way to resolve the signing algorithms, we can think of something else later (Maybe another v function to resolve the signing configs?) And they are all private, which gives us flexibility to change the API in the future.

waahm7: We can rename `force_regular_signing` to something better. The Endpoints ruleset has the `disable_s3_express_session_auth` parameter; we can have something similar here like disable_s3_express_signing.

include/aws/s3/s3_client.h
```diff
@@ -82,6 +82,7 @@ enum aws_s3_meta_request_type {
      * - only {bucket}/{key} format is supported for source and passing arn as
      *   source will not work
      * - source bucket is assumed to be in the same region as dest
+     * - source bucket and dest needs both be directory bucket or regular bucket.
```
waahm7, line 85:
```suggestion
     * - source bucket and the dest bucket must both be either directory buckets or regular buckets.
```

waahm7 (approved):

PR #456, opened 2024-10-15
Validate Invalid Network Interface Names at Client Initialization 

waahm7:
*Description of changes:*
- If there are any typos in the network_interface_names, we can catch them at client initialization time instead of when we try to make a request and it fails due to timeout.


source/s3_client.c
```diff
@@ -352,6 +356,7 @@ struct aws_s3_client *aws_s3_client_new(
     struct aws_s3_buffer_pool_usage_stats pool_usage = aws_s3_buffer_pool_get_usage(client->buffer_pool);
 
     if (client_config->max_part_size > pool_usage.mem_limit) {
+        aws_s3_buffer_pool_destroy(client->buffer_pool);
```
graebm, line 359: trying to clean up behind yourself is hard in C
which is why we prefer the "goto error, cleanup everything" pattern,
and why our cleanup/destroy functions are the only ones that tolerate unexpected NULL args

This function already has 2 different "goto error" labels.
Now 1 of those labels you have to be careful to also cleanup a specific variable before you goto it. this is too fragile 😓

Seems better to completely remove the `on_early_fail` label from this function, and just always try to cleanup everything via `goto on_error`?

Another option: instead of the `on_error:` label sharing 90% the same cleanup code as `s_s3_client_finish_destroy_default()`, have them share code, and tolerate destroying a half-assembled client

waahm7: Thanks, I have moved code around to get rid of `on_early_fail`.

source/s3_client.c
```diff
@@ -343,31 +343,10 @@ struct aws_s3_client *aws_s3_client_new(
         }
     }
 
-    client->buffer_pool = aws_s3_buffer_pool_new(allocator, part_size, mem_limit);
-
-    if (client->buffer_pool == NULL) {
-        goto on_early_fail;
-    }
```
graebm, line 350: do we still need to move any of this code around? Can we keep it where it was, and change ~goto on_early_fail~ -> `goto on_error` and call it a day?

waahm7: Yeah, we need to make sure we acquire the bootstrap reference before `on_error`, since `on_error` releases the reference.

waahm7: Thanks, updated with a test.

graebm (approved):

PR #459, opened 2024-11-08
Support trailing checksum with no signing

TingDaoK:
*Issue #, if available:*

- When you do trialing checksum, apparently, you need to have the x-amz-content-sha256 header to be set to the special value even if you are not signing the request.

*Description of changes:*


tests/s3_mock_server_tests.c
```diff
@@ -167,6 +167,206 @@ TEST_CASE(multipart_upload_mock_server) {
...
+        .tls_mode = AWS_MR_TLS_DISABLED,
+    };
+
+    ASSERT_SUCCESS(aws_s3_tester_bind_client(&tester, &client_config, AWS_S3_TESTER_BIND_CLIENT_REGION));
+
+    struct aws_s3_client_vtable s3_client_get_requests_header_vtable = g_s3_client_default_vtable;
+    s3_client_get_requests_header_vtable.http_connection_make_request = s_get_requests_header_make_request;
+    struct aws_s3_client *client = aws_s3_client_new(allocator, &client_config);
+    ASSERT_NOT_NULL(client);
+    client->vtable = &s3_client_get_requests_header_vtable;
```
graebm, line 231: An alternative to making the vtable global:

```suggestion
    struct aws_s3_client *client = aws_s3_client_new(allocator, &client_config);
    ASSERT_NOT_NULL(client);
    
    // Patch the vtable
    struct aws_s3_client_vtable s3_client_get_requests_header_vtable = *client->vtable;
    s3_client_get_requests_header_vtable.http_connection_make_request = s_get_requests_header_make_request;
    client->vtable = &s3_client_get_requests_header_vtable;
```

include/aws/s3/private/s3_client_impl.h
```diff
@@ -524,6 +528,9 @@ void aws_s3_client_update_upload_part_timeout(
     struct aws_s3_request *finished_upload_part_request,
     int finished_error_code);
 
+AWS_S3_API
+extern struct aws_s3_client_vtable g_s3_client_default_vtable;
```
graebm, line 532: trivial: seems a tiny bit sketchy, making this a public symbol

source/s3_meta_request.c
```diff
@@ -1064,6 +1058,17 @@ static void s_s3_meta_request_request_on_signed(
 
 finish:
 
+    /**
+     * Add the x-amz-content-sha256 header to support trailing checksum.
```
graebm, line 1062:
```suggestion
     * Add "x-amz-content-sha256: STREAMING-UNSIGNED-PAYLOAD-TRAILER" header to support trailing checksum.
```

source/s3_meta_request.c
```diff
@@ -1064,6 +1058,17 @@ static void s_s3_meta_request_request_on_signed(
...
+     */
+    if (error_code == AWS_ERROR_SUCCESS && request->send_data.require_streaming_unsigned_payload_header) {
```
graebm, line 1064: is it possible to move this earlier in the function. It's an odd flow, having it after the `finish` goto label, and with the `if (error_code == AWS_ERROR_SUCCESS`

source/s3_meta_request.c
```diff
@@ -936,17 +941,6 @@ void aws_s3_meta_request_sign_request_default_impl(
         (void *)request,
         (void *)request->send_data.message);
 
-    if (request->send_data.signable == NULL) {
```
graebm, line 939: nice job shuffling around these log statements and if-checks
makes more sense now
👍

source/s3_meta_request.c
```diff
@@ -1062,6 +1056,17 @@ static void s_s3_meta_request_request_on_signed(
             metric->time_metrics.sign_end_timestamp_ns - metric->time_metrics.sign_start_timestamp_ns;
     }
 
+    /**
+     * Add "x-amz-content-sha256: STREAMING-UNSIGNED-PAYLOAD-TRAILER" header to support trailing checksum.
+     */
+    if (request->send_data.require_streaming_unsigned_payload_header) {
+        struct aws_http_headers *headers = aws_http_message_get_headers(request->send_data.message);
+        error_code = aws_http_headers_set(
```
graebm, line 1064: 3 things:
1) bug: if there's an error here, `error_code` will be -1, but it's *supposed* to be the actual error code
2) style: please add an explicit `goto finish` after setting error_code. I know it's not technically necessar because this is currently the last thing that can go wrong, but if some future person inserts more logic after this if-branch, they might miss that this if-branch would need a goto
3) utterly trivial: move this so it happens before `signing_duration_ns` gets set, since that seems like the last thing that should happen

graebm (approved):
fix & ship

PR #467, opened 2024-12-02
meta request: support for should_continue cancellation callback

grrtrr:
This restores an updated version of #238.

### Rationale
Add a boolean `continue_callback`, which polls a request-cancellation flag set by the user. This matches the use-case of the `ContinueRequestHandler` / `ContinueRequest` / `ShouldContinue` boolean functions used by `aws-sdk-cpp`.

Passive polling is preferred over active cancellation, since the latter would require to store a reference to the s3 meta request. The polling function is evaluated at the following points during the lifetime of the meta request:
1. _Request preparation_ (`aws_s3_meta_request_prepare_request`).
2. _Request update_ (the `.update` handler invoked via `aws_s3_meta_request_update`).
3. _Http stream completion_ (`.on_complete` handler invoking `aws_s3_meta_request_send_request_finish_default`).
4. _Writing download data_ (`aws_s3_meta_request_stream_response_body_synced`).

If a `continue_callback` is provided and if it returns `false` at any of these 4 evaluation times, the meta request completes with the `error_code` `AWS_ERROR_S3_CANCELED`.

By using a boolean flag in the body of the function, the polling overhead can be kept minimal.


grrtrr:
@jmklix @TingDaoK - this is an updated version of #238, which takes into account Dengke's changes in #395 and #399.

We need the `should_continue` callback function in order to implement the AWS C++ SDK `Cancel()` function for the `S3CrtClient`.

PR #482, opened 2025-01-09
Improve Copy Operation by taking the Source URI

waahm7:
*Issue #, if available:*
https://github.com/awslabs/s3-connector-for-pytorch/issues/295

*Description of changes:*
The current copy implementation has a lot of limitations because we try to parse the source host and path from the `x-amz-copy-source` header and the destination URI. This falls apart in many cases, such as when the bucket has a `.` in its name or during cross-region copies where we can't generate the source URL correctly due to missing information. Allowing the user to pass the source URL can bypass these limitations.

- Code coverage action was broken, fixed by updating gcc to match newer ubuntu.


source/s3_copy_object.c
```diff
@@ -92,10 +91,22 @@ struct aws_s3_meta_request *aws_s3_meta_request_copy_object_new(
     copy_object->synced_data.content_length = UNKNOWN_CONTENT_LENGTH;
     copy_object->synced_data.total_num_parts = UNKNOWN_NUM_PARTS;
     copy_object->threaded_update_data.next_part_number = 1;
+    if (options->copy_source_uri.len != 0) {
+        if (aws_uri_init_parse(&copy_object->source_uri, allocator, &options->copy_source_uri)) {
+            AWS_LOGF_ERROR(
+                AWS_LS_S3_META_REQUEST,
+                "Unable to parse the copy_source_uri provided in the request: " PRInSTR "",
+                AWS_BYTE_CURSOR_PRI(options->copy_source_uri));
+            goto on_error;
```
TingDaoK, line 100: the `copy_object->synced_data.part_list` was initialized above, it needs to be cleaned up here.

Or move the initialization of the list after anything can fail

We could add a simple test to pass in an invalid URI to add a bit more test coverage.

source/s3_request_messages.c
```diff
@@ -457,8 +458,37 @@ static const struct aws_byte_cursor s_slash_char = AWS_BYTE_CUR_INIT_FROM_STRING
  */
 struct aws_http_message *aws_s3_get_source_object_size_message_new(
     struct aws_allocator *allocator,
-    struct aws_http_message *base_message) {
+    struct aws_http_message *base_message,
+    struct aws_uri *source_uri) {
     struct aws_http_message *message = NULL;
+
+    message = aws_http_message_new_request(allocator);
```
TingDaoK, line 465: Lot of the error handling down there assumes there was nothing to clean up and just returns null.

Need to update them since we now allocate the message at the beginning of the function.

source/s3_request_messages.c
```diff
@@ -457,8 +458,37 @@ static const struct aws_byte_cursor s_slash_char = AWS_BYTE_CUR_INIT_FROM_STRING
...
+    if (message == NULL) {
+        goto error_cleanup;
```
TingDaoK, line 467: `head_object_host_header` have not been defined yet, but we will try to clean it up from this goto.

source/s3_copy_object.c
waahm7: Thanks, good catch. Fixed with a test.

source/s3_request_messages.c
waahm7: Thanks, fixed.

source/s3_request_messages.c
```diff
@@ -457,30 +458,63 @@ static const struct aws_byte_cursor s_slash_char = AWS_BYTE_CUR_INIT_FROM_STRING
...
-    struct aws_http_message *base_message) {
-    struct aws_http_message *message = NULL;
+    struct aws_http_message *base_message,
+    struct aws_uri *source_uri) {
+
+    struct aws_http_message *message = aws_http_message_new_request(allocator);
     struct aws_byte_buf head_object_host_header;
     AWS_ZERO_STRUCT(head_object_host_header);
 
+    if (message == NULL) {
```
DmitriyMusatkin, line 468: do we need to worry about the signing region here today as well? or we are keeping the source must be in the same region as dest behavior?

TingDaoK (approved):

DmitriyMusatkin (approved):

PR #492, opened 2025-01-23
S3Express CreateSession Allowlist Headers

waahm7:
*Issue #, if available:*
Adding support for SSE and ReadOnly sessions for Mountpoint.

*Description of changes:*
- As of now, the S3Express [CreateSession](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateSession.html#API_CreateSession_RequestSyntax) API allows the users to set 5 headers. This PR adds an experimental support for these headers so that if the request contains these headers, we will pass them to the CreateSession call. 
- Updates the caching layer to include all allowed headers and we will create a new session for any difference in headers. 
- This API uses an allow list instead of the usual exclude list for other APIs. Mainly because we want tighter control over what we are passing to the CreateSession call and what we are caching.


dannycjones:
@waahm7 we'll also want to set the [S3 Express session mode](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateSession.html#API_CreateSession_RequestSyntax), will you account for that in this PR?

waahm7:
@dannycjones Thanks, I have accounted for it in this PR.

source/s3_request_messages.c
```diff
@@ -27,6 +27,7 @@ const struct aws_byte_cursor g_s3_create_multipart_upload_excluded_headers[] = {
     AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-checksum-sha1"),
     AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-checksum-sha256"),
     AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("if-none-match"),
+    AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-create-session-mode"),
```
DmitriyMusatkin, line 30: sse headers passed to create, upload and complete? seems kinda counter intuitive

source/s3express_credentials_provider.c
```diff
@@ -388,16 +396,22 @@ static void s_on_request_finished(
...
 
     struct aws_http_header host_header = {
         .name = g_host_header_name,
         .value = host_value,
     };
-
-    if (aws_http_message_add_header(request, host_header)) {
-        goto error;
+    /* NOTE: Only for Tests.
+     * Don't add the host header for endpoint override.
```
DmitriyMusatkin, line 409: why? sounds like even for overridden endpoint we still need host

source/s3_request_messages.c
waahm7: Yes, sse headers are same as in CreateMPU and Upload Part. From [CreateMultipart](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html) Docs:
"""
In the Zonal endpoint API calls (except [CopyObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html) and [UploadPartCopy](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html)) using the REST API, the encryption request headers must match the encryption settings that are specified in the CreateSession request. You can't override the values of the encryption settings (x-amz-server-side-encryption, x-amz-server-side-encryption-aws-kms-key-id, x-amz-server-side-encryption-context, and x-amz-server-side-encryption-bucket-key-enabled) that are specified in the CreateSession request. You don't need to explicitly specify these encryption settings values in Zonal endpoint API calls, and Amazon S3 will use the encryption settings values from the CreateSession request to protect new objects in the directory bucket.
"""

waahm7: They are always excluded in the completeMPU.

source/s3express_credentials_provider.c
waahm7: The previous logic was passing the endpoint override to meta_request options, so we didn't need to add the host header. I have updated it to just add the host header from override as that makes it more understandable.

waahm7: Found a bug in our host parsing logic: https://github.com/awslabs/aws-c-s3/pull/492/files#diff-5ee13dd0ee005b828bde6d60ff89582070983e3cb7e5d92389b629a9850c27f8R1033

DmitriyMusatkin (approved):

PR #504, opened 2025-04-01
add request ID to debug level log

TingDaoK:
*Issue #, if available:*

- AWS CLI don't expose the option to set the trace level log from us. https://github.com/aws/aws-cli/blob/v2/awscli/logger.py#L80
- Our debug level log doesn't have request ID which is very help to debug the issue.
- Add the request id to our debug level log

*Description of changes:*
- Well, we need to string compare for every response headers now, I can run some benchmark, but likely it's not the bottle neck of anything

### DIFF
--- a/source/s3_meta_request.c
+++ b/source/s3_meta_request.c
@@ -1629,7 +1629,8 @@ void aws_s3_meta_request_send_request_finish_default(
 
     AWS_LOGF_DEBUG(
         AWS_LS_S3_META_REQUEST,
-        "id=%p: Request %p finished with error code %d (%s) and response status %d, x-amz-request-id: %s, x-amz-id-2: %s",
+        "id=%p: Request %p finished with error code %d (%s) and response status %d, x-amz-request-id: %s, x-amz-id-2: "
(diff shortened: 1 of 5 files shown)


waahm7 (approved):

source/s3_meta_request.c
```diff
@@ -1624,12 +1629,14 @@ void aws_s3_meta_request_send_request_finish_default(
 
     AWS_LOGF_DEBUG(
         AWS_LS_S3_META_REQUEST,
-        "id=%p: Request %p finished with error code %d (%s) and response status %d",
+        "id=%p: Request %p finished with error code %d (%s) and response status %d, request id: %s, amz_id_2: %s",
```
graebm, line 1632: trivial: Refer to both of these by either: their full literal header name, or their modeled names ... actually apparently neither of these are in the service model

```suggestion
        "id=%p: Request %p finished with error code %d (%s) and response status %d, x-amz-request-id: %s, x-amz-id-2: %s",
```

graebm (approved):

PR #506, opened 2025-04-04
async pause for get

TingDaoK:
*Issue #, if available:*

- Our current pause doesn't support GetObject
- The pause is currently just grabbing the status while background threads still running
- When CRT control the write to the file, it means the file write can happen after the pause is called (race between those two threads.)
- When downstream wants to gather the file last modified time, it makes the API hard to use. User needs to call the pause, and then wait for the `finish` callback to gather more data from user's side.
- Add an async API for pause to notify caller when the write/events completed.

*Description of changes:*
- Schedule an event for pause, when the event invokes, the other events should have finished the callbacks.
- _If people invoke the pause and wait for it to complete from one of the callbacks, they may end up with dead lock..._ 
  - warnings??


source/s3_auto_ranged_get.c
```diff
@@ -42,6 +45,8 @@ static struct aws_s3_meta_request_vtable s_s3_auto_ranged_get_vtable = {
     .finished_request = s_s3_auto_ranged_get_request_finished,
     .destroy = s_s3_meta_request_auto_ranged_get_destroy,
     .finish = aws_s3_meta_request_finish_default,
+    .pause = s_s3_auto_ranged_get_pause,
+    .pause_async = aws_s3_meta_request_pause_async_default,
```
waahm7, line 49: Having both pause and pause_async seems bad. I think complexity is stemming from trying to not break the existing interface. If Java_SDK is the only user of this interface, then it is probably fine to break this interface internally. If we do break it, can we simplify it like pause is always async or have different types of events callbacks (see third paragraph)?

I also think that for now, the workaround on the Java SDK side to just wait for the on_finish callback seems reasonable and might be less complicated if we want to recommend that they should wait for the on_finish callback in all cases, because in the future we might have code that does something between pause callback and finish callback? 

Another option is instead of having a separate callback, just make it another type of progress event. If you pause the meta_request, you will get a final event callback to let you know it was paused, but this doesn't fit into our current design where the single purpose of the progress callback is to just let you know how many bytes are completed.

I'm interested to see what other people think.

Issue #514, opened 2025-04-25
Requests SHOULD NOT contain Content-Length: 0 for GET/HEAD/DELETE requests


joeshaw:
### Describe the bug

(Originally filed as https://github.com/awslabs/mountpoint-s3/issues/1381)

I work on an S3-compatible product and we've seen compatibility issues with s3-mountpoint as a result of it sending `Content-Length: 0` and including it as a signed header on HTTP requests that don't require it.  Our server platform strips `Content-Length` for requests using an HTTP method where a request body is not expected, and subsequently our signature verification code fails because it isn't present when we calculate the signature.  We've worked around it but this is a common source of HTTP client/server compatibility issues generally and I thought you might be interested in changing it for `GET`, `HEAD`, and `DELETE` requests.

The current HTTP RFC is [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110).  In [Section 8.6](https://www.rfc-editor.org/rfc/rfc9110#name-content-length), it says:

> A user agent SHOULD send Content-Length in a request when the method defines a meaning for enclosed content and it is not sending Transfer-Encoding. For example, a user agent normally sends Content-Length in a POST request even when the value is 0 (indicating empty content). **A user agent SHOULD NOT send a Content-Length header field when the request message does not contain content and the method semantics do not anticipate such data.**

(emphasis mine)

Under [Section 9.3](https://www.rfc-editor.org/rfc/rfc9110#name-method-definitions) where it goes over the individual methods, for `GET`, `HEAD` and `DELETE` it says:

> A client SHOULD NOT generate content in a GET [or HEAD or DELETE] request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

No `Content-Length: 0` header sent for GET/HEAD/DELETE requests.

### Current Behavior

`Content-Length` is sent on all requests, including `Content-Length: 0` for GET/HEAD/DELETE.

### Reproduction Steps

This was bought to us as an incompatibility between [mountpoint-s3](https://github.com/awslabs/mountpoint-s3) and [Fastly Object Storage](https://www.fastly.com/products/storage) from one of our customers.

We've worked around the issue on our end, so it no longer produces an error and can't be easily reproduced, but before that was deployed you would get a signature mismatch error.

[...]

PR #516, opened 2025-05-02
default request without body

TingDaoK:
*Issue #, if available:*

- https://github.com/awslabs/aws-c-s3/issues/514
- If there is no body for the default type request, don't assign a body
- But, in case of `AWS_S3_REQUEST_TYPE_PUT_OBJECT` and `AWS_S3_REQUEST_TYPE_UPLOAD_PART`, to support trailing checksum, even if the object has empty body, still need an aws-chunked encoding and content-length


tests/s3_data_plane_tests.c
```diff
@@ -4000,18 +4002,22 @@ static int s_test_s3_round_trip_default_get_fc_helper(
 
 AWS_TEST_CASE(test_s3_round_trip_default_get_fc, s_test_s3_round_trip_default_get_fc)
 static int s_test_s3_round_trip_default_get_fc(struct aws_allocator *allocator, void *ctx) {
-    return s_test_s3_round_trip_default_get_fc_helper(allocator, ctx, false, AWS_TEST_FOC_NONE);
+    return s_test_s3_round_trip_default_get_fc_helper(allocator, ctx, false, 1 /*object_size_mb*/, AWS_TEST_FOC_NONE);
+}
+AWS_TEST_CASE(test_s3_round_trip_empty_fc, s_test_s3_round_trip_empty_fc)
+static int s_test_s3_round_trip_empty_fc(struct aws_allocator *allocator, void *ctx) {
+    return s_test_s3_round_trip_default_get_fc_helper(allocator, ctx, false, 0, AWS_TEST_FOC_NONE);
```
waahm7, line 4009: I don't fully get the changes the tests. Where are we asserting that we are not adding any content-length with GetObject?

PR #521, opened 2025-06-06
Fix race condition between mem acquire and cancel

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
canceling meta request had a race condition with memory buffer acquisition. cancelling at the right time, might have resulted in seg fault due to some bookkeeping info gettting incorrect.

this pr covers several changes:
- cancel no longer removes bookkeeping nodes for pending buffer futures (and relies on future callback to do that)
- future callback is now registered on a separate thread to avoid any deadlocks between cancel and mem acquisition


source/s3_client.c
```diff
@@ -1909,7 +1909,10 @@ void s_acquire_mem_and_prepare_request(
         }
         /* END CRITICAL SECTION */
 
-        aws_future_s3_buffer_ticket_register_callback(payload->buffer_future, s_on_pool_buffer_reserved, payload);
+        /* Note: run callback async on event loop. this is done to avoid any race conditions between cancelling
+           which requires a lock on meta request and buffer acquire callback which also requires a lock. */
+        aws_future_s3_buffer_ticket_register_event_loop_callback(
```
TingDaoK, line 1914: looks like when the future fails, `s_on_pool_buffer_reserved` will not remove the node from the list.

DmitriyMusatkin: fixed and added test

include/aws/s3/private/s3_request.h
```diff
@@ -139,8 +139,14 @@ struct aws_s3_request {
...
     struct aws_byte_buf request_body;
 
+    /**
+     * Ticket to acquire the buffer.
+     */
     struct aws_s3_buffer_ticket *ticket;
 
+    /* Result of memory acquisition */
+    int buffer_acquire_result;
+
```
TingDaoK, line 149: trivial: never used?

TingDaoK (approved):
Fix&ship

PR #531, opened 2025-06-26
don't regenerate the checksum

TingDaoK:
*Issue #, if available:*
- We previously clean up the checksum after prepare.
- It results in the checksum to be recalculated during retry
- It's violating the durability guarantee 
- However, since we already buffer the input in memory during retry, it won't read the customer input again, which mitigated the impact.

*Description of changes:*
- Revamp the code a bit to allow the passed in checksum to the input stream wrappers. So that it won't recalculate the checksum

### TODO:
- The whole checksum logic is already complicated this makes it more weird, need to revamp it...
- I'd prefer to push the revamp of the checksum logic later when we implement the full object checksum. 


TingDaoK:
Included in https://github.com/awslabs/aws-c-s3/pull/532

PR #532, opened 2025-07-02
Revamp checksum

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/s3_request_messages.c
```diff
@@ -352,17 +353,18 @@ struct aws_http_message *aws_s3_create_multipart_upload_message_new(
...
+ * Create a new put object request from an existing put object request.  Currently just optionally adds part information
+ * for a multipart upload.
+ **/
 struct aws_http_message *aws_s3_upload_part_message_new(
     struct aws_allocator *allocator,
     struct aws_http_message *base_message,
     struct aws_byte_buf *buffer,
     uint32_t part_number,
     const struct aws_string *upload_id,
     bool should_compute_content_md5,
```
DmitriyMusatkin, line 366: not for now, but we should really unify md5 with the rest of checksum at some point

TingDaoK:
> should we sanity check that the size of data pushed through the checksum matches expected payload size? we can even check response size here?

In the case of the checksum was calculated for some portion of the part, and client somehow reused the checksum, the server side will be able to catch the error since the checksum won't match the full part of data client sent. So, it's less a concern here.

DmitriyMusatkin (approved):

PR #547, opened 2025-08-15
Checksum stream with context

TingDaoK:
*Issue #, if available:*

- Let the checksum stream take a context.
- In which case, when the stream read to the end, it finalizes the checksum to the context.
- Since, the caller may not know when exact the stream finishes reading.
  - when the s3 request takes the checksum stream directly, we want to reuse the checksum if the previous read finish reading during retry.
  - But, in the meantime, S3 request don't know the stream finished reading or not. So, still let the stream to finalize the checksum after reading the end of the stream.

*Description of changes:*


TingDaoK:
included in https://github.com/awslabs/aws-c-s3/pull/549

PR #549, opened 2025-08-18
Streaming intergrate

TingDaoK:
*Issue #, if available:*
- Put everything together
- Add integration tests for file streaming.

*Description of changes:*
- Fix bunch of lifetime issue found from testing
- Add shutdown future to parallel input stream, so that it can be the last thing after all the event loop tasks finish.
- Blocking wait for the future from the same ELG can lead to dead lock. So, convert it to a callback from `part_streaming_input_stream`
- But `part_streaming_input_stream` is a input stream interface, that doesn't support async shutdown. 
- Introduce the async shutdown to the `parallel_input_stream`.

### TODO:
- Direct IO support

### DIFF
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -183,14 +183,15 @@ uint32_t aws_s3_client_get_max_active_connections(
         client->max_active_connections_override < max_active_connections) {
         max_active_connections = client->max_active_connections_override;
     }
-    if (meta_request && meta_request->fio_opts.should_stream && meta_request->fio_opts.disk_throughput > 0) {
+    if (meta_request && meta_request->fio_opts.should_stream && meta_request->fio_opts.disk_throughput_gbps > 0) {
         return aws_min_u32(
-            s_get_ideal_connection_number_from_throughput(meta_request->fio_opts.disk_throughput),
+            s_get_ideal_connection_number_from_throughput(meta_request->fio_opts.disk_throughput_gbps),
             max_active_connections);
     }
-    if (client->fio_opts.should_stream && client->fio_opts.disk_throughput > 0) {
+    if (client->fio_opts.should_stream && client->fio_opts.disk_throughput_gbps > 0) {
         return aws_min_u32(
-            s_get_ideal_connection_number_from_throughput(client->fio_opts.disk_throughput), max_active_connections);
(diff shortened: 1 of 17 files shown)


source/s3_client.c
```diff
@@ -149,6 +150,19 @@ void aws_s3_set_dns_ttl(size_t ttl) {
...
+    double ideal_connection_count_double = throughput_gps / s_throughput_per_connection_gbps;
+    /* round up and clamp */
+    ideal_connection_count_double = ceil(ideal_connection_count_double);
+    ideal_connection_count_double = aws_min_double(UINT32_MAX, ideal_connection_count_double);
+    return (uint32_t)ideal_connection_count_double;
+}
+
 /* Returns the max number of connections allowed.
  *
  * When meta request is NULL, this will return the overall allowed number of connections.
```
DmitriyMusatkin, line 168: if you expand the doc, this is no longer accurate right? it will now return the number of connections that can be made for that specific meta request and not just meta request type

source/s3_client.c
```diff
@@ -1170,6 +1189,39 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
     }
     if (options->send_filepath.len > 0) {
         ++body_source_count;
+        if (!content_length_found) {
+            /* The content length was not set, but it's a file based upload. Derive the content length from the file
+             * length */
+            struct aws_string *file_path = aws_string_new_from_cursor(client->allocator, &options->send_filepath);
```
DmitriyMusatkin, line 1195: can we just pull it into a helper that can be easily tested?

source/s3_client.c
```diff
@@ -1763,6 +1815,17 @@ static bool s_s3_client_should_update_meta_request(
         }
     }
 
+    if (meta_request->fio_opts.streaming_upload) {
```
DmitriyMusatkin, line 1818: can you add more doc somehow of how high level fio streaming reqs are scheduled. 
i.e. it will not schedule any more reqs if there are too many reqs in flight. 
what happens if we have fio streaming req along with other reqs? other reqs will keep pipeline full and num reqs in flight will never go down? does that mean that fio streaming does not get scheduled?

PR #575, opened 2025-10-03
Dynamic default part size

TingDaoK:
*Issue #, if available:*
- When the client is configured to allow using more memory, it makes more sense to default to larger range instead of a static 8MiB range. Based on the object characteristics stored in S3, memory constraints, and throughput target, we now dynamic calculate the range for getting the object.
- Parse the etag response from s3, as it will be `<hash>-<number of part>`, so that we can have an estimated part size that S3 stores the object.
- Keep the discovery phase to use the 8MiB as a default fallback.

*Description of changes:*
- Fix a race condition in the cancel tests and introduced a compiling flag to allow the test to stub into the function for testing the specific case. 

TODO: 
- ~Buffer Pool Refactor~
  - Currently the default buffer pool implementation will not handle the large range allocation from the pool.
  - To avoid wasting too many memory for small objects, we cannot just increase the buffer pool chunk size
  - We will refactor the buffer pool to handle different size chunks as a quick follow up.
  - https://github.com/awslabs/aws-c-s3/pull/584
- ~Integrate~
  - Integrate the buffer pool refactor and the dynamic size change
  - Apply reasonable upper bound and alignment for the implementation
  - Tests
  - https://github.com/awslabs/aws-c-s3/pull/586
- ~S3 Express Optimization~
  - Add a reasonable upbound for the range to fetch, maybe 1GiB for S3 GP, and 128MiB for S3Express
  - https://github.com/awslabs/aws-c-s3/pull/589
- ~Fuzz tests~
  - https://github.com/awslabs/aws-c-s3/pull/592
  - https://github.com/awslabs/aws-c-s3/pull/587
- ~The other PRs reviewed separately~
  - https://github.com/awslabs/aws-c-s3/pull/591
  - https://github.com/awslabs/aws-c-s3/pull/588

To follow up:
-  Refactor the math to derive the number of connections for s3 express. referring to https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/105

### DIFF
--- a/include/aws/s3/private/s3_util.h
+++ b/include/aws/s3/private/s3_util.h
@@ -169,6 +169,14 @@ extern const uint64_t g_default_max_part_size;
 
 AWS_S3_API
 extern const uint64_t g_s3_optimal_range_size_alignment;
+
+AWS_S3_API
+extern const uint32_t g_s3express_connection_limitation;
+AWS_S3_API
+extern const uint64_t g_s3express_connection_limitation_part_size_threshold;
+AWS_S3_API
+extern const uint64_t g_s3express_connection_limitation_object_size_threshold;
+
 /**
  * Returns AWS_S3_REQUEST_TYPE_UNKNOWN if name doesn't map to an enum value.
  */
@@ -359,6 +367,7 @@ int aws_s3_calculate_client_optimal_range_size(
  *
  * @param client_optimal_range_size The client-level optimal range size from initialization
  * @param estimated_object_stored_part_size Estimated size of object stored parts in S3
+ * @param is_express If the request is a s3express request or not.
  * @param out_request_optimal_range_size Output parameter for calculated request-level optimal range size
  * @return AWS_OP_SUCCESS on success, AWS_OP_ERR on failure (caller should fall back to client size)
  */
@@ -366,6 +375,7 @@ AWS_S3_API
 int aws_s3_calculate_request_optimal_range_size(
     uint64_t client_optimal_range_size,
     uint64_t estimated_object_stored_part_size,
+    bool is_express,
     uint64_t *out_request_optimal_range_size);
 
 /**
--- a/source/s3_auto_ranged_get.c
+++ b/source/s3_auto_ranged_get.c
@@ -787,7 +787,7 @@ static void s_s3_auto_ranged_get_request_finished(
                 error_code = AWS_ERROR_S3_MISSING_ETAG;
                 goto update_synced_data;
             }
-            /* Extract number of parts from ETag and calculate estimated part size */
+            /* Extract number of parts stored in S3 from ETag and calculate estimated part size */
             uint32_t num_parts = 0;
             if (aws_s3_extract_parts_from_etag(etag_header_value, &num_parts) == AWS_OP_SUCCESS && num_parts > 0) {
                 auto_ranged_get->estimated_object_stored_part_size = object_size / num_parts;
@@ -801,6 +801,7 @@ static void s_s3_auto_ranged_get_request_finished(
                     num_parts,
                     auto_ranged_get->estimated_object_stored_part_size);
             } else {
+                num_parts = 1;
                 /* Failed to parse ETags */
                 AWS_LOGF_WARN(
                     AWS_LS_S3_META_REQUEST,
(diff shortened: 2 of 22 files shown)


CMakeLists.txt
```diff
@@ -108,6 +108,7 @@ install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config.cmake"
 
 include(CTest)
 if (BUILD_TESTING)
+    add_definitions(-DAWS_C_S3_ENABLE_TEST_STUBS)
```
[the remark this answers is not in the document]
DmitriyMusatkin: do we enable this ourselves when we package java, python, etc...?
probably not a big concern in practice, it add a private api. we already expose a bunch of private apis used by tests only

include/aws/s3/private/s3_request.h
```diff
@@ -122,8 +122,19 @@ struct aws_s3_request_metrics {
         int error_code;
         /* Retry attempt. */
         uint32_t retry_attempt;
+        /* Is the memory for the request allocated from the buffer pool or not. */
+        bool memory_allocated_from_pool;
     } crt_info_metrics;
 
+    struct {
```
DmitriyMusatkin: thinking out loud:
maybe we should unify this with https://github.com/awslabs/aws-c-s3/blob/332dd22c47a7ed139eee71e7f219b764ef8cdf4c/include/aws/s3/private/s3_meta_request_impl.h#L297. 
we can probably do some of the same sanity checks in upload path as well

include/aws/s3/private/s3_util.h
```diff
@@ -240,12 +249,21 @@ void aws_s3_add_user_agent_header(struct aws_allocator *allocator, struct aws_ht
...
-    struct aws_allocator *allocator,
     struct aws_http_headers *response_headers,
     uint64_t *out_range_start,
     uint64_t *out_range_end,
     uint64_t *out_object_size);
 
+/* Given a Content-Range header value as a byte cursor, parses the range-start, range-end and
+ * object-size. All output arguments are optional. This is an optimized version that doesn't
+ * require string allocation. */
+AWS_S3_API
```
DmitriyMusatkin: its in private header? should we just get rid of non optimized version?

source/s3_auto_ranged_get.c
```diff
@@ -757,29 +764,77 @@ static void s_s3_auto_ranged_get_request_finished(
...
+                AWS_OP_SUCCESS) {
+                AWS_LOGF_TRACE(
+                    AWS_LS_S3_META_REQUEST,
+                    "id=%p ETag received for the meta request. value is: " PRInSTR "",
+                    (void *)meta_request,
+                    AWS_BYTE_CURSOR_PRI(etag_header_value));
 
-            if (aws_http_headers_get(request->send_data.response_headers, g_etag_header_name, &etag_header_value)) {
+                if (!auto_ranged_get->initial_message_has_if_match_header) {
+                    /* Store ETag if needed for If-Match header */
```
DmitriyMusatkin: why only if needed? isnt it easier to just always store it (i.e. the old way)

source/s3_util.c
```diff
@@ -398,44 +417,57 @@ void aws_s3_add_user_agent_header(struct aws_allocator *allocator, struct aws_ht
...
+int aws_s3_parse_content_range_cursor(
+    struct aws_byte_cursor content_range_cursor,
     uint64_t *out_range_start,
     uint64_t *out_range_end,
     uint64_t *out_object_size) {
-    AWS_PRECONDITION(allocator);
-    AWS_PRECONDITION(response_headers);
 
-    struct aws_byte_cursor content_range_header_value;
+    /* Expected Format of header is: "bytes StartByte-EndByte/TotalObjectSize" */
```
DmitriyMusatkin: does s3 always respect that or there are cases where it skips start and end (which is valid in http spec)

include/aws/s3/private/s3_util.h
TingDaoK: the previous version is taking the `struct aws_http_headers`. I already updated the previous function to skip the string allocation. I'll update the doc here to be less confusing.

source/s3_util.c
```diff
@@ -787,3 +829,194 @@ int aws_s3_check_headers_for_checksum(
...
+    struct aws_byte_cursor substr = {0};
+    struct aws_byte_cursor remaining_cursor = etag_cursor;
+    int split_count = 0;
+    struct aws_byte_cursor parts_count = {0};
+
+    /* Count splits and extract parts */
+    while (aws_byte_cursor_next_split(&remaining_cursor, '-', &substr)) {
+        split_count++;
+        if (split_count == 2) {
+            /* The ETag should follow the pattern <hash>-<parts_count>, so the second part is the parts count. */
```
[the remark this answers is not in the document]
TingDaoK: We need to confirm this with S3, if the value contains `-`, find it from the back will still result in unexpected behaviors. 

As far as I find, the etag value should be a hash value? So it should not contain `-`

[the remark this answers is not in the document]
TingDaoK: I'd rather to force the etag to be exactly have one or less `-` in it. So that if it's not the expected format, we will error out and fallback, instead of reading from the back try to parse the case where it's not formatted as expect.

PR #578, opened 2025-10-11
Add new metrics

azkrishpy:
*Issue #, if available:*

*Description of changes:*
added metrics for measuring, connection acquisition duration, retry delay duration, service call and full request duration.


include/aws/s3/private/s3_request.h
```diff
@@ -144,7 +144,7 @@ struct aws_s3_request_metrics {
         /* The host address of the request. */
         struct aws_string *host_address;
         /* The the request ID header value. */
-        struct aws_string *request_id;
+        struct aws_string *request_attempt_id;
```
TingDaoK: nope, let's keep this as the request id. Since it's a common name for server to fetch server side logs.

include/aws/s3/private/s3_request.h
```diff
@@ -158,6 +158,8 @@ struct aws_s3_request_metrics {
         struct aws_string *ip_address;
         /* The pointer to the connection that request was made from */
         void *connection_id;
+        /* The pointer to the request that the request attempt was made from */
+        void *request_id;
```
TingDaoK: let's keep this as our internal, ptr or something to be different from the request id for the server.

include/aws/s3/s3_client.h
```diff
@@ -1390,6 +1390,17 @@ int aws_s3_request_metrics_get_request_id(
...
     const struct aws_string **out_request_id);
 
+/**
+ * Get the extended request ID from aws_s3_request_metrics.
+ * If unavailable, AWS_ERROR_S3_METRIC_DATA_NOT_AVAILABLE will be raised.
+ * If available, out_amz_id_2 will be set to a string. Be warned this string's lifetime is tied to the metrics
+ * object.
+ **/
+AWS_S3_API
+int aws_s3_request_metrics_get_amz_id_2(
```
TingDaoK: let's rename this to extended request id. 
https://docs.aws.amazon.com/AmazonS3/latest/API/get-request-ids.html#sdk-request-ids
Looks like that is used in the s3 public docs.

The amz-id-2 is the header name.

TingDaoK (approved):

PR #624, opened 2026-04-24
Fix recognition of user provided unknown checksums

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Client might not be aware of all possible checksum. Add spec compliant handling for unknown checksum algos.

### DIFF
--- a/source/s3_checksums.c
+++ b/source/s3_checksums.c
@@ -472,7 +472,7 @@ static int s_init_and_verify_checksum_config_from_headers(
             return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
         }
 
-        if (aws_byte_cursor_starts_with_ignore_case(&header->name, &s_checksum_prefix)) {
+        if (aws_byte_cursor_starts_with_ignore_case(&header.name, &s_checksum_prefix)) {
             has_checksum_header = true;
             break;
         }
--- a/source/s3_checksums.c
+++ b/source/s3_checksums.c
@@ -455,6 +455,16 @@ int aws_checksum_compute(
 
 static const struct aws_byte_cursor s_checksum_prefix = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-checksum-");
 
+void s_byte_buf_to_upper(struct aws_byte_buf *buf) {
+    AWS_PRECONDITION(buf);
+
+    for (size_t i = 0; i < buf->len; ++i) {
+        if (buf->buffer[i] >= 'a' && buf->buffer[i] <= 'z') {
+            buf->buffer[i] = buf->buffer[i] + ('A' - 'a');
+        }
+    }
+}
+
 static int s_init_and_verify_checksum_config_from_headers(
     struct aws_s3_meta_request_checksum_config_storage *checksum_config,
     const struct aws_http_message *message,
@@ -466,6 +476,7 @@ static int s_init_and_verify_checksum_config_from_headers(
     AWS_ZERO_STRUCT(header_value);
 
     bool has_checksum_header = false;
+    struct aws_byte_cursor checksum_header_name;
     for (size_t i = 0; i < aws_http_headers_count(headers); ++i) {
         struct aws_http_header header;
         if (aws_http_headers_get_index(headers, i, &header)) {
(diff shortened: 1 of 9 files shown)


source/s3_checksums.c
```diff
@@ -453,6 +453,18 @@ int aws_checksum_compute(
     }
 }
 
+static const struct aws_byte_cursor s_checksum_prefix = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-checksum-");
+
+void s_byte_buf_to_upper(struct aws_byte_buf *buf) {
```
TingDaoK: trivial: static

should be moved to common, but next time :)

source/s3_checksums.c
```diff
@@ -463,34 +475,57 @@ static int s_init_and_verify_checksum_config_from_headers(
...
+    for (size_t i = 0; i < aws_http_headers_count(headers); ++i) {
+        struct aws_http_header header;
+        if (aws_http_headers_get_index(headers, i, &header)) {
+            return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        }
+
+        if (aws_byte_cursor_starts_with_ignore_case(&header.name, &s_checksum_prefix)) {
+            checksum_header_name = header.name;
+            has_checksum_header = true;
+            break;
```
TingDaoK, line 495: there could be `x-amz-checksum-type`, `x-amz-checksum-mode` and `x-amz-checksum-algorithm`?

DmitriyMusatkin: renamed it to mention checksum value. dont think we care about other headers here

TingDaoK: It's not very clear to me, but i think it makes sense because the other headers are NOT for put object? Maybe add some comments about it?

source/s3_checksums.c
```diff
@@ -463,34 +475,57 @@ static int s_init_and_verify_checksum_config_from_headers(
...
+        }
+    }
+
+    if (has_checksum_value_header) {
+        for (size_t i = 0; i < AWS_ARRAY_SIZE(s_checksum_algo_priority_list); i++) {
+            enum aws_s3_checksum_algorithm algorithm = s_checksum_algo_priority_list[i];
+            const struct aws_byte_cursor algorithm_header_name =
+                aws_get_http_header_name_from_checksum_algorithm(algorithm);
+            if (aws_http_headers_get(headers, algorithm_header_name, &header_value) == AWS_OP_SUCCESS) {
```
TingDaoK, line 504: just an idea that we can try to go through the first loop and in the first loop to get the algo name from the header, upper case it name to map it to our enum, and anything not in the map is unknown. So that we are not getting these couple loops here.

tests/s3_data_plane_tests.c
```diff
@@ -2332,6 +2332,118 @@ static int s_test_s3_put_object_multiple_with_filepath_streaming(struct aws_allo
     return s_test_s3_put_object_multiple_helper(allocator, true, true);
 }
 
+/*
+ * Helper that simulates the situation where the checksum value is provided in the message on put,
+ * but also checksum config is specified.
+ */
+static int s_test_s3_put_object_with_checksum_header_helper(struct aws_allocator *allocator) {
```
TingDaoK, line 2339: are we expecting the single part upload to succeed in this case?

TingDaoK (approved):

source/s3_checksums.c
DmitriyMusatkin: i havent changed existing code much.
old flow was to just check whether headers have any of the checksum value headers (i.e. x-amz-checksum-x) and then set algo appropriately. 
new code checks if we have a header that starts with x-amz-checksum- and then whether that matches any of the known ones. and if it does not, its marked as unknown and the rest of the flow is the same. 
im actually not sure if we are handling all those other headers correctly and maybe its something to revisit, but outside of this pr. 
i will leave comment to explain logic

source/s3_checksums.c
DmitriyMusatkin: i experimented with a couple ways to construct a single loop, but they all looked more ugly and harder to read then just breaking it into 2. im fine leaving a note here if we want to try to optimize this loop again

tests/s3_data_plane_tests.c
DmitriyMusatkin: yes. i think we already have tests for a single part checksum provided by user, so only added tests to cover missing behavior

PR #625, opened 2026-04-24
Add some documentation for CRT S3 Client

azkrishpy:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/s3_client.c
+++ b/source/s3_client.c
@@ -45,6 +45,66 @@
 #include <inttypes.h>
 #include <math.h>
 
+/*
+ * The s3 client splits user operations (meta requests) into concurrent HTTP requests
+ * and schedules them across pooled connections.
+ *
+ * A meta request is a user-initiated operation (GetObject, PutObject, etc). Each type
+ * has a state machine that produces individual HTTP requests on demand (range-GETs,
+ * UploadParts, etc). See s3_auto_ranged_get.c, s3_auto_ranged_put.c, s3_copy_object.c,
+ * s3_default_meta_request.c.
+ *
+ * A request goes through: buffer acquire -> prepare (build message + read body) -> sign -> queue -> send.
+ *
+ * An endpoint represents an S3 hostname and owns an HTTP connection manager that pools
+ * connections to that host.
+ *
+ * Threading model:
+ *
+ * All scheduling decisions run on a single event loop thread (process_work_event_loop),
+ * pinned at client init from the bootstrap ELG. This means threaded_data needs no lock.
+ *
+ * Other threads (user threads, prepare callbacks, connection callbacks) push work into
+ * synced_data (protected by a mutex) and call schedule_process_work_synced() to wake
+ * the work loop. The lock is only held long enough to move list pointers.
+ *
+ * Client data falls into three categories:
+ *
+ *   synced_data   - mutex-protected, any thread. Mailbox for pending_meta_request_work
+ *                   and prepared_requests. External threads deposit here, work loop drains.
+ *
+ *   threaded_data - no lock, only the process_work_event_loop thread touches it.
+ *                   Contains request_queue, active meta_requests list, and prep counters.
+ *                   Functions with "_threaded" suffix operate on this data.
+ *
+ *   stats         - atomic counters (num_requests_in_flight, etc). Any thread, no lock.
+ *
+ * Event loop groups:
+ *
+ *   Bootstrap ELG (client_bootstrap->event_loop_group) drives networking. HTTP connections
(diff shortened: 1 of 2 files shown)


source/s3_client.c
```diff
@@ -45,6 +45,66 @@
 #include <inttypes.h>
 #include <math.h>
 
+/*
```
DmitriyMusatkin, line 48: we have a design doc folder that in the past we intended as a place to put detailed description kinda docs. does that belong there?

source/s3_client.c
```diff
@@ -45,6 +45,66 @@
...
+ * s3_default_meta_request.c.
+ *
+ * A request goes through: buffer acquire -> prepare (build message + read body) -> sign -> queue -> send.
+ *
+ * An endpoint represents an S3 hostname and owns an HTTP connection manager that pools
+ * connections to that host.
+ *
+ * Threading model:
+ *
+ * All scheduling decisions run on a single event loop thread (process_work_event_loop),
```
DmitriyMusatkin, line 64: important point here: the way scheduler works is when it kicks off it moves all the pending work into separate area, i.e. it captures the state of the world and no one else is supposed to touch that state. then it does all the scheduling work and updates global state after. 
its totally possible that more work will arrive while schedule is running, but it will not be touched until next iteration of scheduler

source/s3_client.c
```diff
@@ -45,6 +45,66 @@
...
+ *   threaded_data - no lock, only the process_work_event_loop thread touches it.
+ *                   Contains request_queue, active meta_requests list, and prep counters.
+ *                   Functions with "_threaded" suffix operate on this data.
+ *
+ *   stats         - atomic counters (num_requests_in_flight, etc). Any thread, no lock.
+ *
+ * Event loop groups:
+ *
+ *   Bootstrap ELG (client_bootstrap->event_loop_group) drives networking. HTTP connections
+ *   are pinned to loops in this group. One loop from this group is designated as the
```
DmitriyMusatkin, line 85: pinned is a bit confusing in this case. you might want to expand on exactly what you mean

source/s3_client.c
```diff
@@ -1524,6 +1584,76 @@ static void s_s3_client_push_meta_request_synced(
...
 
+/*
+ * The scheduling system is a single-threaded work loop driven by event loop tasks.
+ *
+ * Any thread can call schedule_process_work_synced() (under lock) to post a task to
+ * the process_work_event_loop. If a task is already pending, this is a no-op.
+ * The task runs s_s3_client_process_work_task(), which calls the vtable's process_work,
+ * defaulting to s_s3_client_process_work_default(). All scheduling decisions happen there.
+ *
+ * schedule_process_work_synced() is called from:
```
DmitriyMusatkin, line 1595: i feel conflicted about mentioning specific function names in doc. will be really hard to keep in sync with actual code

source/s3_client.c
```diff
@@ -1524,6 +1584,76 @@ static void s_s3_client_push_meta_request_synced(
...
+ *      adjust num_requests_being_prepared, set process_work_task_scheduled=false, unlock.
+ *   2. Add newly arrived meta requests to threaded_data.meta_requests.
+ *   3. Call update_meta_requests_threaded() to ask each meta request for its next request,
+ *      then update_connections_threaded() to assign queued requests to HTTP connections.
+ *   4. Log stats.
+ *   5. If the client is shutting down and all work is drained, call finish_destroy().
+ *
+ * A request moves through the pipeline as follows:
+ *   meta_request->update() produces a request, which goes to s_acquire_mem_and_prepare_request()
+ *   to reserve a buffer from the pool (may block if memory is full). Then vtable->prepare_request()
```
DmitriyMusatkin, line 1613: not block. it will just async wait until mem frees up

source/s3_meta_request.c
```diff
@@ -3,6 +3,48 @@
...
+ * calls vtable->update() repeatedly, and the meta request decides what to produce next
+ * based on its current state and the results of previous requests.
+ *
+ * For PutObject, the state machine starts by producing a CreateMultipartUpload request.
+ * Once the response arrives with an upload_id, subsequent update() calls produce UploadPart
+ * requests until flow-control limits are hit or all parts are started. After all parts
+ * complete, it produces a CompleteMultipartUpload. When that finishes, update() returns
+ * work_remaining=false and the client removes the meta request.
+ *
+ * For GetObject, the state machine first sends a HeadObject or a ranged GET for part 1 to
```
DmitriyMusatkin, line 19: its a lot more complicated than that. there is a doc on that in design folder

PR #629, opened 2026-04-27
Expose max_parts_pending_read as an env variable

azkrishpy:
*Issue #, if available:*

*Description of changes:*
#313 added the concept of a `s_max_parts_pending_read` to control the number of parts pre-read to avoid reading an too many parts and hence throttle other requests. It also prevents edge-cases where where interdependent requests end up deadlocking due to lack of memory.
However, the static default previously set (5) might not perform as well for all types of file systems and disk throughputs.

This PR exposes the static value as an environment variable users can set in order to override the default value we provide.


TingDaoK (commented):
let's add a documentation in readme for the environment variables we read. 
Currently, we have only one in the readme about the memory limitation. 
https://github.com/awslabs/aws-c-s3#memory-limit 

we can have a section for all the environment variables that we support for aws-c-s3.

README.md
```diff
@@ -18,34 +18,59 @@ The AWS-C-S3 library is an asynchronous AWS S3 client focused on maximizing thro
...
-       // ... other configuration
-   };
+   Example Usage:
+
+   ```bash
+   export AWS_CRT_S3_MEMORY_LIMIT_IN_GIB=4  # 4 GiB limit
    ```
 
-2. **Via Environment Variable**: Set the `AWS_CRT_S3_MEMORY_LIMIT_IN_GIB` environment variable:
+   > [!TIP]
```
TingDaoK: trivial: this doesn't seem to be recognized by github

README.md
```diff
@@ -18,34 +18,59 @@ The AWS-C-S3 library is an asynchronous AWS S3 client focused on maximizing thro
...
+   >    };
+   > ```
 
-```bash
-   export AWS_CRT_S3_MEMORY_LIMIT_IN_GIB=4  # 4 GiB limit
+   **Default Behavior**:
+   If neither is set (config is 0 and environment variable is not set), the client sets a default memory limit based on the target throughput.
+
+   **Notes**:
+   * The limit applies per client. If multiple clients created, limit will apply to each separately.
```
TingDaoK, line 37: add that this value only supports positive integer

README.md
```diff
@@ -18,34 +18,59 @@ The AWS-C-S3 library is an asynchronous AWS S3 client focused on maximizing thro
...
+
+   Example Usage:
+
+   ```bash
+   export AWS_CRT_S3_MAX_PARTS_PENDING_READ=20
    ```
 
-**Priority**: The configuration value takes precedence over the environment variable. If `memory_limit_in_bytes` is set to a non-zero value in the config, the environment variable is ignored.
+   **Default Behavior**:
+   If not set, the default value is 5.
```
TingDaoK, line 62: are we aligned to not change the default value?

README.md
```diff
@@ -18,34 +18,59 @@ The AWS-C-S3 library is an asynchronous AWS S3 client focused on maximizing thro
...
+
+   **Notes**:
+   * The limit applies per client. If multiple clients created, limit will apply to each separately.
+   * The environment variable value must be a valid positive integer representing gigabytes (GiB).
+   * The value is converted from GiB to bytes internally (1 GiB = 1024³ bytes).
+   * Invalid values or overflow conditions will cause client creation to fail with `AWS_ERROR_INVALID_ARGUMENT`.
+
+2. **Maximum Parts Pending Read - `AWS_CRT_S3_MAX_PARTS_PENDING_READ`**
+
+   Controls the maximum number of parts that can be pending read from the input stream during a multipart upload. Higher values may improve upload throughput for large files by allowing more parts to be read in parallel, at the cost of higher memory usage.
```
TingDaoK: I think we should be clear that this value is to control one file upload, so that one upload won't take the full bandwidth of the client. 

It may improve the performance if the disk read speed can be benefited from more concurrent read. 

It ONLY really affecting the memory usage for single file use case and we cannot read fast enough. So, I would NOT mention the memory usage here.

README.md
```diff
@@ -18,34 +18,59 @@ The AWS-C-S3 library is an asynchronous AWS S3 client focused on maximizing thro
...
+
+   ```bash
+   export AWS_CRT_S3_MAX_PARTS_PENDING_READ=20
    ```
 
-**Priority**: The configuration value takes precedence over the environment variable. If `memory_limit_in_bytes` is set to a non-zero value in the config, the environment variable is ignored.
+   **Default Behavior**:
+   If not set, the default value is 5.
+
+   **Notes**:
```
TingDaoK, line 64: I think it's worth to mention that the number of parallel prepare will be limited by bandwidth of the client as well.

README.md
azkrishpy: yep my bad, just saw this and fixed it.

README.md
azkrishpy: yes, we will make any specific suggestions or change defaults as a future task if further benchmarking proves it might be worthwhile

README.md
azkrishpy: it is mentioned in the next line. should we change the order?

TingDaoK (approved):

PR #633, opened 2026-05-14
fix unknown checksum handling

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Before this pr sending default request with a body that had any header starting with "x-amz-checksum-" that was not a valid checksum would fail with a generic error about failing to asign the body. Ex. completeMPU call going through default that had a body and x-amz-checksum-type header. Note: this is manual call to completeMPU though default meta request and usual flow is not affected. 

2 changes made in this pr to mitigate the issue:
- dont treat headers like x-amz-checksum-type, x-amz-checksum-mode and x-amz-checksum-algorithm and unknown checksum type
- narrow down the cases where checksum is recomputed for default case. i.e. if req does not need checksum and there is no verification callback then do not compute it. checksum is never recomputed for completeMPU.

Note: there is still a weird case that if checksum is provided for default with body and there is a verification callback, then we recompute checksum. on one side it makes sense to recompute on the other it does not. so leaving as is for now


source/s3_checksum_context.c
```diff
@@ -48,6 +48,7 @@ static struct aws_s3_upload_request_checksum_context *s_s3_upload_request_checks
     context->algorithm = checksum_config->checksum_algorithm;
     context->location = checksum_config->location;
     context->encoded_checksum_size = aws_get_digest_size_from_checksum_algorithm(context->algorithm);
+    context->has_review_callback = checksum_config->full_object_checksum_callback;
```
TingDaoK: `full_object_checksum_callback` is for the caller to provide the full object checksum async after streaming the body.

We also have a `upload_review_callback`.

I believe we should check for `upload_review_callback` instead?

tests/s3_util_tests.c
```diff
@@ -1046,3 +1046,20 @@ static int s_test_s3_extract_parts_from_etag(struct aws_allocator *allocator, vo
...
     return 0;
 }
+
+AWS_TEST_CASE(test_s3_checksum_header, s_test_s3_checksum_header)
+static int s_test_s3_checksum_header(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+    aws_s3_library_init(allocator);
+
+    ASSERT_TRUE(aws_s3_is_checksum_value_header_name(aws_byte_cursor_from_c_str("x-amz-checksum-crc32")));
+    ASSERT_TRUE(aws_s3_is_checksum_value_header_name(aws_byte_cursor_from_c_str("x-amz-checksum-foo")));
```
TingDaoK, line 1056: trivial: I'd add some upper case to the value

source/s3_checksum_context.c
DmitriyMusatkin: oops you are right. wrong callback. fixing

TingDaoK (approved):
LGTM

PR #634, opened 2026-05-18
o_direct download support

TingDaoK:
*Issue #, if available:*

*Description of changes:*
- support direct io for download to file path
- currently only updating the body delivery event to write to file using direct io. So, it's only one thread kick off the write, which will not see much benefit of performance unless the file size is huge enough (in TiBs.)
- Fallback with warning if direct io is not supported, or not align with the page size
---
**Highlight:** With this change, download 5TiB file to a RAID0 disk with 150Gbps throughput, we are able to reach 109 Gb/s with 400GiB memory usage. Meanwhile same configurations and resource usage without direct_io write can only reach 15 Gb/s due to kernel limitation


source/s3_meta_request.c
```diff
@@ -2014,6 +2110,112 @@ static bool s_should_apply_backpressure(struct aws_s3_request *request) {
...
+
+    if (meta_request->recv_file_direct_io) {
+
+        if (aws_file_path_write_to_offset_direct_io(meta_request->recv_filepath, file_offset, *body)) {
+            if (meta_request->recv_file_direct_io_fallback_count == 0) {
+                AWS_LOGF_WARN(
+                    AWS_LS_S3_META_REQUEST,
+                    "id=%p Failed writing to file with O_DIRECT. aws-error:%s. Fallback.",
+                    (void *)meta_request,
+                    aws_error_name(error_code));
```
sbSteveK: error_code is never assigned a value different from `AWS_ERROR_SUCCESS` at this point. We should probably use `aws_last_error()` here.

source/s3_meta_request.c
```diff
@@ -2014,6 +2110,112 @@ static bool s_should_apply_backpressure(struct aws_s3_request *request) {
...
+                ++meta_request->recv_file_direct_io_fallback_count;
+            }
+            aws_reset_error();
+        }
+
+        /* Buffered fallback for unaligned data. recv_file is already open from init.
```
sbSteveK, line 2141: If this is fallback, shouldn't it be within the failure above? It looks like right now we are ALWAYS (error or success of `aws_file_path_write_to_offset_direct_io()` getting here and then running `s_buffered_write_to_recv_file()`

If `aws_file_path_write_to_offset_direct_io()` should we immediately return an `AWS_ERROR_SUCCESS`? i.e. at the end of the `if()` above, `else { return AWS_ERROR_SUCCESS; }` ?

source/s3_meta_request.c
```diff
@@ -2014,6 +2110,112 @@ static bool s_should_apply_backpressure(struct aws_s3_request *request) {
...
+         * Need to seek because O_DIRECT path doesn't update FILE* position. */
+        if (aws_fseek(meta_request->recv_file, (int64_t)file_offset, SEEK_SET) != AWS_OP_SUCCESS) {
+            return aws_last_error();
+        }
+        s_buffered_write_to_recv_file(meta_request, body, &error_code);
```
sbSteveK: This is the same as if we are !recv_file_direct_io. If we ALWAYS run this are we doing a double-write? Once using o-direct as well as whatever is the non-o-direct path? Not familiar enough to understand if this is intentional and we want to be doing both here.

source/s3_meta_request.c
```diff
@@ -569,6 +662,9 @@ static void s_s3_meta_request_destroy(void *user_data) {
             /* If the meta request succeed, the file should be closed from finish call. So it must be failing. */
             aws_file_delete(meta_request->recv_filepath);
         }
+    } else if (meta_request->recv_file_direct_io && meta_request->recv_file_delete_on_failure) {
```
sbSteveK: this `else if` is reached if the above `meta_request->recv_file` is NULL.

in `aws_s3_meta_request_init_base()` we have:
if (!meta_request->recv_file) { goto error; }
within the block, before `meta_request->recv_file_direct_io` gets set to true.

This code block is logically unreachable. If we don't ever intend to allow `recv_file_direct_io` to be used without recv_file, we should remove this dead block. If we do intend to allow it without the fallback, we should leave a comment here that explains that this will be of future use but that it's currently dead?

source/s3_meta_request.c
```diff
@@ -2417,6 +2573,10 @@ void aws_s3_meta_request_finish_default(struct aws_s3_meta_request *meta_request
         if (finish_result.error_code && meta_request->recv_file_delete_on_failure) {
             aws_file_delete(meta_request->recv_filepath);
         }
+    } else if (
```
sbSteveK: Same as above dead code comment. This will never be reached because `recv_file` is never NULL when `recv_file_direct_io` is true.

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -317,6 +317,17 @@ struct aws_s3_meta_request {
...
+    bool recv_file_direct_io;
+    /* Base file offset for O_DIRECT writes. 0 for CREATE_*, recv_file_position for WRITE_TO_POSITION,
+     * existing file size for CREATE_OR_APPEND. The actual write offset for each part is
+     * base_position + delivery_range_start. Only meaningful when recv_file_direct_io is true. */
+    uint64_t recv_file_direct_io_base_position;
+    /* Counter for how many times we fell back from O_DIRECT to buffered I/O for a single part.
+     * Init-time fallbacks (non-Linux, unaligned part_size, unaligned WRITE_TO_POSITION/APPEND offset)
+     * also increment this counter. The warning is only logged when this transitions from 0,
+     * to avoid log spam. */
+    size_t recv_file_direct_io_fallback_count;
```
sbSteveK, line 335: Trivial: If we aren't using the actual number of failures for anything, is there any reason not to use a bool we flip to true instead?

source/s3_meta_request.c
TingDaoK: yeah, that was wrong... I thought I fixed it, but apparently we git branch was messed up.

TingDaoK: aha... https://github.com/awslabs/aws-c-s3/blob/odirect_parallel-chunk/source/s3_meta_request.c#L2299-L2330 I fixed in another branch 🤦‍♂️

include/aws/s3/private/s3_meta_request_impl.h
TingDaoK: we are using it in the test to make sure we are not hitting more fallbacks then expected

sbSteveK (approved):

PR #636, opened 2026-05-28
Add randomize_get_part_order option for auto-ranged GET

erikfuller:
## Description

Adds a new `bool randomize_get_part_order` option to `aws_s3_meta_request_options` that causes the auto-ranged GET to fetch byte-range parts in a shuffled order instead of sequentially.

## Motivation

Enable better caching characteristics for large numbers of consumers downloading the same object. Without this option, consumers are blocked waiting for each sequential range. Randomizing the fetch order maximizes overall throughput when downloading the object by reducing contention on sequential byte ranges in caching layers.

## Design

- **Windowed shuffle**: Parts are shuffled within adjacent windows of 8 (matching `s_conservative_max_requests_in_flight`). This ensures the delivery pipeline can always make sequential progress within its buffering capacity, avoiding stalls or deadlocks on large objects while still distributing fetch order across concurrent consumers.
- **HEAD for discovery**: When enabled, object size is discovered via HEAD request (not a ranged GET of part 1), so all parts including part 1 participate in the shuffle.
- Delivery order to the caller is unaffected — the existing priority queue reorders parts back to sequential order.
- Graceful fallback to sequential ordering on allocation failure or if the random source is unavailable.
- Uses `aws_device_random_u64` for randomness via an internal `aws_s3_shuffle_uint32_array` helper.
- Only applies to `AWS_S3_META_REQUEST_TYPE_GET_OBJECT`.

## Testing

- `meta_request_auto_ranged_get_randomize_part_order` — unit test of shuffle function (valid permutation, non-sequential)
- `meta_request_shuffle_edge_cases` — edge cases (empty, single element, two elements)
- `test_s3_get_object_randomize_part_order` — integration test downloading a 10MB object with 1MB parts, verifies all parts fetched successfully

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*


source/s3_auto_ranged_get.c
```diff
@@ -970,6 +975,52 @@ static void s_s3_auto_ranged_get_request_finished(
...
+                    } else {
+                        /* Initialize with sequential part numbers [1..total] */
+                        for (uint32_t i = 0; i < total; ++i) {
+                            auto_ranged_get->shuffled_part_order[i] = i + 1;
+                        }
+
+                        /* Fisher-Yates shuffle. Modulo bias is negligible for typical part counts (< 10,000). */
+                        for (uint32_t i = total - 1; i > 0; --i) {
+                            uint32_t rand_val = 0;
+                            if (aws_device_random_u32(&rand_val)) {
```
erikfuller: kiro@
This won't block waiting for entropy, right?

source/s3_auto_ranged_get.c
```diff
@@ -970,6 +975,52 @@ static void s_s3_auto_ranged_get_request_finished(
...
+                    auto_ranged_get->shuffled_part_order =
+                        aws_mem_calloc(meta_request->allocator, total, sizeof(uint32_t));
+
+                    if (auto_ranged_get->shuffled_part_order == NULL) {
+                        AWS_LOGF_WARN(
+                            AWS_LS_S3_META_REQUEST,
+                            "id=%p: Failed to allocate shuffled part order array, falling back to sequential.",
+                            (void *)meta_request);
+                    } else {
+                        /* Initialize with sequential part numbers [1..total] */
```
erikfuller: kiro@
this comment is redundant - clearly visible in code. Please remove

source/s3_auto_ranged_get.c
erikfuller, line 982: kiro@
I see some a struct declared in `s3_auto_ranged_get.h` called `synced_data` that mentions locking, I'm assuming this is where that locking happens. Can you explain why our new fields don't need to participate in this locking?

tests/s3_data_plane_tests.c
```diff
@@ -5654,6 +5654,43 @@ static int s_test_s3_round_trip_dynamic_range_size_download_single_part(struct a
     return 0;
 }
 
+AWS_TEST_CASE(test_s3_get_object_randomize_part_order, s_test_s3_get_object_randomize_part_order)
+static int s_test_s3_get_object_randomize_part_order(struct aws_allocator *allocator, void *ctx) {
```
erikfuller, line 5658: kiro@
It's not clear to me what this test is validating beyond "it didn't fail". Is that enough? Other tests include some brief comments explaining what the test is looking for - please add some comments here.

tests/s3_meta_request_test.c
```diff
@@ -53,6 +54,87 @@ TEST_CASE(meta_request_auto_ranged_get_new_error_handling) {
...
+        .part_size = 5 * 1024 * 1024,
+    };
+    ASSERT_SUCCESS(aws_s3_tester_client_new(&tester, &client_options, &client));
+
+    struct aws_http_message *message = aws_http_message_new_request(allocator);
+    ASSERT_SUCCESS(aws_http_message_set_request_method(message, aws_http_method_get));
+    ASSERT_SUCCESS(aws_http_message_set_request_path(message, aws_byte_cursor_from_c_str("/test-object")));
+    struct aws_http_header host_header = {
+        .name = g_host_header_name,
+        .value = aws_byte_cursor_from_c_str("test-bucket.s3.us-east-1.amazonaws.com"),
```
erikfuller: kiro@
This might be a real bucket, not one for these tests. I think I saw other tests referencing `aws-c-s3-test-bucket`, have a look and see if there's a more appropriate test value to use here.

tests/s3_meta_request_test.c
```diff
@@ -53,6 +54,87 @@ TEST_CASE(meta_request_auto_ranged_get_new_error_handling) {
...
+    };
+    ASSERT_SUCCESS(aws_http_message_add_header(message, host_header));
+
+    struct aws_s3_meta_request_options options;
+    AWS_ZERO_STRUCT(options);
+    options.message = message;
+    options.type = AWS_S3_META_REQUEST_TYPE_GET_OBJECT;
+    options.randomize_get_part_order = true;
```
erikfuller: kiro@
This library gets wrapped in other libraries. Can you put some example code here what it would look like to use this feature from the Java or Python libraries? 
* https://github.com/awslabs/aws-crt-java
* https://github.com/awslabs/aws-crt-python

tests/s3_meta_request_test.c
```diff
@@ -53,6 +54,87 @@ TEST_CASE(meta_request_auto_ranged_get_new_error_handling) {
...
+
+    struct aws_s3_meta_request *meta_request =
+        aws_s3_meta_request_auto_ranged_get_new(allocator, client, 5 * 1024 * 1024, false, &options);
+    ASSERT_NOT_NULL(meta_request);
+
+    struct aws_s3_auto_ranged_get *auto_ranged_get = meta_request->impl;
+    ASSERT_TRUE(auto_ranged_get->randomize_part_order);
+
+    /* Simulate what happens after HEAD response: allocate and shuffle a part order array.
+     * We test the shuffle logic directly since triggering the full HEAD flow requires network. */
```
erikfuller: kiro@
This seems pretty backwards. Would it make more sense to create a function which can calculate the shuffled array? Alternatively, could we inspect the shuffled ordering directly?

There must be another test that does ranged gets. It may be better to try and simply replicate that with the randomized setting enabled. Maybe there's a way to capture with the mocks the ordering which the parts are requested? Then we can validate at least that all parts are fetched and it wasn't in pure sequential order.

tests/s3_tester.c
```diff
@@ -1628,6 +1628,7 @@ int aws_s3_tester_send_meta_request_with_options(
             }
             meta_request_options.message = message;
             meta_request_options.force_dynamic_part_size = options->get_options.force_dynamic_part_size;
+    meta_request_options.randomize_get_part_order = options->get_options.randomize_get_part_order;
```
erikfuller: kiro@
this leading whitespace looks off

erikfuller (commented):
Tagged items for Kiro to review and address.

source/s3_auto_ranged_get.c
erikfuller: No, it will not block. `aws_device_random_u32` delegates to `aws_device_random_buffer` which reads from `/dev/urandom` (non-blocking, never waits for entropy). Source: `aws-c-common/source/posix/device_random.c`.

source/s3_auto_ranged_get.c
erikfuller: Done — removed.

source/s3_auto_ranged_get.c
erikfuller: The new fields don't need synced_data locking because:

1. `randomize_part_order` (bool) is set once at construction in `aws_s3_meta_request_auto_ranged_get_new()` and never modified — it's effectively immutable.

2. `shuffled_part_order` (array pointer) is written once inside the locked section where `total_num_parts` is set (the `update_synced_data:` path after HEAD response), and only read afterward in `s_s3_auto_ranged_get_update()` which also runs under the same meta request lock. So the write-once/read-many access is already protected by the existing lock that guards `synced_data`.

The pointer itself acts as the gate — if it's NULL, we fall back to sequential. It transitions from NULL → allocated exactly once, under the lock.

tests/s3_data_plane_tests.c
erikfuller: Good point — added a block comment explaining the test validates that the streaming layer correctly reorders randomly-fetched parts back to sequential delivery. A successful download with correct content proves the end-to-end path works.

tests/s3_meta_request_test.c
erikfuller: Fixed — changed to `aws-c-s3-test-bucket.s3.us-east-1.amazonaws.com` which matches the project's standard `g_test_bucket_name` default.

tests/s3_meta_request_test.c
erikfuller: Neither Java nor Python CRT bindings currently expose `force_dynamic_part_size` or similar per-request boolean options yet, so this would be new plumbing in both. Here's what it would look like once added:

**Java** (`aws-crt-java` — `S3MetaRequestOptions` uses fluent setters):
```java
S3MetaRequestOptions opts = new S3MetaRequestOptions()
    .withMetaRequestType(MetaRequestType.GET_OBJECT)
    .withHttpRequest(request)
    .withRandomizeGetPartOrder(true)
    .withResponseHandler(handler);
S3MetaRequest metaRequest = s3Client.makeMetaRequest(opts);
```

**Python** (`aws-crt-python` — `S3Client.make_request()` uses kwargs):
```python
s3_request = s3_client.make_request(
    type=S3RequestType.GET_OBJECT,
    request=http_request,
    randomize_get_part_order=True,
    on_done=on_done_callback,
)
```

Both would require adding the field to their respective JNI/CPython bridge layers to pass through to the C `aws_s3_meta_request_options` struct.

tests/s3_meta_request_test.c
erikfuller: Fair point. I've updated the test comment to explain why we duplicate the logic: the production shuffle runs inside the locked update path after HEAD response sets `total_num_parts`, which we can't trigger without network. The test now clearly states it's validating array properties (valid permutation, non-sequential) and that the integration test (`test_s3_get_object_randomize_part_order`) exercises the full end-to-end path.

Re: inspecting the shuffle directly — the test does inspect `auto_ranged_get->shuffled_part_order` directly after populating it. The "backwards" part is that we replicate the shuffle rather than calling the production code path.

Re: using metrics to capture request ordering — that's a great idea for the integration test. The telemetry system (`test_results.synced_data.succeed_metrics`) captures part numbers via `aws_s3_request_metrics_get_part_number()`. I could enhance the integration test to verify parts were requested in non-sequential order. Want me to add that?

tests/s3_tester.c
erikfuller: Fixed — aligned to 12 spaces to match the surrounding block.

DmitriyMusatkin:
Without out of order delivery of data to consumer this option might have big impact on perf - s3 client limits how much work can be in flight at any given point in time and current delivery mechanism will cache parts until they can be delivered. so what might end up happening on large objects is that data delivery to consumer is blocked on the next part, but because of the randomization next part might not be finished for a while, which might stall the pipeline or deadlock in the worst case (hit mem limit on how many parts we can buffer, but we cannot flush delivery queue to free up mem because next part is not ready).
not against this change in general and we considered it in the past, but i think we would need to couple this with out of order delivery for it to work

tests/s3_meta_request_test.c
erikfuller: Did some iterations to get this to a better place.

tests/s3_data_plane_tests.c
```diff
@@ -5654,6 +5654,74 @@ static int s_test_s3_round_trip_dynamic_range_size_download_single_part(struct a
...
+        if (type != AWS_S3_REQUEST_TYPE_GET_OBJECT) {
+            continue;
+        }
+        uint32_t part_number = 0;
+        aws_s3_request_metrics_get_part_number(metrics, &part_number);
+        get_part_numbers[get_count++] = part_number;
+    }
+
+    /* With 1MB parts and a 10MB object, we expect ~10 parts */
+    ASSERT_TRUE(get_count >= 2);
```
erikfuller: kiro@
Why can't we know predictably what the `get_count` should be here?

tests/s3_data_plane_tests.c
```diff
@@ -5654,6 +5654,74 @@ static int s_test_s3_round_trip_dynamic_range_size_download_single_part(struct a
     return 0;
 }
 
+AWS_TEST_CASE(test_s3_get_object_randomize_part_order, s_test_s3_get_object_randomize_part_order)
+static int s_test_s3_get_object_randomize_part_order(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+
+    /* Verify that downloading with randomize_get_part_order=true succeeds and
+     * that byte-range parts were NOT requested in sequential order. */
```
erikfuller: kiro@ 
I don't think there's a validation in this test for non-sequential, only a test that all parts were retrieved.

erikfuller (commented):
Comments for agent review/response

tests/s3_data_plane_tests.c
erikfuller: Fixed — now asserts exactly 10 parts (1MB part_size × 10MB object).

tests/s3_data_plane_tests.c
erikfuller: You're right. I attempted adding a non-sequential assertion but it's not reliable here — the metrics record *completion* order, not *request issue* order. With small 1MB parts on parallel connections, S3 responds fast enough that completion order often appears sequential despite requests going out shuffled.

The non-sequential request ordering is validated by:
1. The unit test (`meta_request_auto_ranged_get_randomize_part_order`) which directly tests the shuffle function
2. Debug logs (visible at DEBUG level) showing "Returning request for part X of 10" in shuffled order

This integration test validates the end-to-end path works correctly: all 10 parts fetched, no duplicates, no missing parts, download succeeds with correct data delivery.

erikfuller: Correction to my earlier reply: the metrics are recorded in *delivery order* (after the priority queue reorders parts back to sequential for the caller), not completion order. So metrics will always show parts 1, 2, 3... regardless of fetch order — they can't be used to observe the shuffle.

The non-sequential fetch ordering is validated by:
1. The unit test (`meta_request_auto_ranged_get_randomize_part_order`) which directly tests `aws_s3_shuffle_uint32_array`
2. Debug logs showing "Returning request for part X of 10" in shuffled order

This integration test validates the end-to-end correctness: exactly 10 parts fetched, all present (valid permutation of 1..10), download succeeds with correct sequential data delivery to the caller.

TingDaoK:
- https://github.com/awslabs/aws-c-s3/pull/679 out of order write.
- https://github.com/awslabs/aws-c-s3/pull/682 issue range-get cross regions. 

We plan to enable an opt-in for out of order delivery and once out of order delivery enabled, we will issues its range requests across as many far-apart regions of the object as it has connections, which should address the contention on sequential byte ranges in caching layers as well. 

Will it meet your expectation here?

PR #637, opened 2026-06-01
Add gpu instance platform info

azkrishpy:
*Issue #, if available:*

*Description of changes:*
Adding new instance types: p5e, p5en, p6-b200 and p6-b300. Throughput values are configured to maximum single NIC throughput for each instance because CRT uses one NIC by default unless overridden by client configuration (at which point the target throughput can be overridden too).

Recommended throughput values are based on single NIC throughput of each instance gathered from [official documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-acc-inst-types.html).


include/aws/s3/s3.h
```diff
@@ -77,8 +77,7 @@ struct aws_s3_platform_info {
     struct aws_byte_cursor instance_type;
     /* max throughput for this instance type, in gigabits per second */
     double max_throughput_gbps;
-    /* The current build of this library specifically knows an optimal configuration for this
-     * platform */
+    /* If set to true, allows cli and boto to load CRT on the instance by default */
```
TingDaoK: Avoid mentioning downstream, specially in the public doc. 

```
/* Whether this library recommends automatically applying its optimized
  * configuration for this instance type without explicit user opt-in. */
bool has_recommended_configuration;
```

TingDaoK (approved):

PR #644, opened 2026-06-12
fix(auto-ranged-get): honor object_size_hint

### LINKED ISSUE #639: Use object_size_hint to size the discovery buffer for small, non-ranged GetObject
### Describe the feature

When downloading many small objects in parallel via auto-ranged-get, each meta request's first (size-discovery) request reserves a full `part_size` buffer, which is pulled from the buffer pool's primary area and pins an entire part-sized chunk. For workloads doing N parallel small gets, peak memory scales as `N * part_size` even though each object is tiny. A caller-supplied `object_size_hint` should be allowed to shrink this first-buffer reservation so small objects don't each pin a full chunk.

### Current behavior (main)

For a **non-ranged** GetObject, `s_s3_get_request_type_for_discovering_object_size` (`source/s3_auto_ranged_get.c`) selects the discovery request type before it ever inspects the hint:

```c
if (!meta_request->checksum_config.validate_response_checksum)
    return AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_RANGE;     // returns here by default
...
if (object_size_hint_available && object_size_hint <= part_size)         // unreachable on the default path
    return AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_PART_NUMBER_1;
```

Two independent gaps result:

1. **The hint is not consulted at all** for a non-ranged get unless response-checksum validation is enabled — the `!validate_response_checksum` early return shadows the `object_size_hint` branch.
2. **Even when it is consulted**, the discovery request is still reserved at `part_size` (`request->part_range_end = part_size - 1`). The reservation size equals `aws_s3_request_get_payload_size()` = `part_range_end - part_range_start + 1` (`source/s3_request.c`), so the pool always reserves `part_size`.

Because the default buffer pool acquires the **primary** area in whole chunks (`chunk_size == part_size`, rounded up — see `s_primary_acquire_synced` in `source/s3_default_buffer_pool.c`), a `part_size` reservation pins one full chunk per in-flight get, regardless of the actual object size.

Net effect: `object_size_hint` (`include/aws/s3/s3_client.h`) currently has no impact on download buffer memory on any path. It is referenced only in the constructor and in the one branch above.

This gap is already acknowledged in-code by an existing TODO directly above that branch (https://github.com/awslabs/aws-c-s3/blob/main/source/s3_auto_ranged_get.c#L182):
```c
/* If the object_size_hint indicates that it is a small one part file, then try to get the file directly
 * TODO: Bypass memory limiter so that we don't overallocate memory for small files
 */
```

### Use Case

A client issuing many concurrent small, non-ranged `GetObject` calls observes memory ~ `concurrency * part_size`. Supplying `object_size_hint` does not reduce this today.

### Proposed Solution

When `object_size_hint` is available and small, use it to size the first discovery request's buffer instead of `part_size`:

1. Make the hint reachable on the non-ranged path (not gated behind `validate_response_checksum`), so a small hint can route discovery to `GET_OBJECT_WITH_PART_NUMBER_1` (or an equivalently small ranged get).
4. Size the discovery request's reservation to `aws_min_u64(object_size_hint, part_size)` (aligned) rather than `part_size`, so the pool reserves the smaller amount.

### Other Information

_No response_

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

azkrishpy:
Issue #, if available:

Fixes #639 — `object_size_hint` was silently ignored for the common (no-checksum) download path, and even when it did take effect it still over-reserved a full `part_size` buffer from the pool.

Description of changes:

When object_size_hint fits within part_size, we skip the HEAD request and issue GET_OBJECT_WITH_PART_NUMBER_1 directly, reserving only min(hint, part_size) bytes from the buffer pool instead of the full part_size. This saves a round-trip and reduces memory usage for small objects.

If the hint is wrong, we cancel before any body bytes arrive and retry. If it exceeds buffer_size, we use the content-length information we got and proceed to the standard ranged-get path.

changed `AWS_ERROR_S3_INTERNAL_PART_SIZE_MISMATCH_RETRYING_WITH_RANGE`
 to `AWS_ERROR_S3_INTERNAL_BUFFER_SIZE_MISMATCH_RETRYING_WITH_RANGE`


include/aws/s3/private/s3_auto_ranged_get.h
```diff
@@ -40,6 +40,9 @@ struct aws_s3_auto_ranged_get {
 
     uint64_t object_size_hint;
     bool object_size_hint_available;
+    /* Set when a hint-based PART_NUMBER_1 request was cancelled because the actual content was larger than the hint.
+     * Causes the next PART_NUMBER_1 attempt to use a full part_size buffer instead of the (wrong) hint size. */
```
TingDaoK: I'd image that with the first PART_NUMBER_1 that found out the content was larger than the hint, we will NOT do another PART_NUMBER_1. Since we already have the content length, we can move on to the phase after the discovery.

source/s3_meta_request.c
```diff
@@ -1480,20 +1480,21 @@ static int s_s3_meta_request_headers_block_done(
...
      */
     if (request->request_type == AWS_S3_REQUEST_TYPE_GET_OBJECT &&
         request->request_tag == AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_PART_NUMBER_1) {
         uint64_t content_length;
+        uint64_t reserved_size = request->part_range_end - request->part_range_start + 1;
         if (!aws_s3_parse_content_length_response_header(
                 request->allocator, request->send_data.response_headers, &content_length) &&
-            content_length > meta_request->part_size) {
+            content_length > reserved_size) {
             return aws_raise_error(AWS_ERROR_S3_INTERNAL_PART_SIZE_MISMATCH_RETRYING_WITH_RANGE);
```
TingDaoK: well, although it's an existing behavior, but is there a content-range header? can we can get full object size and move on instead retry the discovery phase?

source/s3_auto_ranged_get.c
```diff
@@ -961,7 +970,11 @@ static void s_s3_auto_ranged_get_request_finished(
             auto_ranged_get->synced_data.object_range_empty = (object_size == 0);
             auto_ranged_get->synced_data.object_range_start = object_range_start;
             auto_ranged_get->synced_data.object_range_end = object_range_end;
-            if (!first_part_size_mismatch && first_part_size) {
```
azkrishpy, line 974: Not sure if I am missing something, but we can store the first_part_size regardless of whether there was a mismatch since we are restarting from 0 anyway and the range calculation deals with the first_part_size we record.

source/s3_auto_ranged_get.c
```diff
@@ -173,18 +173,19 @@ static enum aws_s3_auto_ranged_get_request_type s_s3_get_request_type_for_discov
...
+     * If the hint is wrong and the object is larger, the request will be cancelled and retried with ranged gets. */
+    if (auto_ranged_get->object_size_hint_available && auto_ranged_get->object_size_hint > 0 &&
+        auto_ranged_get->object_size_hint <= meta_request->part_size) {
+        return AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_PART_NUMBER_1;
+    }
+
     /* If we don't need checksum validation, then discover the size of the object while trying to get the first part. */
     if (!meta_request->checksum_config.validate_response_checksum) {
         return AWS_S3_AUTO_RANGE_GET_REQUEST_TYPE_GET_OBJECT_WITH_RANGE;
     }
```
TingDaoK, line 188: If we switch this order, we should update the design graph we have as well.

tests/s3_data_plane_tests.c
```diff
@@ -9338,3 +9338,144 @@ static int s_test_s3_default_get_without_content_length(struct aws_allocator *al
...
+    ASSERT_UINT_EQUALS(1, aws_array_list_length(&test_results.synced_data.succeed_metrics));
+    {
+        struct aws_s3_request_metrics *metrics = NULL;
+        ASSERT_SUCCESS(aws_array_list_get_at(&test_results.synced_data.succeed_metrics, (void **)&metrics, 0));
+        enum aws_s3_request_type request_type = AWS_S3_REQUEST_TYPE_UNKNOWN;
+        aws_s3_request_metrics_get_request_type(metrics, &request_type);
+        ASSERT_UINT_EQUALS(AWS_S3_REQUEST_TYPE_GET_OBJECT, request_type);
+        uint64_t range_start = UINT64_MAX;
+        aws_s3_request_metrics_get_part_range_start(metrics, &range_start);
+        ASSERT_UINT_EQUALS(0, range_start);
```
TingDaoK, line 10098: we should also capture the request range end as well? Man, I don't even know we will do a range end with the part size or the object size for this case.

azkrishpy: Turns out we actually use the object size. 

https://github.com/awslabs/aws-c-s3/blob/57acb2b8fd757959acab410255353816f0143262/source/s3_util.c#L663-L665

docs/images/GetObjectFlow.svg
TingDaoK, line 1: for 3, we should also mention:
```
This avoids a HEAD request and sizes the buffer reservation to the hint rather than full part_size.
```

TingDaoK (approved):
fix & ship

PR #645, opened 2026-06-17
Support deferred buffer reservations in async-write path

yerzhan7:
*Issue #, if available:* N/A

*Description of changes:*

Lets a custom buffer pool **defer** fulfilling an async-write part-buffer reservation
instead of having the write path fail. Previously `aws_s3_meta_request_poll_write()`
required the reservation to be granted synchronously; if a pool handed back a
not-yet-done future, the call treated it as illegal usage and terminated the meta
request.

It mirrors the existing deferral model used by the regular (non-write) request path
(`s_on_pool_buffer_reserved` in `source/s3_client.c`).

- When the pool returns a not-yet-done reservation future, `poll_write()` now:
  - stores the future in the new `synced_data.async_write.buffered_data_ticket_future` field,
  - registers the caller's `waker` and returns `result.is_pending = true`,
  - acquires a reference on the meta request and registers an event-loop completion
    callback on the future.
- New completion callback `s_s3_meta_request_on_async_write_buffer_reserved()` (runs on
  the meta request's `io_event_loop`):
  - on success, claims the buffer so the resumed `poll_write()` can copy data into it;
  - on error, fails the meta request and kicks the process-work loop;
  - handles the "fulfilled after the meta request already finished" case by returning the
    ticket to the pool;
  - invokes the waker and releases the meta-request reference.
- `aws_s3_meta_request_finish_default()` now errors out any pending reservation future on
  finish/cancel

---


source/s3_meta_request.c
```diff
@@ -2644,23 +2734,34 @@ struct aws_s3_meta_request_poll_write_result aws_s3_meta_request_poll_write(
...
                     (void *)meta_request);
-                illegal_usage_terminate_meta_request = true;
-                aws_future_s3_buffer_ticket_release(buffered_ticket_future);
+
+                meta_request->synced_data.async_write.buffered_data_ticket_future = buffered_ticket_future;
+                meta_request->synced_data.async_write.waker = waker;
+                meta_request->synced_data.async_write.waker_user_data = user_data;
+                result.is_pending = true;
+
+                aws_s3_meta_request_acquire(meta_request);
```
TingDaoK, line 2757: Trivial: add some comments about the lifetime here, 

```suggestion
                /* Keep the meta request alive until the ticket future callback invokes. */
                aws_s3_meta_request_acquire(meta_request);
```

source/s3_meta_request.c
```diff
@@ -578,6 +578,7 @@ static void s_s3_meta_request_destroy(void *user_data) {
             aws_s3_buffer_pool_release_special_size(meta_request->client->buffer_pool, meta_request->part_size);
         }
         aws_s3_buffer_ticket_release(meta_request->synced_data.async_write.buffered_data_ticket);
+        AWS_ASSERT(meta_request->synced_data.async_write.buffered_data_ticket_future == NULL);
```
TingDaoK, line 582: trivial: this lifetime management is tricky and easy to be messed up, make some comments to help people understanding the process. 
```suggestion
        /* pending buffer acquisition will keep meta request alive from destroying.  */
        AWS_ASSERT(meta_request->synced_data.async_write.buffered_data_ticket_future == NULL);
```

source/s3_meta_request.c
```diff
@@ -2384,7 +2385,15 @@ void aws_s3_meta_request_finish_default(struct aws_s3_meta_request *meta_request
...
 
         /* Clean out any pending async-write future */
-        if (meta_request->synced_data.async_write.waker != NULL) {
+        if (meta_request->synced_data.async_write.buffered_data_ticket_future != NULL) {
+            /* We're waiting on the buffer pool to fulfill a (deferred) ticket reservation.
+             * Error it out so its completion callback fires; that callback owns invoking the
+             * waker and cleaning up the future (and releasing the ref it holds on the meta request).
+             * NOTE: set_error schedules the callback as an event-loop task (it was registered with
+             * register_event_loop_callback), so this does NOT run the callback synchronously here. */
+            aws_future_s3_buffer_ticket_set_error(
```
TingDaoK, line 2395: the only other concern I have for this is we will invoke the callback of waker with a different order now, the `finish_callback` will happen in this call, and then the future complete callback will be invoked as a new event afterward. 

Usually I'd think the finish callback to be basically the last thing to happen. 
But, arguably the waker is not part of the meta request callbacks. And you have a test for this case already.  

I think it's okay here, just that the caller may need to be careful. eg: don't cleanup the on-going write from the finish callback.

TingDaoK (approved):
thanks! looks good to me, just couple trivial comments on the lifetime, it is tricky to manage the lifetime with async threads going on.

source/s3_meta_request.c
yerzhan7: Fixed in https://github.com/awslabs/aws-c-s3/pull/645/commits/d2dd5b1cf4690c339d6ccc49ad521289f178e050

Thanks for reviewing!

source/s3_meta_request.c
yerzhan7: Fixed

PR #649, opened 2026-06-25
[feat.] async pause and on_error_resume_token

TingDaoK:
## Summary

Adds **async pause** support and an **`on_error_resume_token`** callback to `aws-c-s3`, extending pause/resume from upload-only to also cover downloads (GET), and generalizing on-error token capture across meta request types.

## Motivation

Today, `aws_s3_meta_request_pause()` (synchronous) only supports upload (PUT) meta requests — calling it on a download fails with `AWS_ERROR_UNSUPPORTED_OPERATION`. There's also no way to get a resume token when a meta request fails unexpectedly (only on explicit pause). This PR:
- Adds `aws_s3_meta_request_pause_async()`, which works for **both uploads and downloads**, and waits for in-flight work (file writes for GET, in-flight parts for PUT) to flush before delivering the resume token via callback.
- Adds `on_error_resume_token`, an optional callback on `aws_s3_meta_request_options` that fires with a resume token when a meta request fails unexpectedly, so callers can persist state and resume later without re-transferring completed parts — supported for both PUT and GET.

## Key changes

**Public API (`s3_client.h`)**
- New `aws_s3_meta_request_pause_complete_fn` typedef, used by both `pause_async`'s `on_complete` and the new `on_error_resume_token` option.
- New `aws_s3_meta_request_pause_async()` API.
- Clarified doc on the existing synchronous `aws_s3_meta_request_pause()`: upload-only; downloads must use `pause_async`.
- Extended the resume token getters to cover download-specific fields: `etag`, `version_id`, `s3_object_last_modified`, `object_size`, `object_range_start`/`object_range_end`, `continues_downloaded_bytes`, `total_downloaded_bytes`, `file_last_modified_epoch_ns`. Getters return 0/empty when called on the "wrong" token type (use `aws_s3_meta_request_resume_token_type()` to disambiguate).

## TODO

- Consuming a download resume token to actually resume a GET is not yet supported — there's no public constructor to build a GET meta request from an externally-provided download resume token. Workaround: manually issue a new ranged GET using `aws_s3_meta_request_resume_token_continues_downloaded_bytes()` (offset from `object_range_start`) as the new range start, through the end of the original download (`object_range_end`, or `object_size - 1` for an unranged download).


include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -223,6 +230,13 @@ struct aws_s3_meta_request {
          * failed.)*/
         uint32_t num_parts_delivery_completed;
 
+        /* Total number of response-body bytes successfully delivered to the caller's sink
+         * (file or body callback), in order with no gaps. Used to build the download resume
+         * token on pause/error.
+         * TODO: delivery is strictly sequential today, so this is both the contiguous prefix and the total; a future
+         * parallel-write delivery path will need to track the two separately. */
+        uint64_t num_bytes_delivered;
```
sbSteveK, line 238: Trivial: If we will have a separate bytes_delivered in the future, should we explicitly label this as sequential now? `num_bytes_delivered_sequential`? That could make things clearer for what's being used where it's used. Especially if we have a second similar non-sequential field later.

include/aws/s3/s3_client.h
```diff
@@ -760,6 +760,20 @@ struct aws_s3_checksum_config {
  * 5) If the data is already in memory and you want the client to upload it with no extra copy or
  *    allocation, set `request_body` (DEFAULT meta request only).
  */
+/**
```
sbSteveK: Trivial: we might want to expand the typedef since this is a public header. We currently explain when the callback delivers the resume token for downloads only. Might be useful to expand to the three contexts when it's called. e.g. 
* Invoked in three contexts:
* - After aws_s3_meta_request_pause_async() completes on an upload (PUT):
*   waits for in-flight parts to finish, then delivers the upload resume token.
* - After aws_s3_meta_request_pause_async() completes on a download (GET):
*   waits for in-flight file writes to flush, then delivers the download resume token.
* - When a meta request fails and on_error_resume_token was set:
*   delivers whatever resumable state was captured at failure time.

include/aws/s3/private/s3_client_impl.h
```diff
@@ -453,19 +453,30 @@ struct aws_s3_meta_request_resume_token {
...
 
     /* Note: this field is used only when s3 tells us that upload id no longer
     exists, and if this indicates that all parts have already been uploaded,
     request is completed instead of failing it.*/
     size_t num_parts_completed;
+
+    /* Download-specific fields */
+    struct aws_string *etag;
+    struct aws_string *version_id;
+    struct aws_string *s3_object_last_modified; /* HTTP-date format, e.g. "Wed, 09 Oct 2024 22:28:00 GMT" */
```
DmitriyMusatkin: ISO 8601? i dont think HTTP-date format is common term for htis format

include/aws/s3/private/s3_client_impl.h
```diff
@@ -453,19 +453,30 @@ struct aws_s3_meta_request_resume_token {
...
+    uint64_t continues_downloaded_bytes;        /* contiguous prefix bytes downloaded (no gaps) */
```
DmitriyMusatkin: nit: typo

include/aws/s3/s3_client.h
DmitriyMusatkin: im fine with get only for now, but it should work for put in future as well

source/s3_meta_request.c
```diff
@@ -538,6 +539,45 @@ int aws_s3_meta_request_pause(
     return meta_request->vtable->pause(meta_request, out_resume_token);
 }
 
+int aws_s3_meta_request_pause_async(
+    struct aws_s3_meta_request *meta_request,
+    aws_s3_meta_request_pause_complete_fn *on_complete,
+    void *user_data) {
+    AWS_PRECONDITION(meta_request);
```
DmitriyMusatkin: nit: should we use error precondition for those?

source/s3_meta_request.c
```diff
@@ -2363,6 +2404,10 @@ static void s_s3_meta_request_event_delivery_task(struct aws_task *task, void *a
                 /* 6. Completion tracking */
                 event.u.response_body.bytes_delivered += response_body.len;
                 meta_request->io_threaded_data.num_bytes_delivery_completed += response_body.len;
+                if (error_code == AWS_ERROR_SUCCESS) {
+                    /* Only count bytes the sink actually accepted. */
+                    num_bytes_delivered += response_body.len;
```
DmitriyMusatkin, line 2410: is this different from all the other bytes delivered counters?

source/s3_meta_request.c
```diff
@@ -2583,11 +2655,39 @@ void aws_s3_meta_request_finish_default(struct aws_s3_meta_request *meta_request
     if (meta_request->recv_file) {
         fclose(meta_request->recv_file);
         meta_request->recv_file = NULL;
-        if (finish_result.error_code && meta_request->recv_file_delete_on_failure) {
+
+        bool delete_on_failure = finish_result.error_code && finish_result.error_code != AWS_ERROR_S3_PAUSED &&
```
DmitriyMusatkin, line 2666: should this be an option passed in somehow? if failed, call failure callback an support generating resume token from whats been delivered?

DmitriyMusatkin (approved):

include/aws/s3/s3_client.h
TingDaoK: it already works for both get and put, will update the doc to make it clear

include/aws/s3/private/s3_client_impl.h
TingDaoK: https://www.rfc-editor.org/info/rfc9110/#section-5.6.7 it's the term used in rfc, i'll link the rfc in the doc

DmitriyMusatkin: s3 refers to that as iso 8601. but yeah i think we are talking about the same thing

include/aws/s3/private/s3_meta_request_impl.h
TingDaoK: i'd prefer to leave the naming update to the future me :p. 
As for now, this naming makes most sense. and it's private, we are free to change if it's not making sense in the future.

source/s3_meta_request.c
TingDaoK: i don't see other bytes delivered counters, there are `num_parts_delivery_completed` to track the delivery by parts. Am i missing something?

source/s3_meta_request.c
TingDaoK: the token already generated, this is a check to see if we want to delete the half-downloaded file or not.

DmitriyMusatkin: whats the flow? downloads errors out with partial data on disk? on error callback is called? -> user explicitly generates pause token, which causes error code to switch to paused?  and file is not deleted?

source/s3_meta_request.c
DmitriyMusatkin: ugh you are right, misread the other 2 counter names - bytes allowed and parts delivered

source/s3_meta_request.c
```diff
@@ -2583,11 +2656,39 @@ void aws_s3_meta_request_finish_default(struct aws_s3_meta_request *meta_request
...
+            if (reopened_file != NULL) {
+                uint64_t last_modified_ns = 0;
+                if (aws_file_get_last_modified_epoch(reopened_file, &last_modified_ns) == AWS_OP_SUCCESS) {
+                    token->file_last_modified_epoch_ns = last_modified_ns;
+                }
+                fclose(reopened_file);
+            }
+        }
+
+        if (delete_on_failure) {
```
TingDaoK, line 2682: note: if the file deleted, we should probably just return null as the token.

include/aws/s3/s3_client.h
```diff
@@ -1354,39 +1400,152 @@ struct aws_s3_meta_request_resume_token *aws_s3_meta_request_resume_token_releas
...
- * Only valid for tokens returned from upload operation. For all other operations
- * this will return empty.
+ * Multipart upload id of the paused upload.
+ * Upload tokens only; empty cursor for download tokens.
  */
 AWS_S3_API
 struct aws_byte_cursor aws_s3_meta_request_resume_token_upload_id(
     struct aws_s3_meta_request_resume_token *resume_token);
 
+/* TODO: the download token is currently not supported to resume on (no public constructor exists to
```
azkrishpy, line 1460: Might be worth noting what kind of validation complexity exists because of which we are currently not offering a resume api?

source/s3_meta_request.c
```diff
@@ -538,6 +539,46 @@ int aws_s3_meta_request_pause(
...
+    }
+
+    meta_request->synced_data.async_pause_requested = true;
+    meta_request->synced_data.pause_complete_callback = on_complete;
+    meta_request->synced_data.pause_complete_user_data = user_data;
+
+    /* Stop scheduling new work and cancel in-flight HTTP streams and pending buffer reservations */
+    aws_s3_meta_request_set_fail_synced(meta_request, NULL, AWS_ERROR_S3_PAUSED);
+    aws_s3_meta_request_cancel_cancellable_requests_synced(meta_request, AWS_ERROR_S3_PAUSED);
+    aws_s3_meta_request_cancel_pending_buffer_futures_synced(meta_request, AWS_ERROR_S3_PAUSED);
```
azkrishpy, line 571: we should clear the new priority queue here as well?

source/s3_meta_request.c
azkrishpy, line 2682: I believe already noted but we should not return a token if delete_on_failure is true.

azkrishpy (approved):

source/s3_meta_request.c
TingDaoK: no, that should be handled separately. `aws_s3_meta_request_cancel_pending_buffer_futures_synced` covers it.

include/aws/s3/s3_client.h
TingDaoK: validation is implementation specific, I don't want to suggest the implementation here. The main reason I am pushing this is to allow java sdk owns its own implementation, which I am not 100% agreed on, but it's their current implementation and we don't have a strong reason to change it.

source/s3_meta_request.c
TingDaoK: discussed offline. 
```
     * WARNING: for a file download with recv_file_delete_on_failure, the deletion is
     * respected — the partial file is deleted on error, leaving nothing to resume on,
     * and this callback fires with a NULL token. Do not set recv_file_delete_on_failure
     * if you intend to resume from this callback's token.
     */
    aws_s3_meta_request_pause_complete_fn *on_error_resume_token;
```


alternative we think of:
1. when callback setup, override `recv_file_delete_on_failure` or  reject `recv_file_delete_on_failure`. In the binding implementation, we usually always set the callback even if the end user not hook it up for simplicity. Rely on the callback set or not increases the complication of the binding.
2. refactor callback to let callback decide if delete should happen or not. A bit over complicate, and the token has the file last modified time. And also callback can implement this simply if wanted. 

As the `recv_file_delete_on_failure` is an option to pass-in, let's document it out the side effect should be good enough to let customer provide the expected options here.

source/s3_meta_request.c
azkrishpy: you're right my bad, forgot that we do a force drain.

include/aws/s3/s3_client.h
azkrishpy: you're right, I only suggested it because this comment makes it seem like a trivial addition that we have made a todo item.. java sdk also currently does not support gaps so what is stopping us from creating a resume API that takes the token and creates a ranged GET and change behavior later (to do any validation we want to / support gaps) can be documented.. Anyway I don't mind leaving it as it is either.

PR #651, opened 2026-06-26
Add request_body option to send body from caller-owned memory

yerzhan7:
*Issue #, if available:* N/A

*Description of changes:* 
- Add new option to supply request body for **default** meta request type only: `struct aws_byte_cursor request_body;`
  - Allows users to supply fixed buffer to the client (instead of the client copying into it's own buffer like in body-stream option using `malloc`)
  - Users should only cleanup buffer after meta request is finished


source/s3_client.c
```diff
@@ -1330,11 +1330,43 @@ static struct aws_s3_meta_request *s_s3_client_meta_request_factory_default(
     if (options->send_async_stream != NULL) {
         ++body_source_count;
     }
+    if (options->request_body.ptr == NULL && options->request_body.len > 0) {
```
TingDaoK: We can use `aws_byte_cursor_is_valid` instead.

TingDaoK (approved):
trivial: fix and ship

tests/s3_mock_server_tests.c
```diff
@@ -707,6 +707,167 @@ TEST_CASE(single_upload_unsigned_with_trailer_checksum_mock_server) {
...
+        .throttle_until_attempt = throttle_until_attempt,
+        /* Sentinel: distinct from any real Content-Length (including 0) so the assertion only passes if the hook
+         * actually observed the header on the wire. */
+        .observed_content_length = SIZE_MAX,
+    };
+    tester.user_data = &body_tester;
+
+    struct aws_uri mock_server;
+    ASSERT_SUCCESS(aws_uri_init_parse(&mock_server, allocator, &g_mock_server_uri));
+    const struct aws_byte_cursor *host_cursor = aws_uri_authority(&mock_server);
```
TingDaoK:
```suggestion
   struct aws_byte_cursor host_cursor = *aws_uri_authority(&mock_server);
```
to workaround the const limit 🤦‍♂️

source/s3_client.c
yerzhan7: Fixed, thanks!

tests/s3_mock_server_tests.c
yerzhan7: Thanks, fixed

PR #653, opened 2026-07-08
fix(copy-object): forward request-payer to the source-size HEAD

### LINKED ISSUE #652: CopyObject returns 403 on Requester Pays buckets: the source-size HEAD drops x-amz-request-payer
**Describe the bug**

A client configured for [Requester Pays](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RequesterPaysBuckets.html) (i.e. sending `x-amz-request-payer: requester`) succeeds on `GetObject`, `PutObject`, and `HeadObject` against a Requester Pays bucket, but its `CopyObject` meta-request fails with **403 Forbidden** — even for a same-bucket copy where the caller holds both `s3:GetObject` and `s3:PutObject`.

**Root cause**

Every `CopyObject` meta-request first issues a `HEAD` on the source object to learn its size, before any small-vs-multipart branching (`source/s3_copy_object.c`, `AWS_S3_COPY_OBJECT_REQUEST_TAG_GET_OBJECT_SIZE`). That HEAD is built in `aws_s3_get_source_object_size_message_new()` (`source/s3_request_messages.c`) from a **fresh** `aws_http_message` that inherits **none** of the original request's headers — so `x-amz-request-payer` is never set on it. On a Requester Pays bucket the un-headered HEAD is rejected with 403, failing the whole copy before any other sub-request runs.

This is unique to the size-HEAD. The other copy sub-requests forward headers and are unaffected:
- Small-object `BYPASS` (`CopyObject`) uses `aws_s3_message_util_copy_http_message_no_body_all_headers` → copies all headers.
- `CreateMultipartUpload` / `UploadPartCopy` / `CompleteMultipartUpload` / `AbortMultipartUpload` use exclusion-list copies that do **not** exclude `x-amz-request-payer`.

It is also why plain `GetObject` / `PutObject` / `HeadObject` (and the auto-ranged GET/PUT paths) work — they copy all headers from the template message.

**Reproduction**

1. Cross-account: bucket in account A with Requester Pays enabled; caller in account B granted `s3:GetObject` + `s3:PutObject` via bucket policy.
2. Configure the client to send `x-amz-request-payer: requester`.
3. `GetObject` / `PutObject` / `HeadObject` → **OK**. `CopyObject` (e.g. a same-bucket rename) → **403**.

**Proposed fix**

In `aws_s3_get_source_object_size_message_new()`, forward the request-scoped headers from the original request onto the source-size HEAD, in **both** the `copy_source_uri` branch and the `x-amz-copy-source` fallback branch:
- `x-amz-request-payer` → forwarded verbatim.
- `x-amz-source-expected-bucket-owner` → mapped to `x-amz-expected-bucket-owner` on the HEAD. This is the header that asserts the *source* owner on a `CopyObject`; the plain `x-amz-expected-bucket-owner` asserts the *destination* owner and must **not** be forwarded onto a read of the source object, or legitimate cross-account copies would newly 403.

**Motivating context**

Surfaced by awslabs/s3-connector-for-pytorch#435 (adds a `requester_pays` client option; closes awslabs/s3-connector-for-pytorch#430). With that option enabled, GET/PUT/HEAD work but `S3FileSystem.rename` (implemented as copy + delete) 403s on Requester Pays buckets. The connector reaches aws-c-s3 through mountpoint-s3 → mountpoint-s3-crt-sys, so the fix belongs here.

**Environment**

- aws-c-s3: `main` (reproduced at commit `448ec5e`)
- Also affects the version vendored by `mountpoint-s3-crt-sys` 0.13.1.

A fix plus regression test is ready; PR to follow.

ph-tgs:
*Issue #, if available:* Fixes #652

*Description of changes:*

Every `CopyObject` meta-request first issues a `HEAD` on the source object to determine its size, before any small-vs-multipart branching. That HEAD is built from a fresh `aws_http_message` in `aws_s3_get_source_object_size_message_new()` and inherited none of the original request's headers, so `x-amz-request-payer` was never sent on it. Against a Requester Pays bucket the size-HEAD is rejected with **403** and the whole copy fails — even though the later bypass / multipart-copy sub-requests *do* forward the header.

This change forwards the request-scoped headers from the original request onto the source-size HEAD, in **both** the `copy_source_uri` branch and the `x-amz-copy-source` fallback branch:

- `x-amz-request-payer` — forwarded verbatim (the actual 403 fix).
- `x-amz-source-expected-bucket-owner` — mapped onto the HEAD's `x-amz-expected-bucket-owner`. This is deliberate: on a `CopyObject`, the source owner is asserted via `x-amz-source-expected-bucket-owner`, while the plain `x-amz-expected-bucket-owner` asserts the *destination* owner. Forwarding the destination owner onto a read of the *source* object would newly 403 legitimate cross-account copies, so it is intentionally **not** forwarded.

*Testing:*

- New unit test `test_s3_get_source_object_size_message_new` covering both branches: asserts `x-amz-request-payer` is forwarded, `x-amz-source-expected-bucket-owner` maps onto the HEAD's `x-amz-expected-bucket-owner`, and the destination `x-amz-expected-bucket-owner` is not leaked onto the source HEAD.
- Verified adversarially: with the source change reverted the new test **fails** (`x-amz-request-payer` absent → `AWS_ERROR_HTTP_HEADER_NOT_FOUND`); with the fix it passes. The existing message-builder tests (copy bypass, MPU create / upload-part / complete / abort, ranged-get) all still pass.

### DIFF
--- a/source/s3_request_messages.c
+++ b/source/s3_request_messages.c
@@ -536,6 +536,55 @@ error_clean_up:
 }
 
 static const struct aws_byte_cursor s_slash_char = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("/");
+
+/**
+ * Forward the request-scoped headers from the original CopyObject request onto the source-object HEAD that is used to
+ * size the copy. This HEAD is built from a fresh message, so headers that gate access to the source object are dropped
+ * unless we copy them explicitly:
+ *  - x-amz-request-payer: required, or the HEAD is rejected with 403 on a Requester Pays bucket.
+ *  - x-amz-source-expected-bucket-owner: on a CopyObject this asserts the *source* bucket owner, which maps to
+ *    x-amz-expected-bucket-owner on a HEAD of the source object. (The plain x-amz-expected-bucket-owner on a CopyObject
+ *    asserts the *destination* owner and must NOT be forwarded to the source HEAD.)
+ */
+static int s_s3_copy_source_object_head_forward_headers(
+    struct aws_http_message *base_message,
+    struct aws_http_message *message) {
+
+    struct aws_http_headers *base_headers = aws_http_message_get_headers(base_message);
+    if (base_headers == NULL) {
+        return AWS_OP_SUCCESS;
+    }
+
+    const struct {
+        struct aws_byte_cursor src_name;
+        struct aws_byte_cursor dst_name;
+    } forwarded_headers[] = {
+        {
+            .src_name = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-request-payer"),
+            .dst_name = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-request-payer"),
+        },
+        {
+            .src_name = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-source-expected-bucket-owner"),
+            .dst_name = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("x-amz-expected-bucket-owner"),
+        },
+    };
+
+    for (size_t i = 0; i < AWS_ARRAY_SIZE(forwarded_headers); ++i) {
+        struct aws_byte_cursor value;
+        if (aws_http_headers_get(base_headers, forwarded_headers[i].src_name, &value) == AWS_OP_SUCCESS) {
+            struct aws_http_header header = {
+                .name = forwarded_headers[i].dst_name,
+                .value = value,
+            };
+            if (aws_http_message_add_header(message, header)) {
+                return AWS_OP_ERR;
+            }
+        }
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
 /**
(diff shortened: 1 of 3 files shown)


source/s3_request_messages.c
```diff
@@ -536,6 +536,55 @@ struct aws_http_message *aws_s3_upload_part_copy_message_new(
...
+        struct aws_byte_cursor value;
+        if (aws_http_headers_get(base_headers, forwarded_headers[i].src_name, &value) == AWS_OP_SUCCESS) {
+            struct aws_http_header header = {
+                .name = forwarded_headers[i].dst_name,
+                .value = value,
+            };
+            if (aws_http_message_add_header(message, header)) {
+                return AWS_OP_ERR;
+            }
+        }
```
TingDaoK, line 582:
```suggestion
        }
        /* Avoid leaking the error code. */
        aws_reset_error();
```

TingDaoK (commented):
LGTM, I moved the PR to https://github.com/awslabs/aws-c-s3/pull/654 which applies couple trivial fix.

ph-tgs:
Thanks for the format fix! Closing

PR #655, opened 2026-07-13
fix ordering of read from stream

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Buffer pool can complete reservation futures out of order they were placed in rare circumstances. Couple notable cases:
- if 2 requests complete at the same time and trigger pending cleanup, there a potential race on which pending will arrive first
- tiny last parts of large files can trigger an out of order reservation completion.
Added a general priority queue to every meta request to make sure we serialize processing buffers when input is a stream for put operation even when those buffers arrive out of order


source/s3_client.c
```diff
@@ -2080,9 +2101,29 @@ static void s_on_pool_buffer_reserved(void *user_data) {
     }
     /* END CRITICAL SECTION */
 
-    aws_s3_meta_request_prepare_request(request->meta_request, request, payload->callback, payload->user_data);
-    aws_future_s3_buffer_ticket_release(payload->buffer_future);
-    aws_mem_release(payload->allocator, payload);
+    bool needs_ordered_prepare = request->part_number > 0 &&
+                                 meta_request->type == AWS_S3_META_REQUEST_TYPE_PUT_OBJECT &&
+                                 !meta_request->request_body_parallel_stream;
```
TingDaoK, line 2143: I'd prefer to simplify this to be applied to all meta request. 

It seems not worth the performance gain comparing to the simplicity of the state. especially they all happening from one thread and the pool have to complete the acquisition in the right order. There should not be any potential performance concerns here.

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -305,6 +305,18 @@ struct aws_s3_meta_request {
         /* Total number of bytes that have been attempted to be delivered. (Will equal the sum of succeeded and
          * failed.)*/
         uint64_t num_bytes_delivery_completed;
+
+        /* Priority queue of requests that have acquired their buffer ticket but must wait
+         * for earlier parts to be dispatched for prepare first. This ensures sequential
+         * input stream reads happen in part-number order even when buffer pool reservations
+         * complete out of order. Only used for uploads with sequential streams. */
+        struct aws_priority_queue pending_prepare_queue;
```
TingDaoK: Let's add a note that the queue will keep the entry by value, it's a bit easy to make mistake here. 
We basically deep copy each entry to the queue, which is fine since the entry is just 3 pointer, but different than how I was expecting from what I thought to start with.

TingDaoK (commented):
Overall, I think we should add two documentations:
1. around `aws_s3_buffer_pool_reserve` and the vtable function, we should clearly document that it has to be completed in the order of received, otherwise, a risk of dead lock.
2. Around where we read the stream, ~we should add a double sanity check, we are actually at the part we want to read~ (didn't notice already have that). And doc about the the prepare in this case will always be sequential in the right order.

include/aws/s3/private/s3_meta_request_impl.h
DmitriyMusatkin: done

source/s3_client.c
DmitriyMusatkin: agree. simplified.

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -305,6 +305,20 @@ struct aws_s3_meta_request {
         /* Total number of bytes that have been attempted to be delivered. (Will equal the sum of succeeded and
          * failed.)*/
         uint64_t num_bytes_delivery_completed;
+
+        /* Priority queue of requests that have acquired their buffer ticket but must wait
+         * for earlier parts to be dispatched for prepare first. This ensures sequential
+         * input stream reads happen in part-number order even when buffer pool reservations
+         * complete out of order. Only used for uploads with sequential streams.
```
TingDaoK: trivial: 
```suggestion
         * complete out of order.
```

TingDaoK (approved):
let's also add a regression test for https://github.com/awslabs/aws-c-s3/issues/656

source/s3_meta_request.c
```diff
@@ -2663,6 +2681,20 @@ struct aws_future_bool *aws_s3_meta_request_read_body(
         aws_http_message_get_body_stream(meta_request->initial_request_message);
     AWS_FATAL_ASSERT(synchronous_stream);
 
+    /* Sequential stream reads must happen in increasing part order. If a read
+     * arrives with a part_range_start less than the last one seen, it means parts
+     * were prepared out of order. */
+    if (offset < meta_request->io_threaded_data.next_read_offset) {
```
TingDaoK, line 2698: why we only check this for synchronous_stream? should it be applied to any read without the offset?

source/s3_client.c
```diff
@@ -2080,9 +2102,39 @@ static void s_on_pool_buffer_reserved(void *user_data) {
...
+
+    if (needs_ordered_prepare) {
+        /* Insert into priority queue and drain in order to ensure sequential
+         * stream reads happen in part-number order. */
+        struct aws_s3_pending_prepare_entry entry = {
+            .request = request,
+            .callback = payload->callback,
+            .user_data = payload->user_data,
+        };
+        aws_priority_queue_push(&meta_request->io_threaded_data.pending_put_prepare_queue, &entry);
```
TingDaoK, line 2153: Had a long discussion with Krish, 
Looks like we enforce the order after we reserve the memory from the pool.
So, before we read from the buffer, we need the order to be consistent. 

Which means, nothing async or retry should happen in between. It feels a bit fragile to have this assumption? 

For now, looks like nothing can result in retrying in between and all we do is scheduling the task to the meta request thread after we reserved the memory. So, we should be good at this time. 

But, ideally, the enforcement of the order should happens right before the read. Or moved as closer to the read as possible?

azkrishpy:
We exit the callback if we set an error (which we do when we cancel a request) as seen here.
https://github.com/awslabs/aws-c-s3/blob/448ec5e49eb181687f9af39b0d3b83a563b15039/source/s3_client.c#L2060

With that in mind, if we think of a scenario where 2 gets queued before 1 and we cancel the request.. the pending queue would never get drained. we should add a cleanup of pending_prepare_queue to cancel as well? (or some other way to cleaner way if that exists)

https://github.com/awslabs/aws-c-s3/blob/448ec5e49eb181687f9af39b0d3b83a563b15039/source/s3_meta_request.c#L504

How can we test for this? ** melting brain **

tests/CMakeLists.txt
```diff
@@ -373,6 +373,10 @@ add_net_test_case(test_s3_upload_review_rejection)
 add_net_test_case(test_s3_upload_review_checksum_location_none)
 add_net_test_case(test_s3_upload_review_checksum_location_none_async)
 add_net_test_case(test_s3_upload_review_checksum_location_none_async_noop_part)
+add_net_test_case(test_s3_upload_in_order_review)
```
sbSteveK, line 376: Trivial: looks like the tests are added twice?

DmitriyMusatkin: fixed in next rev.

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -305,6 +305,20 @@ struct aws_s3_meta_request {
...
+        /* Priority queue of requests that have acquired their buffer ticket but must wait
+         * for earlier parts to be dispatched for prepare first. This ensures sequential
+         * input stream reads happen in part-number order even when buffer pool reservations
+         * complete out of order.
+         * Note: each element is the pending request, callback to the next state and some user data
+         * (aws_s3_pending_prepare_entry). */
+        struct aws_priority_queue pending_put_prepare_queue;
+
+        /* Next part number that should be dispatched from pending_put_prepare_queue. */
+        uint32_t next_put_part_to_prepare;
```
sbSteveK, line 318: trivial: Any reason not to go with `uint64_t`? Recently put in a counter for MQTT manual pubacks recently using 32 and it was suggested there to use 64 so we never get anywhere near the full range despite it being nearly impossible that a client would run long enough to.

sbSteveK: Actually, looking down, assuming this is reset and reused per download as opposed to any long lifetime.

azkrishpy: part numbers are limited to 10k. we are already well above that number with `uint32_t`

source/s3_client.c
```diff
@@ -2036,6 +2036,46 @@ static void s_s3_prepare_acquire_mem_callback_and_destroy(
...
+static void s_drain_pending_put_prepare_queue(struct aws_s3_meta_request *meta_request) {
+    struct aws_priority_queue *queue = &meta_request->io_threaded_data.pending_put_prepare_queue;
+
+    while (aws_priority_queue_size(queue) > 0) {
+
+        struct aws_s3_pending_prepare_entry *top = NULL;
+        aws_priority_queue_top(queue, (void **)&top);
+
+        if (top->request->part_number != meta_request->io_threaded_data.next_put_part_to_prepare) {
+            break;
```
sbSteveK, line 2058: Trivial: Do we want Debug or Trace level logging that this has been hit? Maybe unnecessary but it'd let us know when we are seeing this behavior where things are out of order even though now it's handled.

include/aws/s3/private/s3_meta_request_impl.h
DmitriyMusatkin: yeah, its per meta req. current ceiling is 10k, but it unlikely to ever need anything closer to 32 bits. afaik we use 32 bits for part number in other places as well, so keeping that

source/s3_client.c
DmitriyMusatkin: added it under trace. dont want to spam logs too much

source/s3_meta_request.c
```diff
@@ -671,6 +680,9 @@ static void s_s3_meta_request_destroy(void *user_data) {
     AWS_ASSERT(aws_array_list_length(&meta_request->io_threaded_data.event_delivery_array) == 0);
     aws_array_list_clean_up(&meta_request->io_threaded_data.event_delivery_array);
 
+    AWS_ASSERT(aws_priority_queue_size(&meta_request->io_threaded_data.pending_put_prepare_queue) == 0);
+    aws_priority_queue_clean_up(&meta_request->io_threaded_data.pending_put_prepare_queue);
```
azkrishpy, line 684: This should have a reusable helper with cancel but looks like pause has a separate place in upload doing the cancelling of pending queues. We might have to add the check there too?

Ref:
https://github.com/awslabs/aws-c-s3/blob/448ec5e49eb181687f9af39b0d3b83a563b15039/source/s3_auto_ranged_put.c#L1862

azkrishpy (approved):

source/s3_meta_request.c
azkrishpy: ignore this

PR #657, opened 2026-07-16
fix: restore old enum

azkrishpy:
*Issue #, if available:*

*Description of changes:*
Renaming an enum is a breaking change as customers using the old enum would encounter compilation failures against the new version. In this change we attempt to restore the old enum 


azkrishpy:
done better on #658

PR #661, opened 2026-07-30
fix resume token build

TingDaoK:
*Issue #, if available:*
- forgot to check for null

### DIFF
--- a/include/aws/s3/private/s3_meta_request_impl.h
+++ b/include/aws/s3/private/s3_meta_request_impl.h
@@ -233,7 +233,7 @@ struct aws_s3_meta_request {
         /* Total number of response-body bytes successfully delivered to the caller's sink
          * (file or body callback), in order with no gaps. Used to build the download resume
          * token on pause/error.
-         * TODO: delivery is strictly sequential today, so this is both the contiguous prefix and the total; a future
+         * TODO: delivery is strictly sequential today, so this is both the continues prefix and the total; a future
          * parallel-write delivery path will need to track the two separately. */
         uint64_t num_bytes_delivered;
 
--- a/include/aws/s3/private/s3_meta_request_impl.h
+++ b/include/aws/s3/private/s3_meta_request_impl.h
@@ -233,7 +233,7 @@ struct aws_s3_meta_request {
         /* Total number of response-body bytes successfully delivered to the caller's sink
          * (file or body callback), in order with no gaps. Used to build the download resume
          * token on pause/error.
-         * TODO: delivery is strictly sequential today, so this is both the continues prefix and the total; a future
+         * TODO: delivery is strictly sequential today, so this is both the contiguous prefix and the total; a future
          * parallel-write delivery path will need to track the two separately. */
         uint64_t num_bytes_delivered;
 
--- a/include/aws/s3/s3_client.h
+++ b/include/aws/s3/s3_client.h
@@ -1521,7 +1521,7 @@ AWS_S3_API
 uint64_t aws_s3_meta_request_resume_token_object_range_end(const struct aws_s3_meta_request_resume_token *resume_token);
 
 /*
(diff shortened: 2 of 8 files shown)


source/s3_auto_ranged_get.c
```diff
@@ -1144,7 +1144,7 @@ static struct aws_s3_meta_request_resume_token *s_build_download_resume_token_sy
             aws_string_new_from_string(meta_request->allocator, auto_ranged_get->s3_object_last_modified);
     }
 
-    /* Delivery is strictly sequential today, so the delivered bytes are both the contiguous
+    /* Delivery is strictly sequential today, so the delivered bytes are both the continues
```
azkrishpy:
```suggestion
    /* Delivery is strictly sequential today, so the delivered bytes are both the contiguous
```

include/aws/s3/s3_client.h
```diff
@@ -1521,7 +1521,7 @@ AWS_S3_API
 uint64_t aws_s3_meta_request_resume_token_object_range_end(const struct aws_s3_meta_request_resume_token *resume_token);
 
 /*
- * Number of bytes downloaded contiguously from the start of the range, with no
+ * Number of bytes downloaded continuesly from the start of the range, with no
```
azkrishpy:
```suggestion
 * Number of bytes downloaded continuously from the start of the range, with no
```

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -233,7 +233,7 @@ struct aws_s3_meta_request {
         /* Total number of response-body bytes successfully delivered to the caller's sink
          * (file or body callback), in order with no gaps. Used to build the download resume
          * token on pause/error.
-         * TODO: delivery is strictly sequential today, so this is both the contiguous prefix and the total; a future
+         * TODO: delivery is strictly sequential today, so this is both the continues prefix and the total; a future
```
azkrishpy:
```suggestion
         * TODO: delivery is strictly sequential today, so this is both the contiguous prefix and the total; a future
```

source/s3_auto_ranged_put.c
```diff
@@ -328,7 +328,9 @@ static int s_try_init_resume_state_from_persisted_data(
     auto_ranged_put->synced_data.num_parts_noop = 0;
     auto_ranged_put->synced_data.create_multipart_upload_sent = true;
     auto_ranged_put->synced_data.create_multipart_upload_completed = true;
-    auto_ranged_put->upload_id = aws_string_clone_or_reuse(allocator, resume_token->multipart_upload_id);
+    if (resume_token->multipart_upload_id) {
+        auto_ranged_put->upload_id = aws_string_new_from_string(allocator, resume_token->multipart_upload_id);
```
azkrishpy, line 332: this is also clone_or_reuse I think?

source/s3_auto_ranged_put.c
```diff
@@ -1916,7 +1920,9 @@ static int s_s3_auto_ranged_put_pause(
      * - complete MPU started - return success, generate token and try to cancel
      *   complete MPU
      */
-    *out_resume_token = s_s3_auto_ranged_put_build_resume_token_synced(meta_request);
+    if (auto_ranged_put->synced_data.create_multipart_upload_completed) {
+        *out_resume_token = s_s3_auto_ranged_put_build_resume_token_synced(meta_request);
```
azkrishpy: we are already checking this within `s_s3_auto_ranged_put_build_resume_token_synced`

azkrishpy (approved):
minor nits only

source/s3_auto_ranged_put.c
TingDaoK: i think they should be the same. I'll keep the new_from, since we should just keep it around always. Don't want to risk the resume path.

PR #670, opened 2026-09-01
Sub 10Gbps S3 Throughput Handling

sbSteveK:
We previously only had default memory pool sizing for the s3 clients at greater than 10 Gbps throughput and all slower throughputs shared a native memory pool sized to 2GiB. This PR adds more pool sizes for throughputs below 10 Gbps that size the pool to 512 MiB or 256 MiB. This retains performance at those throughput speeds while reducing the memory resource requirements.

This PR also adds a check for EC2 instances which will right-size the throughput value resulting in the appropriate native memory pool size. We currently only apply this check for EC2 instance families that have less than 10 Gbps NIC capability. We may open up the right-sizing to higher performance EC2 instances and non-EC2 in the future.


source/s3_client.c
```diff
@@ -96,7 +96,8 @@ static const uint32_t s_endpoints_cleanup_time_offset_in_s = 5;
 /**
  * The environment variable name for memory limit control.
  */
-static const char *s_memory_limit_env_var = "AWS_CRT_S3_MEMORY_LIMIT_IN_GIB";
+static const char *s_memory_limit_gib_env_var = "AWS_CRT_S3_MEMORY_LIMIT_IN_GIB";
+static const char *s_memory_limit_bytes_env_var = "AWS_CRT_S3_MEMORY_LIMIT_IN_BYTES";
```
TingDaoK, line 100: add this to the README. 

Also, do we really want it in bytes? Can we do MB?

source/s3_default_buffer_pool.c
```diff
@@ -210,6 +210,17 @@ struct aws_s3_buffer_pool *aws_s3_default_buffer_pool_new(
             "if its not sufficient to transfer data within the maximum number of parts");
     }
 
+    if (chunk_size > config.memory_limit) {
```
TingDaoK, line 213: the config has a max_part_size, which is 1/2 of the memory pool size. 
```
    /* default max part size is the smallest of either half of mem limit or default max part size */
    size_t max_part_size = aws_min_size(mem_limit / 2, (size_t)g_default_max_part_size);
    if (client_config->max_part_size != 0) {
        if (client_config->max_part_size > SIZE_MAX) {
            max_part_size = SIZE_MAX;
        } else {
            max_part_size = (size_t)client_config->max_part_size;
        }
    }
```

we probably should check against that

source/s3_default_buffer_pool.c
```diff
@@ -192,11 +192,11 @@ struct aws_s3_buffer_pool *aws_s3_default_buffer_pool_new(
 
     size_t chunk_size = config.part_size;
 
-    if (config.memory_limit < GB_TO_BYTES(1)) {
+    if (config.memory_limit < MB_TO_BYTES(256)) {
```
TingDaoK, line 195: what if the memory limit in the config is some random number, eg: 256MiB +1234 Bytes. Does it make more sense to make the limitation align with page size or align with MiB?

source/s3_client.c
DmitriyMusatkin: +1. i dont see us reasonably supporting under megabyte changes

source/s3_platform_info.c
```diff
@@ -18,128 +18,271 @@
...
+    {"g5",                                               100.0,    192,        false,       NULL},
+    {"g4dn",                                             100.0,     96,        false,       NULL},
+    {"inf2",                                             100.0,    192,        false,       NULL},
+    {"inf1",                                             100.0,     96,        false,       NULL},
+
+    /* === Burstable === */
+    {"t4g",                                                5.0,      8,        false,       NULL},
+    {"t3a",                                                5.0,      8,        false,       NULL},
+    {"t3",                                                 5.0,      8,        false,       NULL},
+    {"t2",                                                 1.0,      8,        false,       NULL},
```
DmitriyMusatkin, line 184: baseline is max on instance right? and based on num cpu it can be lower? in future we should think if we want to record both min and max or some granularity (not asking to do it now).

source/s3_platform_info.c
```diff
@@ -18,128 +18,271 @@
...
 
-/****** End c5n.9large *****/
-
-/***** Begin p4d.24xlarge and p4de.24xlarge ****/
-static struct aws_s3_platform_info s_p4d_platform_info = {
-    .instance_type = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("p4d.24xlarge"),
-    .max_throughput_gbps = 400u,
-    .has_recommended_configuration = true,
+/* clang-format off */
+static const struct aws_s3_ec2_instance_type_info s_ec2_instance_type_table[] = {
```
DmitriyMusatkin, line 50: we should consider some sanity checking against smithy model here https://raw.githubusercontent.com/aws/aws-sdk-cpp/refs/heads/main/tools/code-generation/api-descriptions/ec2-2016-11-15.normal.json

source/s3_client.c
sbSteveK: swapped from Bytes to MB.

source/s3_default_buffer_pool.c
sbSteveK: Added check against the max part size

source/s3_default_buffer_pool.c
sbSteveK: Unsure that we want to be making adjustments if a customer is intentionally providing some weird number. I've added a warn log about it. Not something I'm firm on so if you think we should auto-align them to the closest MB down, I can go ahead and do that instead along with the warn log.

source/s3_platform_info.c
sbSteveK: Yup. The range we are using for the 256MiB and 512MiB pool sizes make it so using the max has no difference vs. instances with lower actual.

In the future, if and when we get EC2 to put their actual throughput per instance in IMDS, I'm gonna propose we check for the EC2 files to confirm we're on an EC2 and then just go straight to IMDS to get accurate and up-to-date values instead of having a table at all.

source/s3_platform_info.c
sbSteveK: It appears as though there are lower end (no longer in use) families that we exclude and also some newer families we don't have. On the newer end, I think it's fine for now and we should fully update our list when we decide to open up auto-defaulting for higher perf EC2 but ideally, we won't need to if we can use IMDS. This is the exact case we can present to EC2 about the need for this info to be in IMDS so we can support instances we don't know about without updating hard-coded tables.

source/s3_client.c
```diff
@@ -96,7 +96,8 @@ static const uint32_t s_endpoints_cleanup_time_offset_in_s = 5;
 /**
  * The environment variable name for memory limit control.
  */
-static const char *s_memory_limit_env_var = "AWS_CRT_S3_MEMORY_LIMIT_IN_GIB";
+static const char *s_memory_limit_gib_env_var = "AWS_CRT_S3_MEMORY_LIMIT_IN_GIB";
+static const char *s_memory_limit_mb_env_var = "AWS_CRT_S3_MEMORY_LIMIT_IN_MB";
```
TingDaoK, line 101: Add to README?
And be clear about which one takes precedence?

source/s3_client.c
```diff
@@ -382,22 +419,89 @@ struct aws_s3_client *aws_s3_client_new(
...
+        if (effective_throughput == 0.0) {
+            const struct aws_s3_platform_info *detected_platform = aws_s3_get_current_platform_info();
+            /*
+             * For now, we will only right-size < 10 Gbps to minimize change. We can address right-sizing and
+             * applying proper EC2 throughput values in the future.
+             */
+            if (detected_platform != NULL && detected_platform->max_throughput_gbps > 0.0 &&
+                detected_platform->max_throughput_gbps < 10.0) {
+                effective_throughput = detected_platform->max_throughput_gbps;
+            }
```
TingDaoK, line 446: We should also check if people explicitly set a throughput less than the bandwidth the ec2 allows?

source/s3_client.c
```diff
@@ -382,22 +419,89 @@ struct aws_s3_client *aws_s3_client_new(
...
+         * effective throughput remains 0.0, which maps to the 2 GiB
+         * default in the tier table below.
+         */
+        double effective_throughput = client_config->throughput_target_gbps;
+        if (effective_throughput == 0.0) {
+            const struct aws_s3_platform_info *detected_platform = aws_s3_get_current_platform_info();
+            /*
+             * For now, we will only right-size < 10 Gbps to minimize change. We can address right-sizing and
+             * applying proper EC2 throughput values in the future.
+             */
```
TingDaoK, line 442: this seems to be a hacky magic that only applies to <10Gbps. 
I am fine with it, just feel like we may forgot and bite us in the future.

sbSteveK: My main issue with this is that anything that's NOT an EC2 instance will use our pre-existing defaults but this is a bigger problem than the scope of what this PR is addressing. The saving grace here is that if a customer on a non-EC2 sets their throughput value accurately and it's below the 10 Gbps threshold, they'll now get a right-sized memory pool instead of the one they get by default.

source/s3_client.c
sbSteveK: I prefer not to set an upper-cap on this until we have known values via IMDS. Until then, things might shift and I'd rather allow someone to set it to higher than what we think is the upper limit. I can be convinced to set hard caps but that's my POV.

source/s3_client.c
sbSteveK: added

source/s3_client.c
TingDaoK: no, i mean if someone sets lower than the ec2 allows. we should respect that.

TingDaoK: oh, ignore me. Looks like you have a `if (effective_throughput == 0.0)` that guards against that,.

source/s3_platform_info.c
TingDaoK: i am not a fan of this hard-coded list. If there is a TODO that we require IMDS update, let's leave it as TODO somewhere and make sure we keep it on track and follow up as needed.

source/s3_platform_info.c
```diff
@@ -18,128 +18,271 @@
...
+    {"c6i",                                               50.0,    128,        false,       NULL},
+    {"c6gn",                                             100.0,     64,        false,       NULL},
+    {"c6gd",                                              25.0,     64,        false,       NULL},
+    {"c6g",                                               25.0,     64,        false,       NULL},
+    {"c6a",                                               50.0,    192,        false,       NULL},
+    {"c5n",                                              100.0,     72,        false,       NULL},
+    {"c5ad",                                              20.0,     96,        false,       NULL},
+    {"c5a",                                               20.0,     96,        false,       NULL},
+    {"c5d",                                               25.0,     96,        false,       NULL},
+    {"c5",                                                25.0,     96,        false,       NULL},
```
TingDaoK, line 113: why they are all `false` for `has_recommended_configuration`?

sbSteveK: That's a remnant from the previously hard-coded ec2 instances. That appears to be used downstream in aws-crt-python to maybe CLI or elsewhere to let them know whether there are recommended configurations available for that specific instance. we kept them here with it true on the specific instances they were true before to not make any potential breaking or behavioral changes.

I'm pretty sure it was used to determine whether or not CLI should use the s3 crt client and to set throughput. There's also a test in CLI that requires we match their hardcoded list of expected EC2 instances that are returned from their API call to check for EC2 instances that have recommended configurations. It's all bad and feels fragile and should be removed once we have IMDS support for bandwidth.

source/s3_platform_info.c
sbSteveK: Yea, I hate it too. I'm already in conversation with EC2 and will put something together to review the business need for them to expose explicit bandwidth and hopefully disk throughput per EC2 instance via IMDS after this release. There is a TODO I left regarding this below 

* TODO: Once we are able to auto-detect NICs and add them, default values
     * should be updated with maximum ENA network bandwidth for these instances. */

source/s3_platform_info.c
TingDaoK: does it make more sense to set the new added ones to be true for this? 
It doesn't make sense to me that we have the type and the settings, but `has_recommended_configuration` is false.

TingDaoK (approved):
no blocker from me.

PR #671, opened 2026-09-04
Retry on 502 504 errors

sbSteveK:
S3 considers 502 and 504 retryable errors. We updated the error to be returned on them as `AWS_ERROR_S3_INTERNAL_ERROR` which will allow them to be retried.

The underlying HTTP error that it represents will still be logged. As this specific error path is only dealt with internally for our retry logic, there is no need to create new S3 specific Gateway related error codes to represent the HTTP error codes we are handling.

Some testing helpers and tests to make sure that when we receive those specific errors, our client retries.


azkrishpy (commented):
is this an S3 specific requirement? We do have [a function](https://github.com/awslabs/aws-c-http/blob/2b563f8a7bd67a902a8b558bb44113748045877c/source/http.c#L567) to change this at the http level if we intend to allow 502 and 504 as a standard http code to be retried.

sbSteveK:
> is this an S3 specific requirement? We do have [a function](https://github.com/awslabs/aws-c-http/blob/2b563f8a7bd67a902a8b558bb44113748045877c/source/http.c#L567) to change this at the http level if we intend to allow 502 and 504 as a standard http code to be retried.

This is specific to S3. We probably don't want to change retry behavior elsewhere.

TingDaoK (approved):

DmitriyMusatkin (approved):

PR #672, opened 2026-09-09
Add retry_config to aws_s3_client_config

sbSteveK:
Add a `retry_config` struct to `aws_s3_client_config` that allows the setting of the following retry options:
max_retries
backoff_scale_factor_ms
max_backoff_secs
jitter_mode
initial_bucket_capacity

If set, they will override the default values that the s3 client uses. A provided `aws_retry_strategy` will take precedence over any options set in this struct as the aws_retry_strategy comes fully constructed and can't be modified.

Test that the setting of `max_retries` via the `retry_config` is honored.


source/s3_client.c
```diff
@@ -798,13 +798,24 @@ struct aws_s3_client *aws_s3_client_new(
         aws_retry_strategy_acquire(client_config->retry_strategy);
         client->retry_strategy = client_config->retry_strategy;
     } else {
+        /* max_retries requires explicit S3 default because passing 0 to aws-c-io's
+         * standard retry strategy would use its own default of 3, not the S3 default of 5.
+         * The other fields use 0 = "use aws-c-io defaults" which match the S3 defaults. */
+        uint32_t max_retries = client_config->retry_config.max_retries > 0
```
azkrishpy, line 804: should we guardrail max_retries before setting it? 
we seem to [return null after raising an error](https://github.com/awslabs/aws-c-io/blob/c2d222b73000ddcc5aeed837f4434560a47bad47/source/exponential_backoff_retry_strategy.c#L337) when the number exceeds 63. But the raised error is not read in the client meaningfully based on my read of things.

tests/s3_retry_tests.c
```diff
@@ -119,6 +119,87 @@ static int s_test_s3_client_acquire_connection_fail(struct aws_allocator *alloca
     return 0;
 }
 
+/* Test that retry_config.max_retries is respected when set.
+ * Uses the fail_first helper (one connection failure, then success).
+ * With max_retries = 1, the single failure should be retried and the request should succeed. */
+AWS_TEST_CASE(test_s3_client_retry_config_max_retries, s_test_s3_client_retry_config_max_retries)
+static int s_test_s3_client_retry_config_max_retries(struct aws_allocator *allocator, void *ctx) {
```
azkrishpy, line 127: this was something smart that AI found so sharing it here. Both tests check that retry config works but neither tests whether the actual number `max_retries` is working. The first test would pass for any value that is not zero (we cannot set zero). The second test would fail for any value of `max_retries`. A better test might be to set max_retries to 1 and check that fail first passes and failing twice fails the test?

source/s3_client.c
sbSteveK: Added a null check on the resulting retry strategy and we error and don't return a client if its creation failed.

tests/s3_retry_tests.c
sbSteveK: Updated the test

azkrishpy (approved):

PR #673, opened 2026-09-10
feat: add more metrics to S3 metrics object

azkrishpy:
*Issue #, if available:*

*Description of changes:*
Adds request-scoped telemetry to aws_s3_request_metrics:

- is_https — whether the request was made over TLS, so Java can build a full scheme-qualified endpoint URL
- http_manager_metrics — snapshot of the endpoint's connection manager (available/pending/leased concurrency), taken right before this request asks for a connection, so it reflects contention this request found rather than its own acquire
- Exposes aws_s3_client_get_max_active_connections publicly (was already computed correctly, just lived in a private header) — the connection concurrency cap a client/meta-request is subject to


include/aws/s3/s3_client.h
```diff
@@ -1830,6 +1836,22 @@ void aws_s3_request_metrics_get_host_address(
...
 
+/**
+ * Get whether the request was made over TLS (https) or plaintext (http). This will always be available.
+ */
+AWS_S3_API
+bool aws_s3_request_metrics_get_is_https(const struct aws_s3_request_metrics *metrics);
+
+/**
+ * Get a snapshot of the endpoint's HTTP connection manager metrics, taken right before this request
+ * asks for a connection. This reflects the manager's overall state at that instant, not just this
```
TingDaoK, line 1858: >taken right before this request asks for a connection

This is a bit counter intuitive to me, but as long as it gets propagated correctly.

TingDaoK (approved):

include/aws/s3/s3_client.h
azkrishpy: I agree this is a bit confusing. Confirmed with the sdk and to achieve parity with their existing client, the right behavior would be to snapshot before connection acquisition.

PR #674, opened 2026-09-10
Add S3 Metrics

sbSteveK:
Add the following metrics to the user agent:
    AWS_S3_METRIC_CUSTOM_PART_SIZE - non-default part size configured 
    AWS_S3_METRIC_CUSTOM_THROUGHPUT - non-default throughput target configured
    AWS_S3_METRIC_CUSTOM_MEMORY_LIMIT - non-default memory pool size configured 
    AWS_S3_METRIC_ON_EC2 - running on EC2 instance
    AWS_S3_METRIC_FILE_PATH - request used send_filepath or recv_filepath

Add a test for the adding of metrics.


TingDaoK (commented):
i'd prefer we rename metrics to featureid to be more specific.
We already have `aws_s3_request_metrics`, it will be confusing to add another metrics where we can avoid

source/s3_client.c
```diff
@@ -551,6 +561,40 @@ struct aws_s3_client *aws_s3_client_new(
...
+    /* Initialize client-level business metrics flags for User-Agent m/ section.
+     * These are set AFTER all defaults are resolved so we can compare the effective
+     * values against what the system would have chosen automatically. The metric
+     * tracks "is the customer using non-default behavior?" not "did the customer
+     * touch the API?" */
+    client->business_metrics = AWS_S3_METRIC_CRT_CLIENT; /* always set */
+
+    if (client_config->part_size != 0 && part_size != (size_t)g_default_part_size_fallback) {
+        /* Caller set part_size AND it differs from the 8 MiB default */
+        client->business_metrics |= AWS_S3_METRIC_CUSTOM_PART_SIZE;
```
TingDaoK: meta request can override the part size

source/s3_meta_request.c
```diff
@@ -1104,7 +1104,18 @@ static void s_s3_meta_request_on_request_prepared(void *user_data) {
...
+     *
+     * File upload detection: request_body_parallel_stream is only set when the caller
+     * provided send_filepath in the meta request options (see s3_meta_request.c init).
+     * It is NOT set for streaming uploads (send_async_stream, send_using_async_writes,
+     * or body_stream on the initial request message). This makes it a reliable indicator
+     * that the upload is file-based rather than stream-based. */
+    uint32_t request_metrics = meta_request->client->business_metrics;
+    if (meta_request->request_body_parallel_stream != NULL) {
+        request_metrics |= AWS_S3_METRIC_FILE_UPLOAD;
+    }
```
TingDaoK: can we move this to the meta request new and check against that `recv_filepath`/`send_filepath` is set or not? 

Also, I think it's worth to be generic on filepath vs streaming for both upload and download.

include/aws/s3/private/s3_util.h
```diff
@@ -246,7 +246,24 @@ AWS_S3_API
 int aws_last_error_or_unknown(void);
 
 AWS_S3_API
-void aws_s3_add_user_agent_header(struct aws_allocator *allocator, struct aws_http_message *message);
+void aws_s3_add_user_agent_header(
+    struct aws_allocator *allocator,
+    struct aws_http_message *message,
+    uint32_t business_metrics);
```
azkrishpy: do we need to change the api for this? I see that it is a private API but seems like it should be doable without the API change?

include/aws/s3/private/s3_util.h
```diff
@@ -246,7 +246,24 @@ AWS_S3_API
...
+
+/**
+ * Business metrics flags for the User-Agent `m/` section (UA 2.1 SEP).
+ * These are boolean flags tracked via bitmask on the client and meta request.
+ * Feature IDs are registered in AwsDrSeps/FEATURES.md.
+ */
+enum aws_s3_business_metrics_flag {
```
DmitriyMusatkin: not alternatively you can do something along the lines of 
struct bussiness_metrics {
    uint32_t flag1 : 1;
    uint32_t flag2 : 1;
};

uint32_t will ensure that they are packed into a single uint32_t

include/aws/s3/private/s3_util.h
azkrishpy: I got the flow wrong, ignore this comment

source/s3_meta_request.c
```diff
@@ -1104,7 +1104,18 @@ static void s_s3_meta_request_on_request_prepared(void *user_data) {
...
     }
 
-    aws_s3_add_user_agent_header(meta_request->allocator, request->send_data.message);
+    /* Merge client-level and per-request business metrics for User-Agent m/ section.
+     *
+     * File upload detection: request_body_parallel_stream is only set when the caller
+     * provided send_filepath in the meta request options (see s3_meta_request.c init).
+     * It is NOT set for streaming uploads (send_async_stream, send_using_async_writes,
+     * or body_stream on the initial request message). This makes it a reliable indicator
+     * that the upload is file-based rather than stream-based. */
```
DmitriyMusatkin: are we sure on that? equating request_body_parallel_stream to file upload seems brittle

include/aws/s3/private/s3_util.h
sbSteveK: We could potentially add a new API but this is a private header and I think we're good with touching those.

source/s3_client.c
sbSteveK: Moved per-request into aws_s3_meta_request_init_base. We'll check part size against what's setup in the options.

source/s3_meta_request.c
sbSteveK: Moved setting to aws_s3_meta_request_init_base. Feels like it's a much better home.

sbSteveK:
> i'd prefer we rename metrics to featureid to be more specific. We already have `aws_s3_request_metrics`, it will be confusing to add another metrics where we can avoid

Sounds good. Changed everywhere to feature id

include/aws/s3/private/s3_util.h
sbSteveK: bitmask lets the emission stay a single table plus loop and lets client flags merge with one `or`. A bitfield struct would require `if` per field in aws_s3_add_user_agent_header and a per-field merge. I don't think one is significantly better than the other but bitmask is already in place. Will change if you want to but if you don't feel strongly I'd rather just keep as is for now.

TingDaoK (approved):

DmitriyMusatkin (approved):

PR #676, opened 2026-09-10
Include the real version in the user agent

TingDaoK:
*Issue #, if available:*
- address the TODO the include the real version in the user agent. 

*Description of changes:*
- Cmake will bake the macro, in case where we don't use cmake, just leave it unknown. 


sbSteveK (approved):

DmitriyMusatkin (approved):

TingDaoK:
included in https://github.com/awslabs/aws-c-s3/pull/674

PR #678, opened 2026-09-11
Checksum validation handling cleanup

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Cleaning up some of the existing issues before i mess with it more. General cleanup along with better error handling and fixed several issues:
- if meta req had range that started on the part boundary, the first get that discovers the size might include a checksum that is for the part, but is assumed to be full object. this later on will lead to a checksum validation failure because part checksum will not match full object.
- if meta req had an end only range (i.e. no start) then head is done. full object checksum from that head could be remembered even though only range of file is being downloaded. and then we get checksum failure similar to first bug
- part gets are kicked to default, but checksum validation for them is not correctly propagated (more of reporting bug of validation being done, but not being marked as validated)


source/s3_auto_ranged_get.c
```diff
@@ -906,7 +968,26 @@ static void s_s3_auto_ranged_get_request_finished(
...
+                /* This response's body was already compared against a checksum header of its own, and that body is
+                 * every byte the meta request delivers, so there is nothing left for a meta-request-level sum to
+                 * check. update() reports the outcome through the all-parts-validated promotion instead. */
+                AWS_LOGF_DEBUG(
+                    AWS_LS_S3_META_REQUEST,
+                    "id=%p: Discovery response covered the whole download and was already validated against its own "
+                    "checksum.",
+                    (void *)meta_request);
+            } else if (
+                aws_s3_check_headers_for_checksum(
```
TingDaoK, line 990: do we still want the following in this function?
```
    if (meta_request_level && aws_http_headers_has(headers, g_mp_parts_count_header_name)) {
        /* g_mp_parts_count_header_name indicates it's a object was uploaded as a
         * multipart upload. So, the checksum should not be applied to the meta request level.
         * But we we want to check it for the request level. */
        *out_checksum = NULL;
        return AWS_OP_SUCCESS;
    }
```

TingDaoK (approved):

source/s3_auto_ranged_get.c
DmitriyMusatkin: right, that is no longer needed after all the changes. removed

TingDaoK (approved):

PR #679, opened 2026-09-11
Parallel out of order write to file

TingDaoK:
Downloads to a `recv_filepath` currently write every part sequentially through a single
`FILE *` on the meta request's `io_event_loop`. That caps a download at one writer's
throughput no matter how fast the disk is, and a slow write blocks the same loop that
delivers bodies, runs user callbacks, and prepares requests.

This PR lets received parts be written to the file **in parallel, as they arrive**, on a
thread pool dedicated to file I/O.

### Public API

Two new fields on `aws_s3_client_config`

- `enum aws_tribool out_of_order_delivery` — whether a download may delivery parts
  as they arrive rather than in object order. `AWS_TRIBOOL_UNSET` lets the client decide,
  which currently means out of order wherever the destination allows it. Ignored when a
  response carries a whole-object checksum that can only be verified by hashing the body in
  order; those requests deliver in order and log a warning.
- `uint16_t num_file_io_threads` — size of the file I/O thread pool. Also bounds how many
  parts a download can have in flight to the disk at once. Defaults to the number of event
  loops in the client bootstrap's ELG.

The trade-off `out_of_order_delivery` exposes is what a partial file contains. Ordered
delivery leaves a valid prefix. Out-of-order delivery can leave gaps, so a partial file is
only meaningful together with the download resume token's
`continuous_downloaded_bytes`.

### Dedicated file I/O event loop group

New `aws_s3_client::file_io_elg`, separate from `body_streaming_elg`, so a blocking write
costs only disk throughput instead of also stalling body delivery and request preparation.
Writes are handed out round-robin via `next_write_loop_index`.

The ordered path still writes on the meta request's `io_event_loop` — it is serialized by
definition, so a pool buys it nothing, and moving it would mean moving all of that meta
request's event delivery with it. A TODO notes the upload path's parallel file reads should
also move here rather than borrowing `body_streaming_elg`.

### Per-worker descriptors

`FILE *recv_file` is gone from the meta request. The file is created/truncated once at init
through a short-lived stdio handle; received bytes go only through positional writes on
descriptors held in `struct aws_s3_recv_file_fds`:

- `recv_file_write_fd_slots[]` — one slot per write worker, indexed by its `file_io_elg`
  loop index, opened lazily on that worker's first body. Each slot is owned exclusively by
  one thread, so two workers never share a `struct file` and no slot needs a lock.
- `recv_file_ordered_fds` — the single descriptor for the ordered path.

`recv_file_direct_io_fallback_count` became an `aws_atomic_var` now that several threads can
record a fallback.

### Decoupling the body from the request

New `struct aws_s3_body_write` (`aws_s3_body_write_new_from_request`) moves the response
body, its buffer-pool ticket, and the request metrics off an `aws_s3_request` so the request
can be released as soon as its own completion accounting is done, instead of being held
alive until the bytes reach disk. It holds an owning meta request ref — a pending write may
be the last thing keeping the meta request alive.

### Resume token accounting

`num_bytes_delivered` and `num_bytes_delivered_total` are now tracked separately. They are
equal for ordered delivery, but diverge once parts land past a gap:

- `continuous_downloaded_bytes` ← the gap-free prefix, maintained by
  `next_contiguous_write_part` plus a min-heap (`completed_write_parts_tracker`) of writes
  that finished ahead of an earlier part.
- `total_downloaded_bytes` ← every byte written, gaps included.

### Ranged GET offset fix

Added `recv_file_object_offset_origin`, set by auto-ranged-get when it resolves the object
range. A part's delivery offset is its absolute object offset, so without this the range's
first byte would land that many bytes into the file rather than at the base position.
`recv_file_direct_io_base_position` was renamed `recv_file_base_position` since it now
applies to buffered writes too.


include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -390,21 +448,54 @@ struct aws_s3_meta_request {
...
+     * struct file, so they never contend on its reference count, and a slot needs no lock because
+     * only its own worker thread ever touches it.
+     *
+     * Opened lazily by the owning worker, since a meta request that only ever lands on a few workers
+     * should not pay for descriptors it never uses.
+     *
+     * No lock: the array is allocated before any part is dispatched and never resized, and each
+     * element is owned exclusively by one worker thread. Same argument as combine_slots. */
+    struct aws_s3_recv_file_fds *recv_file_write_fd_slots;
+    size_t recv_file_write_fd_slot_count;
```
TingDaoK, line 495: we don't "need" all those fds, since pwrite should be atomic, and thread safe. It's just an extra protection in case we are not using pwrite or anything change the state of fd.

source/s3_meta_request.c
```diff
@@ -2426,111 +2654,305 @@ static bool s_should_apply_backpressure(struct aws_s3_request *request) {
...
+           top->part_number == meta_request->synced_data.next_contiguous_delivered_part) {
+
+        struct aws_s3_completed_delivery entry;
+        aws_priority_queue_pop(&meta_request->synced_data.completed_deliveries_tracker, &entry);
+        meta_request->synced_data.num_bytes_delivered += entry.bytes;
+        ++meta_request->synced_data.next_contiguous_delivered_part;
+    }
+}
+
+/* Write one part's body to the file. Runs on a body-streaming event loop, concurrently with other
```
TingDaoK: I am here.

include/aws/s3/s3_client.h
```diff
@@ -8,6 +8,7 @@
 
 #include <aws/auth/signing_config.h>
 #include <aws/common/ref_count.h>
+#include <aws/common/tribool.h>
```
sbSteveK, line 11: Couldn't figure out what this was. Found a currently unmerged PR: https://github.com/awslabs/aws-c-common/pull/1272

Just noting that we need to merge that aws-c-common PR before this one or it'll break everything downstream.

source/s3_client.c
```diff
@@ -881,6 +858,60 @@ struct aws_s3_client *aws_s3_client_new(
...
+            .shutdown_callback_user_data = client,
+        };
+
+        client->file_io_elg =
+            aws_event_loop_group_new_default(client->allocator, num_file_io_threads, &file_io_elg_shutdown_options);
+
+        if (!client->file_io_elg) {
+            goto on_error;
+        }
+        client->synced_data.file_io_elg_allocated = true;
```
sbSteveK, line 912: Trivial: Do we need this? It looks like we ALWAYS set file_io_elg.

alternately, is there any case where we DON'T need file_io_elg? If there is, maybe we should be lazily assigning it to make this useful. If we want to eagerly ALWAYS set it, we should probably get rid of this.

tests/s3_tester.c
```diff
@@ -1644,6 +1693,7 @@ int aws_s3_tester_send_meta_request_with_options(
                     filepath_str = aws_s3_tester_create_file(allocator, options->get_options.object_path, NULL);
                 }
                 meta_request_options.recv_filepath = aws_byte_cursor_from_string(filepath_str);
+                printf("##### %s\n", aws_string_c_str(filepath_str));
```
sbSteveK: Delete?

source/s3_auto_ranged_get.c
```diff
@@ -1082,6 +1121,10 @@ static void s_s3_auto_ranged_get_request_finished(
             auto_ranged_get->synced_data.object_range_start = object_range_start;
             auto_ranged_get->synced_data.object_range_end = object_range_end;
             auto_ranged_get->synced_data.object_size = object_size;
+            /* A part's delivery offset is its absolute object offset, so the file sink needs the
+             * range's origin to map the range's first byte to the file's base position. Set before any
+             * body is delivered, since the range is resolved from the first response's headers. */
+            meta_request->recv_file_object_offset_origin = object_range_start;
```
sbSteveK: Worker threads read `recv_file_object_offset_origin` to figure out where each part goes (via origin). For ranged GETs this has to be set before worker thread writes. This is done using the same kind of implicit ordering lock sequence we had a problem with before. If this breaks, every part writes at wrong file offset.

We should make the ordering explicit (that origin is set) so we can detect and report if something is out of order. Just in case we ever move something around or outside of a lock.

include/aws/s3/private/s3_client_impl.h
```diff
@@ -227,6 +227,35 @@ struct aws_s3_client {
...
+     * TODO: route the upload path's parallel file reads here as well, so
+     * `parallel_input_stream_new_from_file` stops borrowing body_streaming_elg.
+     *
+     * Note: the ordered download path still writes on the meta request's io_event_loop, which comes
+     * from body_streaming_elg. That path is serialized by definition, so it has no use for a pool;
+     * moving it would mean moving all of that meta request's event delivery with it. */
+    struct aws_event_loop_group *file_io_elg;
+
+    /* Round-robin cursor handing out file_io_elg loop indices to parallel file writes. Only
+     * ever incremented; callers take the value modulo the loop count.
```
azkrishpy: > Only ever incremented
Is there an upper bound? Should we document it here?

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -70,6 +70,29 @@ struct aws_s3_combine_slot {
...
+
+    /* Set once the open has been attempted, so a failure is not retried per body. */
+    bool open_attempted;
+};
+
+/* One part that reached its sink ahead of an earlier part, parked in
+ * `synced_data.completed_deliveries_tracker` until the gap before it closes.
+ *
+ * Used for tracking the contiguous downloaded length, which is what the download resume token reports. */
+struct aws_s3_completed_delivery {
```
azkrishpy: naming does not convey at what level this struct is used. also can we reuse this struct somehow? https://github.com/awslabs/aws-c-s3/blob/5736a59215790f9c84123fc38b4577c2b30acfc5/include/aws/s3/s3_client.h#L224

source/s3_auto_ranged_get.c
```diff
@@ -808,6 +808,45 @@ static bool s_request_body_is_entire_download(
...
+    return meta_request->meta_request_level_running_response_sum != NULL &&
+           !meta_request->meta_request_level_checksum_combinable;
+}
+
+/* Resolve, once, whether this download delivers bodies out of object order, and record it on the meta
+ * request. Static here because discovery is the only place the question is ever asked: no other meta
+ * request implementation splits a download into parts, so no other one has an order to deliver out of,
+ * and every one of them leaves the flag at the false it was initialized to. */
+static void s_resolve_out_of_order_delivery(struct aws_s3_meta_request *meta_request) {
+    bool out_of_order = aws_s3_allow_out_of_order_delivery(
```
azkrishpy: suggestion: We can have the API be `aws_s3_allow_out_of_order_delivery(struct aws_s3_meta_request *meta_request)`? seems like everything we use within is from the meta_request anyway so might be a cleaner abstraction?

source/s3_client.c
```diff
@@ -898,6 +934,21 @@ struct aws_s3_client *aws_s3_client_new(
 
     *((bool *)&client->enable_read_backpressure) = client_config->enable_read_backpressure;
     *((size_t *)&client->initial_read_window) = client_config->initial_read_window;
+    *((enum aws_tribool *)&client->out_of_order_delivery) = client_config->out_of_order_delivery;
```
azkrishpy, line 937: do we want a a client config override if we do have an env level config exposed? Feels like it goes against our recommendation of using one client on a single machine? I partly understand the need for this considering each cli request opens up a new process and new client.. so maybe we don't need the env level config?

source/s3_meta_request.c
```diff
@@ -170,6 +178,236 @@ static void s_validate_meta_request_checksum_on_finish(
...
 }
 
+/* Bring `recv_filepath` into existence per the requested recv_file_option, settle whether writes
+ * can use O_DIRECT, and allocate the per-worker descriptor slots. No-op when the transfer has no
+ * receive file.
+ *
+ * On failure the error is already raised and the caller's `error:` label destroys the meta request,
+ * which is what deletes a file this transfer created. That is why `recv_filepath` is assigned before
+ * anything that can fail, and why `recv_file_delete_on_failure` is armed only once the file has been
+ * opened -- arming it earlier would let a CREATE_NEW that found a pre-existing file delete a file
```
azkrishpy, line 188: question: Could you clarify the condition in which this can happen and how arming `recv_file_delete_on_failure` can solve the issue?

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -390,21 +435,72 @@ struct aws_s3_meta_request {
...
+     *
+     * No lock: the array is allocated before any part is dispatched and never resized, and each
+     * element is owned exclusively by one worker thread. Same argument as combine_slots. */
+    struct aws_s3_recv_file_fds *recv_file_write_fd_slots;
+    size_t recv_file_write_fd_slot_count;
+
+    /* Descriptor for the ordered delivery path, which has no worker slot of its own. Owned by the meta
+     * request's io_event_loop thread, the only thread that writes when `out_of_order_delivery` is false,
+     * so it is never shared with a worker. */
+    struct aws_s3_recv_file_fds recv_file_ordered_fds;
```
azkrishpy, line 500: can this not be with `recv_file_write_fd_slots`? the slot_count would always be 1 is the contraint right?

source/s3_client.c
TingDaoK: we do use it in the shutdown process. 
```
        bool finish_destroy =
            ...
            client->synced_data.file_io_elg_allocated == false && ...
```
it keeps the client alive until the elg goes away.

source/s3_auto_ranged_get.c
TingDaoK: it's being applied to the non-parallel path as well. check `s_deliver_body_to_sink`

sbSteveK: Are we explicitly checking whether the origin is set in the non-parallel path? If not, we're doing the same implicit ordering that we specifically want to avoid in two places instead of just here.

tests/CMakeLists.txt
```diff
@@ -432,7 +433,21 @@ if(ENABLE_MOCK_SERVER_TESTS)
...
+    add_net_test_case(out_of_order_override_allocates_write_slots_mock_server)
+    add_net_test_case(parallel_write_failure_propagates_mock_server)
+    add_net_test_case(ordered_write_failure_propagates_mock_server)
+    add_net_test_case(noncombinable_checksum_forces_ordered_delivery_mock_server)
+    add_net_test_case(parallel_write_empty_part_mock_server)
+    add_net_test_case(parallel_write_create_or_replace_existing_mock_server)
+    add_net_test_case(parallel_write_create_or_append_mock_server)
+    add_net_test_case(parallel_write_write_to_position_mock_server)
+    add_net_test_case(parallel_write_ranged_get_aligned_start_mock_server)
+    add_net_test_case(parallel_write_ranged_get_unaligned_start_mock_server)
```
sbSteveK, line 451: Trivial: These ranged tests cover non-zero origin (with base position 0) and write to position/append tests cover a non-zero base position (with origin 0). We don't have one that test both at once.

If the file offset is `base_position + (object_offset - oritin)`, a bug that omits or doubles either would pass. A ranged get + write_to_position test would cover it. I guess I don't see how that'd be possible unless something else changes (like moving something out of the lock for the other issue mentioned earlier) so not super necessary but maybe it'll catch if we do something that causes the out of order?

include/aws/s3/private/s3_client_impl.h
```diff
@@ -227,6 +227,35 @@ struct aws_s3_client {
     /* Event loop group for streaming request bodies back to the user. */
     struct aws_event_loop_group *body_streaming_elg;
 
+    /* Event loop group dedicated to file I/O. Separate from body_streaming_elg because a write to a
```
DmitriyMusatkin, line 230: is there any concerns on sizing of elgs? should we have one elg for delivering data (whether streaming or to disk) and another elg just for events and other non-blocking stuff?
not necesserily in this pr, but eventually

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -70,6 +70,29 @@ struct aws_s3_combine_slot {
     size_t digest_len;
 };
 
+/* Wrapper of a file descriptors */
+struct aws_s3_recv_file_fds {
+    /* The owner's descriptor: O_DIRECT when direct I/O is on for this transfer, buffered otherwise or
```
DmitriyMusatkin: nit: this reads a bit funky, doe it have O_DIRECT value or an actual fd value?

include/aws/s3/private/s3_meta_request_impl.h
DmitriyMusatkin: maybe add start/end for some potential sanity checking?

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -390,21 +435,72 @@ struct aws_s3_meta_request {
...
+
+    /* This request's override of the client's `out_of_order_delivery`, from the creation options.
+     * AWS_TRIBOOL_UNSET means defer to the client. Immutable after init. */
+    enum aws_tribool out_of_order_delivery_override;
+
+    /* Whether bodies reach their sink out of object order: to a file through the descriptor of whichever
+     * worker takes the part, or to the body callback without waiting on the part ahead of it. Nonzero
+     * for yes.
+     *
+     * Written once, by the auto-ranged GET implementation as it finishes discovery. That is the earliest
```
DmitriyMusatkin, line 468: a bit of the warning: we will need to merge this carefully with the expected checksum changes

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -390,21 +435,72 @@ struct aws_s3_meta_request {
...
+     * response's headers -- and the last point at which no part but discovery's own exists, so nothing has
+     * been delivered under one answer and then finds another. Read-only from then on.
+     *
+     * Stays false under every other implementation. Out-of-order delivery is a statement about the order of
+     * a download's parts, and an implementation that does not split a download into parts has no such
+     * order. Note that this is about the implementation and not `type`: a GET whose query already names a
+     * partNumber carries type GET_OBJECT but runs as a default meta request, and so leaves this false.
+     *
+     * Atomic because a retried discovery resolves this while the event delivery task may be draining a
+     * telemetry or progress event from the attempt that failed. */
```
DmitriyMusatkin, line 479: those access meta req that is actively being modified? that seems dangerous, there is other fields in there that might be a concern?

include/aws/s3/private/s3_meta_request_impl.h
TingDaoK: sure, but i think to have a separate one will be more clear.

DmitriyMusatkin (approved):

tests/CMakeLists.txt
TingDaoK: I don't think it is necessary, they are independent, they them independently is good enough to me.

source/s3_auto_ranged_get.c
TingDaoK: https://github.com/awslabs/aws-c-s3/pull/685 here we go, the current impl doesn't has the issue since we just buffer every thing and start to stream after we finish the discovery, but adding the test makes me find another long-existing bug for default type.

include/aws/s3/private/s3_client_impl.h
TingDaoK: >should we have one elg for delivering data (whether streaming or to disk) and another elg just for events and other non-blocking stuff?

this is what i do in this PR, just the current sequentially streaming stays on the old elg for now. 
We can probably be more start about it. (eg: maybe default to half of the cpus instead of the current default) I don't have strong opinion yet.

include/aws/s3/private/s3_meta_request_impl.h
TingDaoK: that is the public interface, i think this structure more be overlap a bit, but i want the part index to keep it more straightforward (in case of empty part or some weird corner case). Will renaming though.

include/aws/s3/private/s3_meta_request_impl.h
TingDaoK: yeah, i figured. we are touching the same place here.

source/s3_client.c
TingDaoK: i think most of the client will follow the default here, just incase one of the sdk really don't want this to disable it for the whole sdk.

So, for a sdk user, if we don't enforce sdk to expose the same config to their user, the mitigation to break the behavior of some downstream use case somehow rely on, is to have the envrionment variable to enforce the old behavior.

include/aws/s3/private/s3_meta_request_impl.h
TingDaoK: It's for a rare corner case:
- discovery request fails with a retryable error.
- On re-prepare, `s3_request.c:121 sees num_times_prepared > 0`, finalizes the previous attempt's metrics, and queues a TELEMETRY event — which schedules the event delivery task on io_event_loop.
- That task reaches loads out_of_order_delivery unconditionally, before it looks at any event. So the load happens even though the only thing in the batch is a telemetry event.
- Attempt 2 may at mean time try to resolve this `out_of_order_delivery`.

It's not a real concern since the the event is not affect by this variable.

source/s3_meta_request.c
TingDaoK: for case where the failure happens before we try to create the file, and there is an existing file with the same path, and if we already set `recv_file_delete_on_failure`in the case, it will result in us deleting the file as an mistake.

azkrishpy: but we might have intended to append to the file? it might have been a resume from a previous pause?

azkrishpy (approved):

include/aws/s3/private/s3_client_impl.h
DmitriyMusatkin: ugh, i meant "one size for streaming, and different size for delivery".
its probably ok to size them the same now. but worth considering different sizing in future

DmitriyMusatkin (approved):

PR #680, opened 2026-09-14
move the check and relax it for the file streaming case

TingDaoK:
*Issue #, if available:*
- when we upload a file size that will result in a part size larger than half of the memory pool size, we now fail the request, but when we streaming directly from the file, the size we allocate is only 16MiB instead of the actual full part size. So, we can relax the check
- `aws_s3_request_get_payload_size` incorrect report the reserve size that ignored the `buffer_size` we calculated before

*Description of changes:*
- Move the part size check to after we initialized auto range put request, so that we got all those config resolved. And also avoid creating MPU.
- `AWS_FATAL_ASSERT(meta.size <= buffer_pool->mem_limit);` is no longer fatal, handles it gracefully to avoid crash. 


source/s3_auto_ranged_put.c
```diff
@@ -444,6 +444,28 @@ struct aws_s3_meta_request *aws_s3_meta_request_auto_ranged_put_new(
         aws_mem_release(allocator, auto_ranged_put);
         return NULL;
     }
+    /* The buffer reserved for each request is meta_request->part_size, EXCEPT when the body
+     * streams from a file: aws_s3_request_new then sizes it at 2 * g_streaming_buffer_size
+     * regardless of part size, so a part larger than max_part_size costs nothing extra. */
+    bool file_streaming =
+        auto_ranged_put->base.fio_opts.should_stream && auto_ranged_put->base.request_body_parallel_stream != NULL;
+
+    if (!file_streaming && part_size > client->max_part_size) {
```
azkrishpy, line 453: if it is file streaming, should we check that 
`min(part_size, 2*global_buffer_size) <= client->max_part_size` as a guardrail?

currently this will be true but if we ever support a greater global buffer size or lower max_part_size, we might quietly fail.

azkrishpy (approved):

DmitriyMusatkin (approved):

source/s3_auto_ranged_put.c
TingDaoK: yeah, but basically we are talking about a max_part_size is smaller than 2 MiB... 
And also, there is another level protection where when we reserve the part, we check if the part can fit or not. 

Honestly, I thought about just remove this check for all. But, it's a bit a behavior change for the case where a larger part will result in create MPU first and then fail the reservation.
So, keeping this check is just to keep the old behavior.

PR #681, opened 2026-09-15
API to Get Default S3 Memory Limits

sbSteveK:
Added `aws_s3_default_memory_limit_for_throughput` API to allow downstream to get whatever we default our memory pool size to based on throughput and whatever right-sizing we eventually apply.

Added warning logs related to enable_read_backpressure. If it's true and the read window is 0, the client will stall.


source/s3_client.c
```diff
@@ -902,6 +902,43 @@ struct aws_s3_client *aws_s3_client_new(
...
+     * parts can be scheduled and downloads stall until the caller calls
+     * aws_s3_meta_request_increment_read_window. Fail fast at client
+     * construction rather than allowing a stalled client.
+     *
+     * When backpressure is disabled with a positive window, the window value
+     * is stored but every gating check is bypassed. Warn but do not fail:
+     * the client is still functional at native pool ceiling capacity, just
+     * not throttled the way the caller may have expected. */
+    if (client->enable_read_backpressure) {
+        if (client->initial_read_window < client->part_size) {
```
azkrishpy: we might not always allocate part_size buffers so adding this check at the client level might not work as tests seem to be indicating? if we stream from file we might use lesser.

azkrishpy: Never mind, the tests seem to be different issue but the file streaming case can still fail we need more tests.

TingDaoK (approved):

PR #682, opened 2026-09-17
requests parts in parallel.

TingDaoK:
*Issue #, if available:*

### What

A download that delivers out of order now issues its range requests across as many far-apart regions of the object as it has connections, instead of sweeping the object from start to end. The requests in flight at any moment are in K far-apart regions rather than one contiguous span.

### How

Block-cyclic distribution (the same decomposition MPI uses for distributing array elements across processes). At discovery finish, the remaining parts are divided into K contiguous regions, where K = min(max_active_connections, parts_remaining). The first (N % K) regions get ⌈N/K⌉ parts; the rest get ⌊N/K⌋. A single counter drives the handout — no per-region state, no allocation for the spread itself.

Example: 10 parts across 3 connections → regions of 4, 3, 3 parts. Handout rotates: 2, 6, 9, 3, 7, 10, 4, 8, 11, 5.

The spread decision is made at discovery finish, alongside the delivery order decision. Spreading requires out-of-order delivery (it *is* what hands parts out of order), so it stands down when delivery is ordered.

### Delivery validation mask

A per-part byte mask is allocated at discovery, marked when each part finishes sinking (written to file or delivered through the body callback), and checked before the meta request declares success. Every byte must be 1 — no missing parts, no duplicates. Catches formula bugs and delivery-path bugs at the point of corruption rather than in the finished file.

- Marking happens under the synced_data lock at two sites: the write-task completion (file sink) and the completed_deliveries loop (callback sink, both ordered and out-of-order).
- Out-of-range or duplicate part → `AWS_ERROR_INVALID_STATE`, meta request fails gracefully.
- Success path final check → `FATAL_ASSERT` (reaching success with a gap is an unrecoverable accounting bug).
- Freed on success, on error wind-down (destroy), or on rediscovery retry.

### Environment variables

- `AWS_CRT_S3_FORCE_SEQUENTIAL_REQUESTS`: Makes every download request parts in object order instead of spreading. Only affects request order, not delivery order.

Both are read once per client at construction.


include/aws/s3/private/s3_auto_ranged_get.h
```diff
@@ -72,6 +72,24 @@ struct aws_s3_auto_ranged_get {
         uint32_t num_parts_failed;
         uint32_t num_parts_checksum_validated;
 
+        /* Block-cyclic spread: parts are handed out across `spread_count` far-apart regions of the
+         * object instead of sweeping it from start to end. 0 means object order (no spreading).
+         *
+         * The first `spread_large_lanes` regions get `spread_stride + 1` parts each; the rest get
+         * `spread_stride`. `spread_first_part` is the first part to spread (discovery may have
```
azkrishpy, line 79: Could use the example/definition for each member here.

source/s3_auto_ranged_get.c
```diff
@@ -808,6 +824,104 @@ static bool s_request_body_is_entire_download(
...
+    }
+
+    const uint32_t parts_remaining = total_num_parts - first_part + 1;
+    const uint32_t K =
+        aws_min_u32(aws_s3_client_get_max_active_connections(meta_request->client, meta_request), parts_remaining);
+    if (K <= 1) {
+        return;
+    }
+
+    /* Block-cyclic distribution: the first (N % K) regions get ceil(N/K) parts, the rest get
```
azkrishpy, line 856: might be worth mentioning N is number of parts requested (only K is defined as connections from above code block)

source/s3_auto_ranged_get.c
```diff
@@ -808,6 +824,104 @@ static bool s_request_body_is_entire_download(
...
+     * floor(N/K).
+     *
+     * Example: 10 parts (2-11) across 3 connections.  stride = 10/3 = 3, large_lanes = 10%3 = 1.
+     *
+     *   Region 0 (large):  parts 2  3  4  5   (stride+1 = 4 parts)
```
azkrishpy, line 861: should we define what "Region" means here?

source/s3_auto_ranged_get.c
```diff
@@ -808,6 +824,104 @@ static bool s_request_body_is_entire_download(
...
+        total_num_parts,
+        K);
+}
+
+/* The 1-based part number to hand out next. When spreading is off, parts go out in object order. When on,
+ * parts rotate across K regions via block-cyclic distribution.
+ *
+ * Continuing the example from s_init_spread_synced (10 parts, K=3, stride=3, large_lanes=1, first=2):
+ *
+ *   i=0: lane=0 step=0 -> region_start = 2               -> part 2
```
azkrishpy, line 887: might be worth documenting what a step is

source/s3_meta_request.c
```diff
@@ -3307,15 +3341,22 @@ static void s_s3_meta_request_event_delivery_task(struct aws_task *task, void *a
...
-                struct aws_s3_completed_delivery completed;
-                aws_array_list_get_at(&completed_deliveries, &completed, i);
+
+        /* Mark every delivered part in the validation mask and update byte accounting.
+         * completed_deliveries is populated by both the ordered and out-of-order paths now. */
+        for (size_t i = 0; i < aws_array_list_length(&completed_deliveries); ++i) {
+            struct aws_s3_completed_delivery completed;
+            aws_array_list_get_at(&completed_deliveries, &completed, i);
+
+            if (s_mark_part_delivered(meta_request, completed.part_number) != AWS_OP_SUCCESS) {
```
azkrishpy: question: Are we marking parts as delivered before delivery? is this because it is the end of the request?

azkrishpy (approved):

include/aws/s3/private/s3_auto_ranged_get.h
```diff
@@ -72,6 +72,24 @@ struct aws_s3_auto_ranged_get {
...
+         * already taken part 1). `spread_parts_handed_out` is the counter that drives the formula.
+         *
+         * Together these produce a part number from a single increment -- no per-lane state. */
+        uint32_t spread_count;
+        uint32_t spread_stride;
+        uint32_t spread_large_lanes;
+        uint32_t spread_first_part;
+        uint32_t spread_parts_handed_out;
+
+        /* One byte per part (1-indexed, so element 0 is part 1). Set when a part number is handed out
```
sbSteveK: This doesn't sound like it fits the object_range_known. I also don't see it used in this PR the way it's being described.

source/s3_auto_ranged_get.c
```diff
@@ -391,6 +394,9 @@ static bool s_s3_auto_ranged_get_update(
                     &request->part_range_end);
 
                 ++auto_ranged_get->synced_data.num_parts_requested;
+
+                /* Validate: mark this part in the mask. Catches duplicates immediately and gaps at the
```
sbSteveK: There's no marking of a mask happening here.

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -281,6 +281,14 @@ struct aws_s3_meta_request {
          * failed.)*/
         uint32_t num_parts_delivery_completed;
 
+        /* One byte per part (0-indexed: part 1 is index 0). Set when a part finishes sinking —
```
sbSteveK, line 288: This is where the description in ranged_get header came from.

include/aws/s3/private/s3_auto_ranged_get.h
sbSteveK: Looks like `parts_delivered_mask` was moved to a different header and this was just missed in delete.

TingDaoK: good catch, looks like AI moved the code over, but not the comments...

include/aws/s3/private/s3_auto_ranged_get.h
TingDaoK: they are explained when we assign them, i think it make more sense there with more context?

TingDaoK: renamed and moved the example and comments to the header file here, hope it's more clear now.

include/aws/s3/private/s3_meta_request_impl.h
```diff
@@ -290,6 +290,14 @@ struct aws_s3_meta_request {
...
+         * written to the file or delivered through the body callback. Checked when the meta request
+         * is about to complete: every byte must be 1. Catches any path that drops or duplicates a
+         * part. Allocated by the auto-ranged-get implementation once total_num_parts is known;
+         * NULL for meta request types that do not split a download into parts. */
+        uint8_t *parts_delivered_mask;
```
DmitriyMusatkin, line 298: that would be 10kb worst case for every meta-request? its probably ok, but slightly rubs me the wrong way on how much mem its wasting. Nit: on the naming, mask is not super accurate here, since you are using whole byte to indicate bool condition. 

Any particular reason why keeping next part to be delivered per spread would not work?

source/s3_auto_ranged_get.c
```diff
@@ -441,6 +444,16 @@ static bool s_s3_auto_ranged_get_update(
         }
 
         if (!work_remaining) {
+            if (!aws_s3_meta_request_has_finish_result_synced(meta_request)) {
+                /* Success: every part must have been delivered. */
+                if (meta_request->synced_data.parts_delivered_mask != NULL) {
+                    for (uint32_t p = 0; p < meta_request->synced_data.parts_delivered_mask_count; ++p) {
+                        AWS_FATAL_ASSERT(meta_request->synced_data.parts_delivered_mask[p]);
```
DmitriyMusatkin: we had really bad luck with fatals. maybe not have it in production code? error out instead?

source/s3_auto_ranged_get.c
```diff
@@ -808,6 +821,93 @@ static bool s_request_body_is_entire_download(
...
+    }
+
+    /* Spreading is what hands parts out of object order; an ordered download must sweep contiguously. */
+    if (aws_atomic_load_int(&meta_request->out_of_order_delivery) == 0) {
+        return;
+    }
+
+    const uint32_t first_part = auto_ranged_get->synced_data.num_parts_requested + 1;
+    const uint32_t total_num_parts = auto_ranged_get->synced_data.total_num_parts;
+    if (first_part > total_num_parts) {
```
DmitriyMusatkin, line 848: which case does this cover? single part?

source/s3_auto_ranged_get.c
```diff
@@ -808,6 +821,93 @@ static bool s_request_body_is_entire_download(
...
+        return;
+    }
+
+    const uint32_t parts_remaining = total_num_parts - first_part + 1;
+    const uint32_t num_regions =
+        aws_min_u32(aws_s3_client_get_max_active_connections(meta_request->client, meta_request), parts_remaining);
```
DmitriyMusatkin: hmm, max active connections as num regions seems a bit high? whats the logic for using that?

source/s3_auto_ranged_get.c
```diff
@@ -808,6 +821,93 @@ static bool s_request_body_is_entire_download(
...
+        (void *)meta_request,
+        first_part,
+        total_num_parts,
+        num_regions);
+}
+
+/* The 1-based part number to hand out next. When spreading is off, parts go out in object order. When on,
+ * parts rotate across the spread's regions. See the spread fields in s3_auto_ranged_get.h for what each
+ * one means. */
+static uint32_t s_next_part_number_synced(struct aws_s3_meta_request *meta_request) {
```
DmitriyMusatkin, line 886: does it have to be synced? it should only run once in scheduling thread?

source/s3_auto_ranged_get.c
TingDaoK: or the object size is smaller than the part size

source/s3_auto_ranged_get.c
TingDaoK: My thinking was we will have number of connections in parallel, so that the parallelism will be spread evenly cross the object. I am also open to other ideas.

source/s3_auto_ranged_get.c
TingDaoK: it's used in the update vtable function, when the scheduling thread ask the meta request for the next request to work on. 
The synced in the function mostly to protect the race between the discovery request thread (Or all the other request threads) vs the scheduling ask for work. 

This function itself doesn't really need a lock as long as the discovery completes, which protected by the other boolean protected by the lock. 

But, we are grabbing the lock in the update function either way, I don't think it's much a concern to have this math in the lock?

include/aws/s3/private/s3_meta_request_impl.h
TingDaoK: updated to bit mask. 🤦‍♂️

source/s3_meta_request.c
```diff
@@ -2933,6 +2934,108 @@ static int s_deliver_body_to_sink(
...
+            "-byte mask, so the mask length and its part count disagree.",
+            (void *)meta_request,
+            part_number,
+            byte_index,
+            meta_request->synced_data.parts_delivered_mask_length);
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    uint8_t *byte = &meta_request->synced_data.parts_delivered_mask[byte_index];
+    const uint8_t bit = (uint8_t)(1u << (bit_index % 8));
```
DmitriyMusatkin, line 2971: nit: maybe call it something like bit mask? cause you are using it to check whether bit is set or not

source/s3_meta_request.c
```diff
@@ -2933,6 +2934,108 @@ static int s_deliver_body_to_sink(
...
+            num_bytes);
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    for (uint32_t byte = 0; byte < num_bytes; ++byte) {
+        /* A whole byte covers 8 parts, so all of its bits are required. A part count that is not a
+         * multiple of 8 leaves a partial byte at the end whose high bits stand for no part at all, so
+         * only its low bits are required. That byte exists only when there is a remainder, which is
+         * exactly when `num_bytes` reaches past `num_full_bytes`. */
+        const uint8_t required = (byte < num_full_bytes) ? 0xFF : (uint8_t)((1u << (num_parts % 8)) - 1);
```
DmitriyMusatkin, line 3020: oh, thats a bit devious :) it pushes 1 to to max position and then subtracts 1 to make everything under 1

DmitriyMusatkin (approved):

PR #684, opened 2026-09-18
Expected checksum and validation mode changes

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Added expected checksum field to meta request. This allows users who already know checksum for the requested bytes to pass it (ex. calling headObject before crt), which will avoid discovery phase for full object validation.

Additionally, added a way to disable full object validation. Use case is user calling ListObject upstream, which does not return checksums, but still wants to skip discovery phase.


source/s3_meta_request.c
```diff
@@ -170,6 +170,81 @@ static void s_validate_meta_request_checksum_on_finish(
...
+    if (expected_checksum.len == 0 || algorithm < AWS_SCA_INIT || algorithm > AWS_SCA_END) {
+        AWS_LOGF_ERROR(
+            AWS_LS_S3_META_REQUEST,
+            "id=%p Cannot create meta request; expected_checksum and expected_checksum_algorithm must both be set, "
+            "to a known algorithm.",
+            (void *)meta_request);
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
+    if (options->type != AWS_S3_META_REQUEST_TYPE_GET_OBJECT) {
```
TingDaoK, line 203: is there a specific reason why we cannot support `AWS_S3_META_REQUEST_TYPE_DEFAULT` with `operation_name` is `AWS_S3_REQUEST_TYPE_GET_OBJECT`?

include/aws/s3/s3_client.h
```diff
@@ -317,6 +317,43 @@ enum aws_s3_checksum_location {
...
+/**
+ * Which checksums a download is validated against.
+ *
+ * A download is made of one or more part responses, each of which may carry a checksum of its own body, and the
+ * whole download may also be covered by a single checksum (the object's, or one the caller supplies). These modes
+ * pick which of the two the client checks.
+ *
+ * Only applies to AWS_S3_META_REQUEST_TYPE_GET_OBJECT; ignored by other meta request types.
+ */
+enum aws_s3_checksum_validation_mode {
```
TingDaoK, line 329: i am not sure why we need the part only mode here? And how will it work with the default type or the get with only one part? 

I guess the reason is to avoid any extra to fetching the full object checksum, so, should it be more like SKIP_FETCHING (or some better naming)? Just skip any extra step we need to get the full object checksum, but if we already have it, why not use it? 

On the other end, does it make sense to introduce a FULL_OBJECT_ONLY mode? I'd say once we have the full object checksum validated, the parts level is a bit redundant and only be used to catch the error earlier from durability perspective?

DmitriyMusatkin: part is a bit misleading here. what it really means is verify checksum on every request possible. that will support single requests or parallel reqs, etc... (i fixed the gating on default get). 

my mental model was 
- part-only (or request only) - this enabled checking on requests only. no needs to setup cross request checksum validation in any way. and providing expected checksum here is an error because it will not be validated. (this is basically the mode where caller did list objects before and cannot provide individiual checksum for each object)
- full-object mode - do part validation as before, but also attempt to validate full object checksum. that checksum either needs to be discovered or can be provided by the user. (this is basically the mode where caller did head object before hand and can provide checksum)

yeah, i considered full object only mode for this, but that seems out of scope for now. but i can image we will add it in future

TingDaoK: >The client never looks for a checksum covering the whole download

The comment is confusing that says it will not cover the whole download, make me think of what if the whole download is just one request, and wonder what will happen there.

Maybe focus on there will be no cross requests state for checksum validation in the documentation? 

But, I still think it will drop the durability guarantee if we do part only here.

TingDaoK (approved):

include/aws/s3/s3_client.h
DmitriyMusatkin: yeah, let me clarify the comment here. it does cover full object if there is only one request. 
it is a durability loss if caller does not have expected checksum, but also does not want crt to call head object. it  is up to caller to decide if they are ok with durability drop

PR #685, opened 2026-09-18
Fix default range request callback report

TingDaoK:
*Issue #, if available:*
```
typedef int(aws_s3_meta_request_receive_body_callback_fn)(

    /* The meta request that the callback is being issued for. */

    /* The body data for this chunk of the object. */
    const struct aws_byte_cursor *body,

    /* The byte index of the object that this refers to. For example, for an HTTP message that has a range header, the
       first chunk received will have a range_start that matches the range header's range-start.*/
    uint64_t range_start,

    /* User data specified by aws_s3_meta_request_options.*/
    void *user_data);

```
*Description of changes:*
- The `range_start` in the body callback is the `byte index of the object that this refers to.`
- It's confusing, but hey, the ship has sailed since first we add it. Not complain yet
- The auto range get respects this, but the default type get is not.
- Fixing the default type get to respect this.
- Add test so that with a range to get that write to a file will not mess up with file offset
- Add test to make sure the body callback respect the range correctly.


source/s3_meta_request.c
```diff
@@ -2735,6 +2735,11 @@ static bool s_should_apply_backpressure(struct aws_s3_request *request) {
  * a ranged download has to subtract the range's origin for its first byte to land at the base
  * position rather than that many bytes into the file. */
 static uint64_t s_s3_recv_file_offset(const struct aws_s3_meta_request *meta_request, uint64_t object_offset) {
+    /* The mapping is only meaningful once the origin it subtracts is known, and the origin cannot say so
+     * for itself: 0 is what it holds before anything resolves it and also what a whole-object download
+     * resolves it to. Reaching here first would place the body at its absolute object offset rather than
+     * the base position, with nothing about the outcome looking wrong. */
+    AWS_ASSERT(meta_request->recv_file_object_offset_origin_resolved);
```
sbSteveK: AWS_ASSERT gets removed on non-debug builds. Any reason not to keep this and the below check in place for release builds and fail loudly?

sbSteveK (approved):
Curious about only debug fails. Otherwise looks good.

